PR #38, opened 2018-07-05
Rh icon update

markcaron:
Add ALL the icons!


kylebuch8:
I think the resulting rh-icon.compiled.js file is too big. It's coming in at 594 KB. I think we should look at breaking this down into sets of icons that we deliver. What do you think?

markcaron:
Agreed. Let's discuss how to properly break it up.

mwcz:
Should we close this PR?

mwcz:
Closing for now.

PR #48, opened 2018-07-13
run prettier on everything

mwcz:
Many files haven't had prettier run on them, because they haven't changed since prettier was added.  This PR includes an initial run of prettier on every file (except those in .prettierignore).  Should have done this when prettier was first added, but better late than never.

I'm also activating prettier for scss files, so they will have uniform formatting in addition to JS and JSON.


mwcz:
Nevermind, prettier formatting of scss comments is [very buggy](https://github.com/prettier/prettier/search?q=scss+comments&type=Issues).

PR #50, opened 2018-07-16
adding a story for rh-datetime

kylebuch8:


elements/rh-datetime/rh-datetime.story.js
```diff
@@ -0,0 +1,191 @@
...
+  const yearuntil = new Date(
+    new Date().setFullYear(new Date().getFullYear() + 1)
+  );
+  const yearsuntil = new Date(
+    new Date().setFullYear(new Date().getFullYear() + 10)
+  );
+
+  function timer() {
+    document.getElementById("realtime").setAttribute("datetime", new Date());
+    window.requestAnimationFrame(timer);
```
mwcz: I think this would be better as a setInterval every second.  As written it's equivalent to a setInterval every 16 milliseconds; all that rAF adds is synchronizing the interval to match the screen's refresh rate.

kylebuch8: But rAF makes sure that the timer is *super* accurate.

mwcz: It seems like overkill to me but in a story it's not that big a deal.  If it were in the component itself I'd urge setInterval or a throttle within rAF.

kylebuch8: Ok, let's roll with it.

PR #62, opened 2018-07-19
apply new method of injecting HTML and CSS

mwcz:
Here it is, the big one.  Let me know if you want to meet up to go through this.


kylebuch8:
@mwcz - We should also update the generator with these changes.

mwcz:
Yeah, I meant to mention that in the comment.  Still need to make those changes.

PR #76, opened 2018-07-23
Adding bullseye back and fixing the grid gutters

### LINKED ISSUE #75: Add bullseye and story back to rh-layouts
We ran into a bunch of conflicts with the old PR (https://github.com/RHElements/rhelements/pull/39) so we just need to do the work over again real quick.

kylebuch8:
Fixes #75

### DIFF
--- a/elements/rh-layouts/demo/index.html
+++ b/elements/rh-layouts/demo/index.html
@@ -143,6 +143,21 @@
         </section>
 
       </section>
+
+      <section>
+        <h2>Bullseye</h2>
+        <p>Used for centering all children (typically a single item) both vertically and horizontally within a parent element.</p>
+        <section>
+          <h3>Basic Bullseye</h3>
+          <div class="rh-l-bullseye">
+            <div class="rh-l-bullseye__item">Bam!</div>
+          </div>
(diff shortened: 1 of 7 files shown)


mwcz:
The demo page for this rh-layouts looks completely different than the other demo pages.  Is this one a sign of where all the others are headed?

kylebuch8:
@markcaron - Can you comment on the styling of the demo page?

markcaron:
The demo in Storybook?

I had to adjust the background so that the grid system (its columns) were evident. We should look to make them more consistent in future version, but it's not necessarily _the_ direction we're going.

mwcz:
@markcaron I don't think there is a Storybook story for rh-layouts.  I was asking about the demo page at [elements/rh-layouts/demo/index.html](https://github.com/RHElements/rhelements/blob/master/elements/rh-layouts/demo/index.html).  It was edited in this PR which is the only reason I thought to ask.

I just noticed that the demo page looks very different from all the other demo pages and would like to have a plan for convergence.  No need to solve it here, though.

mwcz:
The changes look good to me, but I don't feel confident reviewing the sass.

PR #78, opened 2018-07-24
Dekludge tests

### LINKED ISSUE #24: figure out how to hoist elements/* with lerna
It occurred to me today that the whole `.wct-kludge` "solution" could be avoided if lerna would simply symlink `elements/*` into `node_modules/@rhelements/`.  Then we could point tests at `node_modules/@rhelements/*` instead of `test/.wct-kludge/elements/*`.

Lerna only symlinks packages into node_modules if they are declared as a dependency.  Currently `node_modules/@rhelements/` only contains cp-sass, rh-card, rh-datetime, rhelement, and rh-onebox because those are the only rhelements that are depended upon by other rhelements.

We could potentially create something like `elements/all` which has a dependency on every other rhelement.  There may be a better solution.  I didn't find one while scanning lerna's docs and issues though.  Worth looking into.

mwcz:
fixes #24


kylebuch8:
@mwcz - I just can't get this to work locally. breakpoint-sass continues to be the culprit.

kylebuch8:
@mwcz - If I reclone the repo and have a fresh start, it works fine. I have no idea why I have to do that. Can we look at this again tomorrow?

Issue #86, opened 2018-07-27
Remove demo files from NPM publishing on all elements and themes


markcaron:
Remove demo files from NPM publishing on all elements and themes

mwcz:
Going even further, we should remove tests and *maybe* even the `src/` dir from the published artifact, leaving only the things necessary for production use.

kylebuch8:
I'm not sure about removing the demo files. I think it makes sense to leave them so developers can refer to them if they want. I use the README files and example files in node_modules directories all the time.

mwcz:
Closing, but can be re-opened if anyone is interested.

PR #94, opened 2018-07-31
Moving CP Theme to a new themes dir

markcaron:


mwcz:
@kylebuch8  Already talked about this with Mark, but this change will break your ability to run demo files from node_modules/@rhelements.  The paths won't line up anymore.  Related to #86

mwcz:
@markcaron Are the requested changes doable?

markcaron:
Yep! Sorry for the delay.

PR #104, opened 2018-08-07
hide fouc

### LINKED ISSUE #29: revisit methods of preventing FOUC
links:
 - [:defined](https://developers.google.com/web/fundamentals/web-components/customelements#prestyle)
 - [howto-tabs using :defined](https://github.com/GoogleChromeLabs/howto-components/blob/master/elements/howto-tabs/demo.html)

:defined cannot be polyfilled

### LINKED ISSUE #110: use a CSS animation instead of setTimeout to reveal
If users opt in to rhelements.css, the page will be hidden until all custom elements are upgraded.  There is a timer in reveal.js that will reveal the page after 1 second, as a fallback in case elements are loading too slowly, or in case of a JS error.  To further insulate the reveal from JS problems, we should remove the setTimeout in reveal.js and instead, put a CSS animation to reveal the body into rhelements.css.

Something like this:

```css
@keyframes reveal {
  from {
    opacity: 0;
  }
  to {
    opacity: 0;
  }
}

body[unresolved] {
  opacity: 0;
  animation: reveal 0.16s 1s 1 forwards;
}

### PR DESCRIPTION
**Pull request by mwcz**
fixes #29
fixes #110 

@kylebuch8 This PR is not ready to merge.  I just got it started so the review can begin early.  It needs review and docs.  Also, the example of hiding FOUC is in the rh-card demo file, which is definitely not the right place (all demo files should get no-fouc, but the rh-card demo should be restored to its prior state, not hijacked for no-fouc testing).

Edit: this PR is now ready to merge!

### CODE REVIEW

**Comment by kylebuch8** [comment]
@mwcz Are we good to pull this in now?

**Comment by mwcz** [comment]
@kylebuch8 Once travis stops throwing errors about the "ansi-escapes" npm package.  I just cleared the travis cache, hopefully that fixes it.

File: elements/rhelement/gulpfile.js [hunk 1]
```diff
@@ -6,18 +6,20 @@ const babel = require("gulp-babel");
...
 const replace = require("gulp-replace");
-
+const cleanCSS = require("gulp-clean-css");
 const trim = require("gulp-trim");
 const del = require("del");
 let watcher;
 
 gulp.task("clean", () => {
-  return del(["./*.umd.*"]);
+  return del(["./**/*.umd.*", "./*.min.css"]);
```
mwcz: Nice catch, thanks!

elements/rhelement/package.json
```diff
@@ -21,5 +20,8 @@
   },
   "license": "ISC",
   "dependencies": {},
-  "generator-rhelement-version": "0.3.4"
+  "generator-rhelement-version": "0.3.4",
```
mwcz: @kylebuch8 devDependencies should be in the root package.json.

mwcz: @kylebuch8 Oh yeah, this is my PR, I will just go fix it. :)

kylebuch8:
@KrystalMoore16 Do you want to review this too?

PR #106, opened 2018-08-09
RH-CTA

markcaron:
RH-CTA should be ready to go.


elements/rh-cta/demo/index.html
```diff
@@ -10,29 +10,88 @@
     <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.5/require.min.js"></script>
     <script>require(['../rh-cta.compiled.js'])</script>
```
mwcz: Should be `rh-cta.umd.js`

elements/rh-sass/mixins/_mixins.scss
```diff
@@ -68,6 +68,10 @@
   }
 }
 
+@mixin rh-arrow-color($color) {
```
castastrophe: Why build a mixin for this one property?  Is the goal to evolve it to take care of other styling as well?

elements/rh-sass/mixins/_mixins.scss
```diff
@@ -106,3 +110,35 @@
...
+/// --rh-teaser__subtitle--link-color--visited
+
+/// Defined locally for sanity
+/// --rh-local-[component-block][__element][--context][--state][--PropertyCamelCase]
+/// --rh-local-cta--link-color--inverted
+///
+/// ---------------------------------------------------------------------------
+///
+///                     ↓ component,         ↓ theme,         ↓ fallback
+@mixin rh-define-vars($this-component-var, $this-theme-var, $this-fallback) {
```
castastrophe: After chatting with @starryeyez024 I think I have an idea that would make this mixin take care of more than defining one property.  This mixin would be updated to accept a namespace and 1 map and the component's styles would use a very easy to read map to set these values (example on that code instead of here).

elements/rh-cta/src/rh-cta.scss
```diff
@@ -1,155 +1,367 @@
...
+@mixin make-default-cta-variation(
+    $this-selector: ".primary",
+    $this-name: "primary",
+    $this-color: $rh-cta--default-color,
+    $this-color-hover: $rh-cta--default-color--hover,
+    $this-color-focus: $rh-cta--default-color--focus,
+    $this-color-visited: $rh-cta--default-color--visited,
+    $this-modifier: ""
+  ) {
+
```
castastrophe:
```scss
$cta-settings: (
     primary-link-color: primary-link-color, $rh-cta--default-color,
     primary-link-color--visited: primary-link-color--visited, $rh-cta--default-color--visited,
     primary-link-color--hover: primary-link-color--hover, $rh-cta--default-color--hover,
     primary-link-color--focus: primary-link-color--focus, $rh-cta--default-color--focus
);
```

^ the map contains the variable name as the key with a value that is a list containing the theme value and the fallback color.  If only one value exists in the list, the mixin will still generate the right variable listing.

castastrophe: This example looks a bit redundant with the theme names being the same but there will be a lot of examples where the theme name might not resemble the variable name at all.

elements/rh-sass/functions/_functions.scss
```diff
@@ -127,3 +127,11 @@
     }
     @return $result;
 }
+
+/// RH Var
+/// Uses the name created by rh-define-vars() to retrieve the local var
+/// @requires: @mixin rh-define-vars()
+/// @param {String} $this-local-var - Variable name to be used
+@function rh-var($this-local-var) {
```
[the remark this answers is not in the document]
markcaron: The $fallback should be handled with the `rh-define-vars()` mixin—which sets all the Vars including the fallback. I added the comment `@requires: @mixin rh-define-vars()`

The point of this function is to make calling the localized CSS Var easier. The localized CSS Var simply points to the CSS Vars stack with the fallback (in `rh-define-vars()`).

elements/rh-cta/src/rh-cta.scss
markcaron: Let's meet up next week to work on this idea.

markcaron:
@mwcz, I've updated the Compiled -> UMD request. Also removed breakpoint-sass dependency because it wasn't necessary and causing `npm install` problems for a lot of us.

elements/rh-sass/functions/_functions.scss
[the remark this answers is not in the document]
castastrophe: Yes that's a good point but it might have more reuse as a function if we abstract it a bit.  For example if we rewrite the function like this:

```scss
@function rh-var($var-name, $namespace: local, $fallback: null) {
     @if ($fallback) {
         @return var(--rh-#{$namespace}-#{$var-name}, #{$fallback});
    } @else {
         @return var(--rh-#{$namespace}-#{$var-name});
    }
}
```

Now this becomes a multi-use function that can accept a different namespace and optional fallback (which can be either a color or another variable name).  Just an idea to make this really great function helpful for different use-cases.

elements/rh-sass/mixins/_mixins.scss
castastrophe: Cool!  Can't wait!

elements/rh-sass/functions/_functions.scss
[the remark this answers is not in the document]
markcaron: Good thought! You in next week (like Wednesday)? Let's work this out along with the map idea.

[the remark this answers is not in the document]
castastrophe: Perfect - I'm in the office on Wednesdays so that's works great.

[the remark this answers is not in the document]
castastrophe: Proposed mixin: https://www.sassmeister.com/gist/ceb6de36877eb49c69be87c29b5a1ba2

kylebuch8:
@markcaron Can we close this?

PR #111, opened 2018-08-13
RH cta fork

starryeyez024:


mwcz:
@starryeyez024 looks like this PR has some conflicts with master.  Let me know if I can be of any help resolving them.

starryeyez024:
@mwcz thanks for your offer! this MR is against another feature branch for comparison of Sass approaches. Not sure if we need to merge it or just discuss but I updated it so its a clean MR.

elements/cp-themeset/sass/_colors.scss
```diff
@@ -22,6 +22,7 @@
   //  THEME BASE PALETTE COLORS
   // ========================================================================
   */
+  --rhe-theme--color--transparent:          transparent;
```
castastrophe: I wonder if it makes sense to have colors defined as CSS variables at all.  It seems like colors as Sass variables is sufficient and won't add to the compiled CSS.

PR #120, opened 2018-08-20
Removing older themes in elements directory

markcaron:
Removing elements/cp-themeset and elements/test-themeset


mwcz:
Lake Guymard's Tentacle Monster

mwcz:
That was penance for a botched LGTM above.

Issue #80, opened 2018-07-25
consolidate gulpfiles into a single root gulpfile


mwcz:
And since some elements have customized their gulpfiles, this issue involves creating a scheme for hooking into the root gulpfile with custom tasks, or common tasks shared among a few elements.

mwcz:
Just brainstorming here:

`gulpfile.js` at the root contains all the universal task implementations.  Have the gulpfile accept optional arguments, the names of the elements to build.  If run with no arguments, it builds all elements.

```bash
npm run build # runs gulp and builds all elements
npm run build -- rh-card # builds only rh-card
npm run build -- rh-card rh-cta rh-button # builds all three
```

Create a `scripts/gulp/` dir.  In it, put any common but not universal gulp tasks, one per file.  For example, the merge task which is used by most (but not all!) elements to embed the HTML and CSS/SCSS into the JS output file, would look as follows:

**scripts/gulp/merge.js**
```js
import gulp from "../wherever/is/gulp.js";

export default gulp
    .src(`elements/${rhelement}/src/${rhelement}.js`)
    .pipe(replace(/* ... implementation here ... */))
    .pipe(gulp.dest(`elements/${rhelement}/${rhelement}.js`));
```

Each rhelement should have a `gulptasks.js` file (name up for argument):

```js
export merge from "../../scripts/gulp/merge.js";
```

And that file could also export any other tasks the particular rhelement needs.  It could also export its own unique task instead of re-exporting an existing common one.  Not sure if this should be a recommended practice, or if we should enforce all tasks being added to `scripts/gulp/`...

In this way, the rhelement can hook into the root gulpfile with customizations.

kylebuch8:
@mwcz Can we close this? I think we've accomplished this with the gulpfile factory.

mwcz:
Yeah, good catch, this is done.

PR #122, opened 2018-08-20
Fix umd globals

### LINKED ISSUE #107: allow rhelements to be used with plain old script tags

### LINKED ISSUE #121: lift limitations on third-party dependencies
The current limitations on using third-party dependencies are pretty strict.  Let's lift them.

mwcz:
This PR introduces [Rollup](https://rollupjs.org/) as a bundler.  It takes over from Babel the generation of UMD and ESM files for each RHElement.  Babel is still used to convert es6+ code to es5, for the UMD version.  The ESM version is not transpiled with Babel, since any environment that supports ES Modules is fairly certain to support all other es6 features as well.  This enables using third-party dependencies without the prior draconian limitations, and allows using RHElements with plain old script tags (cc @castastrophe @starryeyez024)!

fixes #107 
fixes #121

Relates to #80 since this PR includes a rollup config factory in `scripts/rollup.config.factory.js`.  Instead of having a duplicate rollup config in every element dir, each element's `rollup.config.js` just imports the factory.


elements/all/test/index.html
```diff
@@ -9,8 +9,11 @@
     <script>
       // Load and run all tests (.html, .js):
       WCT.loadSuites([
-        'fouc-test.html'
+        'fouc-test.html',
+        // 'module-amd-test.html',
```
kylebuch8: @mwcz - Do we want to leave these commented out?

elements/rh-button/demo/index.html
```diff
@@ -9,8 +9,9 @@
     <!-- uncomment the es5-adapter if you're using the umd version -->
     <script src="/@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js"></script>
     <script src="/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.5/require.min.js"></script>
-    <script>require(['../rh-button.umd.js'])</script>
+
+    <script src='../../rhelement/rhelement.umd.js'></script>
```
kylebuch8: @mwcz - I assume that this is the pattern we want developers to follow?

elements/rh-button/src/rh-button.js
```diff
@@ -37,7 +37,10 @@ class RhButton extends RHElement {
 
   constructor() {
     super(RhButton.tag);
+    console.log("creating rh-button");
```
kylebuch8: @mwcz - Can you remove this console.log statement?

elements/all/test/index.html
mwcz: Dang, no, they shouldn't be.  Those commented test files aren't written yet though, I just stubbed out the filenames.  I should finish those before merging this PR.

elements/rh-button/demo/index.html
mwcz: I don't think we should express a preference.  We're offering ESM, AMD, CJS, and globals so developers can choose the module system that's best for their project.

elements/rh-button/src/rh-button.js
mwcz: Don't you mean *may* I remove this console.log statement?

kylebuch8: Yes...

mwcz:
@kylebuch8 I added tests for ESM, AMD, and globals.  They import rh-card via the respective method, and make sure it upgrades.  Perhaps we can think of more things to test in the future, but I think it's good for now.

elements/rh-button/package.json
```diff
@@ -8,7 +12,7 @@
   },
   "main": "index.js",
   "scripts": {
-    "build": "../../node_modules/.bin/gulp",
+    "build": "../../node_modules/.bin/gulp && ../../node_modules/.bin/rollup -c && ../../node_modules/.bin/prettier --ignore-path ../../.prettierignore --write '**/*.{js,json}'",
```
mwcz: @kylebuch8 This was a bit of a quick fix that we should revisit.  Ideally rollup-ing and prettier-ing would be done in the gulpfile, but it would require a larger refactoring of the gulpfile than I was comfortable with in an already-large PR.

PR #123, opened 2018-08-21
DEMO: RH-CTA fork 2.0

castastrophe:
A fork showing an example of the simplified map approach


kylebuch8:
@castastrophe Can we close this?

PR #128, opened 2018-08-23
Add rollup task to all gulpfiles in elements

castastrophe:
@mwcz I took a stab at including the roll-up task to all the build files for the individual elements to fix the defect I found in running `npm run dev` which was not re-bundling on watch.


kylebuch8:
@mwcz Can you review this when you get a chance?

mwcz:
I added one change, removing `rollup -c` from the npm build script.  With rollup entirely in gulp now, the "outer" npm one isn't needed.  Woot!

PR #131, opened 2018-08-29
bug-fix when user don't have web components pollyfill, gets an error

Homa:
We need web components polyfill only when we want to cover IE11. On revel.js, we need first check if user has web components or not.


mwcz:
@Homa Thanks for the PR!  The polyfill is needed for IE11 for sure, but also for Firefox.  At the moment, RHElements requires the polyfill in all cases.  Not forever, but for now we need it.

With only a few small changes, this PR could `console.warn` users who forget to include the polyfill.  I think that would be a great feature to include.  @Homa  would you like to make those changes?

Homa:
Added the message...

Homa:
updated the branch with the requested change...

kylebuch8:
@mwcz One thought here. I don't like the idea of making the use of the polyfill required. Eventually the polyfill will go away and reveal.js will be checking for something that's not needed. I know that future is a ways off. It's just something that I want us to be thinking about.

mwcz:
@kylebuch8 I thought you and I had established the polyfill was required, whether we liked it or not.  But it can be required by necessity without being required by code dependency.  That's what @Homa said when we talked about this issue last Wednesday, she said that however we fix this issue, it should allow the polyfill to be lifted out someday without any side effects or cleanup commits.  Or even lifted out _today_ by a RHElements user who only needs to support Chrome, for example.  She convinced me, I think the warning should either...

- be removed, or
- be behind a debug flag

PR #147, opened 2018-09-07
cp-theme is in a good state for now...

markcaron:
Besides cleaning up red-hat-sass, we'll need to change some of the vars and decide on some final CP colors, but for now, it's ready to go!


elements/rh-cta/demo/index.html
```diff
@@ -37,12 +38,12 @@
     </style>
     <style>
     /* For demoing theme styles... */
-    :root {
+    /* :root {
```
mwcz: I think this would be great in the README for this theme

markcaron:
@mwcz Best LGTM ever.

PR #159, opened 2018-09-12
Add text knob

TylerMartin5:
Wondering if all of the CTAs should change or just the top and markup


TylerMartin5:
@kylebuch8 I think it makes more sense to have the cta priorities as the text like mark mentioned. It's just easier for users to take in.

kylebuch8:
@TylerMartin5 Ok that works for me. I'll merge it once the tests pass.

PR #172, opened 2018-09-20
Moving old vars to deprecated SCSS files, simplifying container-spacer...

markcaron:
- Deprecating vars and colors
- Simplifying spacers
- Adding UI borders to _maps
- Cleaning up RH-CTA and RH-Card to now be more themeable


kylebuch8:
@markcaron We have conflicts that we need to take care of. Could you take a look?

PR #174, opened 2018-09-21
Rh accordion

markcaron:
Add theming.


kylebuch8:
@markcaron - Can you update the README with the themeable styles?

PR #189, opened 2018-09-27
Rh tabs

markcaron:
Made tabs themeable!


elements/rh-tabs/.npmignore
```diff
@@ -0,0 +1 @@
+node_modules
```
mwcz: Isn't node_modules ignored by default?

mwcz: Just checked, nothing is ignored by default, but gitignore is used if npmignore is not present.  I think we should remove npmignore from all the elements (some have it, not all) and let gitignore take over, unless there's a distinct reason we need both?

PR #217, opened 2018-10-16
add clean script

mwcz:
This adds a `npm run clean` script that removes all package-lock files and node_modules dirs.


mwcz:
Lovely Gourmet Tofu Meal!

Issue #89, opened 2018-07-28
DOC | JSdoc auto-documentation


mwcz:
Set up jsdoc (probably) documentation.

castastrophe:
Guide: https://github.com/jsdoc/jsdoc
Gulp: https://github.com/deshaw/gulp-jsdoc3

starryeyez024:
Hey folks, can y'all elaborate (1 or 2 bullet points) about why this is needed, how it's awesome, etc. Thanks!

castastrophe:
The JSDoc plugin will provide an in-code way to document public-facing API in components and all the available methods for PFElement.  This can reduce duplication and improve understanding of what the base class brings to the table.

heyMP:
👍   

It can also improve the developer experience since VSCode can use JSDoc comments to populate its autocomplete, similar to typescript.
https://dev.to/henryjw/better-javascript-type-autocomplete-with-jsdoc-3bdo

And we could convert JSdoc to markdown to potentially auto generate docs.
https://dev.to/henryjw/better-javascript-type-autocomplete-with-jsdoc-3bdo

PR #222, opened 2018-10-18
pfe-avatar

### LINKED ISSUE #214: Avatar component
- [x] use CSS variables for the color
- [x] hash name attribute
- [x] draw pattern to canvas based on hash number
- [x] display initials
- [x] decide on canvas vs svg
- [x] allow using a provided image instead of generated pattern
- [x] for displaying the provided image, decide on `` vs creating a `<rh-img>`
- [x] allow square, rounded square, and circular display
- [x] decide which CSS vars to use, and how to customize them
- [x] come up with a set of 5 or 6 colors (I just used a few colors I saw at brand.redhat.com)
- [x] implement splitting of `--rh-avatar--colors` values (space delim) and using name hash to pick which color (with mod)
- [x] add HSL lib and darken selected color (or lighten, if the color is too close to black)
- [x] figure out sizing of the element and the text within (skipping this in favor of a broader solution related to #238)
- [x] add markup preview to storybook
- [x] write rh-avatar's README
- [x] determine how to choose initials (maybe 1, 2, or 3 instead of just 1 or 2)
- [x] use squares on the Portal 
- [x] write documentation

mwcz:
First stab, much to discuss.

fixes #214


kylebuch8:
Couple of thoughts...

1. If we already have an image for someone's avatar, we need to use the image instead of the autogenerated avatar.
2. I think we need to provide a way to allow for a circular or square avatar or a square avatar with rounded corners.
3. If we are using an image for the avatar, it might make sense to have an image component similar to iron-image so we can use that in the component.

starryeyez024:
is the purpose of this to create generated avatars for people who do not have photos?

mwcz:
@starryeyez024 Very good point, I added a description to the README. :)

elements/pfe-avatar/src/pfe-avatar.js
```diff
@@ -0,0 +1,352 @@
+/*
```
castastrophe: @mwcz You can delete the copyright from this file now.

elements/pfe-avatar/src/pfe-avatar.js
```diff
@@ -0,0 +1,352 @@
...
+      squares: "squares"
+    };
+  }
+
+  static get defaultColors() {
+    return "#3B0083 #f0ab00 #007a87 #00b9e4 #92d400";
+  }
+
+  get name() {
+    return this.getAttribute("name");
```
castastrophe: @mwcz Can you update the attribute names to be pfe prefixed?

elements/pfe-avatar/src/pfe-avatar.js
```diff
@@ -0,0 +1,352 @@
...
+      new CustomEvent(`${PfeAvatar.tag}:connected`, {
+        bubbles: false
+      })
+    );
+  }
+
+  attributeChangedCallback(attr, oldValue, newValue) {
+    super.attributeChangedCallback(...arguments);
+
+    switch (attr) {
```
castastrophe: If these all break without running anything, what's the point of this switch statement?

elements/pfe-avatar/src/pfe-avatar.js
```diff
@@ -0,0 +1,352 @@
...
+    this._squareSize = this._canvas.width / 8;
+    this._triangleSize = this._canvas.width / 4;
+
+    this._ctx = this._canvas.getContext("2d");
+  }
+
+  static _registerColors() {
+    this.colors = [];
+    const themeColors =
+      PFElement.var("--pfe-avatar--colors") || this.defaultColors;
```
castastrophe: Should this get the variable from PFElement or would it be more accurate to get it from this context since the variable is defined on the pfe-avatar host?

elements/pfe-number/pfe-number.js
```diff
@@ -4,23 +4,23 @@ var commonjsGlobal =
   typeof window !== "undefined"
     ? window
     : typeof global !== "undefined"
-      ? global
-      : typeof self !== "undefined"
-        ? self
-        : {};
+    ? global
```
castastrophe: @mwcz Do you want to pull this out of this PR since it's just prettifying a file that was not edited as part of this project?

elements/pfelement/src/pfelement.js
```diff
@@ -93,6 +96,10 @@ class PFElement extends HTMLElement {
     }
   }
 
+  disconnectedCallback() {
+    this.connected = false;
```
castastrophe: I love this!

elements/pfe-avatar/src/pfe-avatar.js
castastrophe: Okay, I see this is calling a function that is declared inside of PFElement.  Does this extending PFElement mean that we could do `this.var` and it would still reference `PFElement.var`?  It might be nice to do that in general if it is the same so that if PFElement name changes again (here's hoping it doesn't), we only need to update that once in the file instead of multiple times.

elements/pfe-avatar/src/pfe-avatar.js
```diff
@@ -0,0 +1,352 @@
...
+  _clear() {
+    this._ctx.clearRect(0, 0, this._canvas.width, this._canvas.height);
+  }
+
+  _drawBackground() {
+    this._ctx.fillStyle = this.color1;
+    this._ctx.fillRect(0, 0, this._canvas.width, this._canvas.height);
+  }
+
+  _drawSquarePattern(pattern) {
```
castastrophe: If you have time, can you add a short comment above each of these functions briefly describing what they do?

elements/pfe-avatar/src/pfe-avatar.js
```diff
@@ -0,0 +1,352 @@
...
+    }
+    gradient.addColorStop(0, gradientColor1);
+    gradient.addColorStop(1, gradientColor2);
+    gradient.addColorStop(1, gradientColor1);
+    this._ctx.fillStyle = gradient;
+    this._ctx.fillRect(0, 0, this._canvas.width, this._canvas.height);
+  }
+}
+
+PfeAvatar._registerColors();
```
castastrophe: Curious why do this outside of the element declaration?

elements/pfe-avatar/src/pfe-avatar.js
mwcz: Okay. :(

elements/pfe-avatar/src/pfe-avatar.js
mwcz: Just a placeholder I forgot to remove.

elements/pfe-avatar/src/pfe-avatar.js
mwcz: Sure, that should work but it needs testing.

elements/pfe-number/pfe-number.js
mwcz: Sure.

elements/pfe-avatar/src/pfe-avatar.js
mwcz: It's tied to the class, not an individual element, so it can't go in the constructor or connectedCallback.  It needs to be called only once and applies to all pfe-avatars on the page.

elements/pfe-avatar/src/pfe-avatar.js
castastrophe: 👍 Just let me know if you'd like to defer this and I can open an issue to address it later

elements/pfe-avatar/src/pfe-avatar.js
castastrophe: Oh very clever!  Minor question, since the function is designed to be run from outside the element's definitions, maybe it shouldn't be prefixed with the underscore?  I understood the underscore to be scoped to internal-only functions.

mwcz: It isn't intended to be run from outside this file, hence the underscore.

elements/pfe-avatar/src/pfe-avatar.js
markcaron: @kylebuch8 and I discovered this same issue. I think we should defer this so we can get some sort of MVP out soon.

castastrophe: I opened this issue: https://github.com/patternfly/patternfly-elements/issues/297

elements/pfe-avatar/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+<pfe-avatar name="Blanca Rohloff" show-initials pattern="triangles"></pfe-avatar>
+
+<pfe-avatar name="Edwardo Lindsey" src="https://clayto.com/2014/03/rgb-webgl-color-cube/colorcube.jpg"></pfe-avatar>
+```
+
+## Attributes
+
+### name (observed) (required)
+
```
castastrophe: @mwcz Should we remove the last sentence from the doc about the name being used to generate displayed initials?

elements/pfe-avatar/README.md
```diff
@@ -0,0 +1,69 @@
...
+| squares                         | triangles                           |
+| ------------------------------- | ----------------------------------- |
+| ![squares image](./squares.png) | ![triangles image](./triangles.png) |
+
+### shape (observed)
+
+The shape of the avatar itself. Supported values are `"square"` (default), `"rounded"`, and `"circle"`.
+
+### show-initials
```
castastrophe: @mwcz Should this be removed?

elements/pfe-avatar/src/pfe-avatar.js
mwcz: Kyle and Mark just noticed the same thing, and we looked at it closer and `this.var` won't actually work here like I thought.  Because `_registerColors` is a static function, within it `this` refers to the class `PfeAvatar`, not a `<pfe-avatar>` element.

I think these are our choices:

1. get the colors from `:root` once and apply them to all avatars (simple, efficient, less flexible)
2. get the colors for each avatar on the page (flexbile, but slow since `getComputedStyles()` is slow)

elements/pfe-avatar/src/pfe-avatar.js
mwcz: :+1:

elements/pfe-avatar/src/pfe-avatar.js
mwcz: Sure, I'll need to do that in a later sprint but definitely.  I have a (fairly old) branch which adds jsdoc-powered documentation which lines up nicely with this request.

mwcz: The issue for that branch is #89

elements/pfe-avatar/README.md
mwcz: Yes! :+1:

elements/pfe-avatar/README.md
mwcz: Yes! :paraguay:

PR #234, opened 2018-10-31
Make RHElements compatible with RequireJS

mwcz:
When loading modules, RequireJS treats paths differently based on whether they end in `.js`.  For example `./foo.js` will resolve to a different path from `./foo`.  This PR removes the `.js` from AMD/CJS dependency paths in RHElements' the UMD builds.  For example, `require("../rhelement/rhelement.umd.js")` becomes `require("../rhelement/rhelement.umd")`.  AMD and CJS both support omitting the `.js`.


elements/rh-cta/src/rh-cta.scss
```diff
@@ -72,7 +72,7 @@ $rh-cta__arrow-size:                        0.313em;
     padding: calc(#{rh-var(container-padding)} * 0.5) calc(#{rh-var(container-padding)} * 2);
     border-radius: calc(#{rh-var(ui--border-radius)} * 20);
     border: #{rh-var(ui--border-width)} #{rh-var(ui--border-style)} transparent;
-    text-decoration: none;
```
castastrophe: @markcaron I'd love to avoid using !important in situations like these.  It makes it difficult to override when that is desirable.

mwcz:
@kylebuch8 Any remaining qualms about merging this in?

mwcz:
@kylebuch8 maybe if I give you an LGTM?

PR #242, opened 2018-11-13
update accordion story

TylerMartin5:


castastrophe:
@TylerMartin5 These updates look good; would you want to incorporate any of the approaches from the hide-show branch Chris is working on here?  We added a tool to generate random lorem ipsum which is really nice. https://github.com/patternfly/patternfly-elements/pull/259/files#diff-aa8cccdbf1a385f803895221e44c14d3

castastrophe:
@mwcz I'd love if we could hold off merging this until this is merged https://github.com/patternfly/patternfly-elements/pull/268 as it would cause a big diff to resolve.

PR #255, opened 2018-12-04
Fix npm start

mwcz:
Restores the doc/index.html page to how @castastrophe had it; generated when `npm start` is run and not tracked in git.  It's also generated when a new element is created, so bouncing `npm start` isn't needed.

I also removed `unresolved` so that the index page will display right away.  If we ever use RHElements on the listing page, we can re-add it.


castastrophe:
The plus-ing-est of ones 👍

PR #257, opened 2018-12-06
Abstract copyright statement

castastrophe:
Moving the copyright statement out of the <element>/src/*.js files and using the LICENSE.txt that exists in each folder to auto-populate that string at the top of each compiled JS file.

Also added a way to push out an updated LICENSE from the root down to the children for copyright updates.


castastrophe:
@kylebuch8 We can run this by legal to confirm but I suspect this would be okay with them as it enables us to make global updates to the copyright more effectively.  If you share the contact information with me of our legal rep on this issue via my Red Hat e-mail, I'm happy to handle the conversation too.

kylebuch8:
@castastrophe - I updated all of the 2018s to 2019s and rebuilt the files.

PR #258, opened 2018-12-06
Moving the element compilation to a dist directory

castastrophe:
This is my proposal for how we might move to a dist directory compilation structure.  For development, the dist directory is not committed but it is force committed during tag releases.  This structure makes it much easier for consumers of the repo to import relevant files and not build tool/ancillary files that are not relevant.


castastrophe:
Closing this pull request for now while I collaborate more with @mwcz on best practices for pulling the repo into our project.  In the meantime, I added the compiled assets to the gitignore for the monorepo which will help with some of the headaches.

PR #264, opened 2018-12-12
Pfe

kylebuch8:
Switching elements from the RHElement and rh- namespace to PFElement and pfe- namespace.


.storybook/config.js
```diff
@@ -3,7 +3,7 @@ import { setOptions } from "@storybook/addon-options";
 const req = require.context("../elements", true, /\.story\.js$/);
 
 setOptions({
-  name: "RHElements",
+  name: "PFElements",
```
castastrophe: PF Elements or PFElements? Space or no-space?

mwcz:
I'm not able to `npm install`, it fails with:

```
lerna ERR! npm install exited 1 in '@rhelements/rhelements'  
lerna ERR! npm install stderr:                               
npm ERR! code E404                                           
npm ERR! 404 Not Found: @pfelements/pfe-sass@*               
```

It seems like `@pfelements/pfe-sass` can't be found on npm, but I wouldn't have expected that to be a problem because the local pfe-sass should be linked up and used instead.

.storybook/config.js
kylebuch8: It should be PatternFly Elements.

kylebuch8: I fix it.

mwcz:
Here goes something.

PR #265, opened 2018-12-17
US162827-band-layout-CSS-fork 

starryeyez024:
Added in cp-theme, added some helper text to the demo file to show what layouts are being applied; Adjusted CSS relating to padding and width for the band wrapper, container, & aside regions


elements/rh-band/demo/index.html
```diff
@@ -17,21 +17,28 @@
     <script>require([
       '../rh-band.umd.js',
       '../../rh-card/rh-card.umd.js',
-      '../../rh-cta/rh-cta.umd.js'
+      '../../rh-cta/rh-cta.umd.js',
+      '../../../themes/cp-theme/cp-theme.umd.js',
     ])</script>
+    <style>
+      .pfe-band-summary {
```
castastrophe: Even though this is just a demo, I think I'd prefer if classes that are not added by the component's template don't use the pfe prefix.

elements/rh-band/src/rh-band.scss
```diff
@@ -5,12 +5,15 @@ $SLOT: "";
...
+
+
 :host {
-  // Spacing values
-  --rh-band--paddingTop:                      calc(#{rh-var(container-spacer)} * 4);
-  --rh-band--paddingRight:                    #{rh-var(container-spacer)};
-  --rh-band--paddingBottom:                   calc(#{rh-var(container-spacer)} * 4);
-  --rh-band--paddingLeft:                     #{rh-var(container-spacer)};
+  // Default spacing values
+  --rh-band--padding-vertical:                 calc( #{rh-var(container-spacer)} * 4);
```
castastrophe: I think this is a case where we would use the underscores for the naming: `--rh-band--padding_vertical`

elements/rh-band/demo/index.html
castastrophe: That way, developers looking at the demo page don't think they should be doing the same for their custom classes

elements/rh-band/src/rh-band.scss
```diff
@@ -5,12 +5,15 @@ $SLOT: "";
...
+  --rh-band--padding-horizontal:               calc( #{rh-var(container-spacer)} * 1);
+  // Bring them together for ease of use below
+  --rh-band--padding-full:                         #{rh-local(padding-vertical)}  #{rh-local(padding-horizontal)};
```
castastrophe: I think for clarity, this should be: `--rh-band--padding` since padding is the name of property this is being assigned to.

elements/rh-band/src/rh-band.scss
```diff
@@ -93,14 +103,13 @@ $aside-lg: 300px;
 
     // Remove background color for print
     @include rh-no-print-background;
+
+
   }
   &__container {
-    padding-top:    0;
```
castastrophe: Why move the padding off the container?

elements/rh-band/src/rh-band.scss
```diff
@@ -37,15 +40,16 @@ $aside-lg: 300px;
   --rh-band--width:                           auto;
 
   // Loop through the other breakpoints
-  @each $size in (sm, md, lg, xl) {
+  @each $size in (md, lg, xl) {
     @media screen and (min-width: rh-breakpoint(#{$size})) {
-      --rh-band--width: calc(#{rh-breakpoint(#{$size})} - calc(#{rh-local(paddingRight)} * 2));
+
+      --rh-band--width: calc( #{rh-breakpoint(#{$size})} - calc(#{rh-local(padding-horizontal)} * 4) );
```
castastrophe: Why x4?

elements/rh-band/src/rh-band.scss
castastrophe: The template is designed for the padding to go on the container element and not the wrapper. The wrapper is for the background and image settings (overlays, etc.)

elements/rh-band/demo/index.html
starryeyez024: sure thing

elements/rh-band/src/rh-band.scss
starryeyez024: sure, updated

elements/rh-band/src/rh-band.scss
```diff
@@ -113,96 +126,96 @@ $aside-lg: 300px;
...
-    }
-    [aside~="left"] & {
-      order: -1;
+    @media screen and (min-width: #{rh-breakpoint(md)}) and (max-width: #{rh-breakpoint(lg)}) {
+      --rh-band_aside--layout: #{$aside-sm}; // set limiter
     }
-    [aside~="right"] & {
-      order: 2;
+  }
+  &__aside {  // walk away from aside SLOT var to invoke gutter
```
starryeyez024: This still might need some help. The only way I could get the SLOT vars to work was to make them global, but then you have to keep overriding them

elements/rh-band/src/rh-band.scss
castastrophe: +1 and I love that table!  That's going in the README!

elements/rh-band/src/rh-band.scss
castastrophe: I can help with that.  They just need to be updated before the function is run.

elements/rh-band/src/rh-band.scss
castastrophe: +1

elements/rh-band/src/rh-band.scss
castastrophe: +1

PR #267, opened 2018-12-17
Add compiled assets to the gitignore

castastrophe:
Compiled assets cause a lot of confusing diffs for users so I'm recommending we add them to the .gitignore.


starryeyez024:
@castastrophe when i look at http://localhost:1234/elements/pfe-card/demo/ the cards have no bg colors

castastrophe:
> @castastrophe when i look at http://localhost:1234/elements/pfe-card/demo/ the cards have no bg colors

What commands did you run to spin up the host?  `npm run dev` and `npm run start`?

starryeyez024:
> > @castastrophe when i look at http://localhost:1234/elements/pfe-card/demo/ the cards have no bg colors
> 
> What commands did you run to spin up the host? `npm run dev` and `npm run start`?

@castastrophe yep, in the pfe-card and root directories respectively

mwcz:
I added a few updates:

 - removed my-cool-element from test/index.html
 - added themes to .gitignore
 - updated js ignore glob (was `.js*`) to not ignore package.json and package-lock.json
 - stopped ignoring gulpfile and rollup files.  the files were already tracked so no big deal, until we add a new element, at which point its new gulpfile and rollup file would be automatically ignored unless we force add them.
 - moved story files from `src/` to `demo/`.  I didn't want to do that in this branch, but it fit so nicely into these changes because the builds copy all js files from src into the elements dir, which meant the story files were being copied.  moving them into demo was a very convenient way to stop that from happening and it's a change I think we're all cool with.

.gitignore
```diff
@@ -2,6 +2,21 @@
...
+# ignore js, map, and css in element dirs
+elements/*/*.js
+elements/*/*.map
+elements/*/*.css
+# but don't ignore tooling files
+!elements/*/gulpfile.js
+!elements/*/rollup.config.js
+
+# and same for themes
+themes/*/*.js
```
castastrophe: @mwcz Can we set the entire themes repo to ignore for now since it's being deprecated anyway?

scripts/gulpfile.factory.js
```diff
@@ -18,7 +18,7 @@ module.exports = function factory({
 
   gulp.task("merge", () => {
     return gulp
-      .src(`./src/**/*.js`)
+      .src([`./src/**/*.js`, `!./src/*.story.js`])
```
castastrophe: @mwcz Since you moved story into the demo directory, is this still needed?

.gitignore
mwcz: I think that should happen in a dedicated PR, one dedicated to deprecating `themes`.

scripts/gulpfile.factory.js
mwcz: Good catch!

PR #268, opened 2019-01-02
Storybook updates

castastrophe:
I created a method of more dynamically building storybook instances.  If a schema is defined for the element, it is the easiest but the properties can be manually defined as well.

The goal of this enhancement is to speed up and make easier the theming process.  If an element can be quickly spun up with the right markup in storybook, we can start styling it much faster.

Eventually I think these tools can help us build a dynamic demo and testing framework as well.


castastrophe:
The storybook utils.js are the crux of the PR.

elements/pfe-card/src/pfe-card.story.js
```diff
@@ -0,0 +1,266 @@
...
+    html => (rendered = html)
+  );
+
+  return `${rendered}
+<pre style="white-space: pre-wrap; padding: 20px 50px; background-color: #f0f0f0; font-weight: bold;border: 1px solid #bccc;">
+${tools.escapeHTML(rendered)}
+</pre>
+`;
+
+  //   valueAttr =
```
kylebuch8: @castastrophe Do we want to remove all of these comments?

elements/pfe-tabs/src/pfe-tabs.story.js
```diff
@@ -0,0 +1,150 @@
...
+  return tools.component(PfeTabs.tag, data.prop, data.slots);
+};
+
+stories.addDecorator(storybookBridge.withKnobs);
+
+stories.add(PfeTabs.tag, () => {
+  let config = {};
+  // const props = PfeTabs.properties;
+  // Manually defining props but this can be done in a schema instead
+  const props = {
```
kylebuch8: @castastrophe - It looks like we don't have an orientation attribute on tabs. I think it's just a matter of adding a vertical attribute.

https://github.com/patternfly/patternfly-elements/blob/master/elements/pfe-tabs/pfe-tabs.js#L247

elements/pfe-accordion/src/pfe-accordion.story.js
```diff
@@ -0,0 +1,93 @@
...
+    {
+      indent: "    ",
+      "remove-attributes": [],
+      wrap: 0
+    },
+    html => (rendered = html)
+  );
+
+  return `${rendered}
+<pre style="white-space: pre-wrap; padding: 20px 50px; background-color: #f0f0f0; font-weight: bold;border: 1px solid #bccc;">
```
kylebuch8: @castastrophe - Would we want to add an additional stylesheet to the preview pane head instead of doing inline styles?

https://storybook.js.org/configurations/add-custom-head-tags/

elements/pfe-card/src/pfe-card.story.js
castastrophe: Yes, I'll do that now

elements/pfe-accordion/src/pfe-accordion.story.js
castastrophe: It might be excessive to create a stylesheet just for this short set of styles.  How about I create a single-source of truth instead via a preview function in the utils that each story would call?

elements/pfe-card/src/pfe-card.story.js
castastrophe: bc9262f1e732cfa23cd8f6f82129f6a475b62d41

elements/pfe-accordion/src/pfe-accordion.story.js
castastrophe: bc9262f1e732cfa23cd8f6f82129f6a475b62d41

elements/pfe-tabs/src/pfe-tabs.story.js
castastrophe: There are styles based on a vertical boolean attribute so: `<pfe-tabs vertical>`

castastrophe: Oh!  I see what you're saying; I just pushed up a fix for this 077a9477765be2d81e97e038dcfae2d5d3e7b77b

PR #269, opened 2019-01-03
example integration with HAX

btopro:
It's not amazing integration but I wanted to post this up here to at least create a working example. This is more or less what the wcfactory tooling kicks out as far as definitions. I like your emphasis on slot usage, I'll have to expand support for slots in the spec a bit more possibly but here's some screen shots showing this in context of the h-a-x tag I've been working on.

Searching for Card
Searching for Card
Simple form from the JSON Schema
Simple form from the JSON Schema

Not intending this to be accepted, just figured a PR would be a good way to generate a conversation on this one.


elements/pfe-card/src/pfe-card.js
```diff
@@ -38,6 +39,67 @@ class PfeCard extends PFElement {
   constructor() {
     super(PfeCard);
   }
+
+  connectedCallback() {
+    super.connectedCallback();
+    let haxProps = {
```
castastrophe: @btopro I'm hesitant to build HAX into the components directly because it seems like each CMS will have their own implementation requirements that might vary.  @mwcz was talking about adding a hook into the connectedCallback to allow things to be attached separately; that means, in theory, we could build out a HAX implementation for each CMS separately and hook them into the components in post.

castastrophe: @starryeyez024 What do you think of this idea?  RH.com could have a HAX JS file that is a compilation of our CMS definitions for each component.  CP could have their own, etc.  So we'd include:
```
- pfelement.js
- pfe-card.js
- rdc-hax.js
```

btopro: Another possible solution would be another custom element that you pile the definitions into and if THAT tag is appended then it fires the correct events to supply the CMS w/ those definitions that make the most sense for it. [Custom element registry](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/get) could be used to test that a tag's been defined and then broadcast as another way of keeping things tidy / abstracted per implementation

This PR was mostly so our talk has some fodder of "look what happens when we all adopt the platform" to ensure this isn't accepted accidentally :)

btopro: @starryeyez024 thank you for the context. Speaking to the options B vs C which could provide a nice happy medium in flexibility vs constraints...

A paradigm that @nikkimk has illustrated w/ [a11y-media-player](https://www.webcomponents.org/element/@lrnwebcomponents/a11y-media-player) is that she's made a pretty ridiculously flexible / super-power laden video player. It's WAAAYYYY too much horsepower for a typical user that would be using a UI to implement. So, we use that element in our templates that are delivered by our media system (Drupal based). This allows us to wire up highly complex features of the player from our content type / fields, leveraging slots etc.

For end users / HAX / content blob areas they get the [video-player tag](https://github.com/elmsln/lrnwebcomponents/blob/master/elements/video-player/video-player.js#L63-L93). The `video-player` tag then leverages a11y-media-player so we get design consistency but we allow HAX and the wrapper tag to provide the constrains needed to make it manageable for end-users.

A happy medium toward the B/C option could also be making semantically named tags (splash-card instead of pfe-card sorta deal) that have the HAX wiring in them while the raw tags currently in patternfly are the most flexible for other devs to implement consistency in design / theme layer.

mwcz:
It seems to me that `connectedCallback` may not be the right place to do the HAX wiring.  Is the wiring something that should happen every time a custom element is added to the DOM, or moved around within the DOM?  Or, can it happen just once, right after `pfe-card.js` (for example) is included?  I'm really not sure how this CMS UI builder stuff works so I await enlightenment. :)

btopro:
@mwcz That's just the way we have it in our examples. It's possible to have a dedicated element that fires the meta-data it's just hyper modular the way we're illustrating it in demos / samples. How it works at a high level:
- h-a-x is a web component tag that wraps a block of HTML text as it's "body" area to work against
- once the `hax-store` tag registers the `hax-store-ready` event, things are safe to fire off and be registered. HAX has 0 concept of what tags it can edit beyond p,iframe,a,img (primitives)
- other tags definitions get appended to the DOM and fire an event up of "hey HAX here's how to modify me"
- hax-store captures this schema firing (which an event library included in each element has invoked to standardize the firing) and this builds the definitions as to how it knows what to edit / allow placement on the UI.
- User clicks on a `video-player` tag in the DOM (that's inside the `h-a-x` tag slot) and hax knows from the meta-data to present certain UI options / that that is a valid tag)

As the comments above have suggested though abstraction of the element schema from the elements themselves would safeguard against theme / template level logic utilizing those same components yet those tags maybe not being what you want end-users using the GUI to have access to modify / add anew instance of.


h-a-x implements a "app-store" specification which helps with auto-configuring all the advanced integrations (and auto-loading element definitions on demand which appends to the UI then deletes the tag just to have got the meta-data). We could modify the listener event attachments to ensure that ONLY the things placed in this auto-loaded area are the things that get definitions added.

castastrophe:
Hey @btopro, was chatting with @starryeyez024 and I think there's a missing layer here between the web components and the CMS.  I wrote up a google doc with some more details and notes (if you request access to this doc I can grant it for external e-mails one-off):

https://docs.google.com/document/d/1js5tBYUG8UnbZJlp1cn2q6fV_b4lWvFhGLtIFiY4Jk4/edit?usp=sharing

The gist is, I think we need a CMS-builder experience and I think HAX can help.

btopro:
> Hey @btopro, was chatting with @starryeyez024 and I think there's a missing layer here between the web components and the CMS. I wrote up a google doc with some more details and notes (if you request access to this doc I can grant it for external e-mails one-off):
> 
> https://docs.google.com/document/d/1js5tBYUG8UnbZJlp1cn2q6fV_b4lWvFhGLtIFiY4Jk4/edit?usp=sharing
> 
> The gist is, I think we need a CMS-builder experience and I think HAX can help.

I requested access and I agree. We abstracted HAX Schema from JSON Schema in part bc of this but I'm excited to see your take on what this all would imply in said document :)

elements/pfe-card/src/pfe-card.js
starryeyez024: @castastrophe @mwcz could we include the haxProps in the connected callback of each of our content & container components? The combos might be the only ones that need site-specific definitions for the fields.

btopro: as an FYI bc of this thread I've recently enhanced the way we can detect if something is HAX capable -- https://github.com/elmsln/lrnwebcomponents/blob/master/elements/hax-body/lib/hax-store.js#L479-L492

english translation - If one of the 3 common methods of implementing HAX into an element is detected, load the definition without needing to fire the connected callback. Fallback on the connected callback for older elements.

This was in part for performance as without having to connect it to the DOM we can get the definitions faster, especially as we move closer to dynamic imports for all with no polyfill or compiling, it'll be lightning compared to having to connect and disconnect every asset at run time.

This code is running on haxtheweb.org and has reduced the time to boot up HAX as a result

mwcz: I'm for adding HAX-compatibility as a "plugin" to each PF Element, but I don't want to include HAX properties in the original source of the elements.  I implemented #270 so that you can add onConnectedCallback-type stuff to an element without modifying its original source.  Since connecting to the DOM isn't required anymore, we can do something like these steps:

 1. include `pfe-card.js`
 2. include `pfe-card.hax.js` which adds haxProperties to the PfeCard class
 3. start creating `<pfe-card>` elements

mwcz:
I'm keeping this bookmarked for future reference but closing the PR itself.  Thank you Bryan!

PR #270, opened 2019-01-07
add 'connected' DOM event

mwcz:
This adds a custom "connected" event that gets fired by connectedCallback.  It replaces the custom queueAction functionality for adding `pfe-type` attributes with standard `addEventListener` and `CustomEvent`, and also opens up a new way of scheduling things to take place when an element is connected.

Things to discuss:

 1. Should the "connected" event bubble up?  I am leaning towards no.
     a. But if we decide it should bubble, should it bubble across light/shadow DOM boundaries?
 3. About registering "connected" event listeners from *outside* the element's own source... the timing is tricky.  In general, event listeners do not get run if they are registered (`addEventListener("connected", callback)`) after the event has fired.  To  so they must be registered before the "connected" event fires.  In RequireJS/AMD mode this is difficult, but in global or ESM mode it's much more predictable.


elements/pfelement/src/pfelement.js
```diff
@@ -79,13 +86,24 @@ class PFElement extends HTMLElement {
...
       this._mapSchemaToProperties(this.tag, this.props);
     }
 
     if (this._queue.length) {
       this._processQueue();
     }
+
+    // if there are any a element-level connected listeners, trigger them
+    this.dispatchEvent(
+      new CustomEvent(`connected`, {
```
castastrophe: Could this event be `${this.tag}:connected`?

mwcz:
This will probably be reopened in the future when the need is more pressing.

PR #256, opened 2018-12-06
add a new dev script at the root

mwcz:
This adds a `npm run dev` script to the root of the repo.  Prior to this, `npm run dev` only existed inside each element directory.  This new script will watch every element by default, and offers a very easy way to watch only the element you care about.

A companion PR (https://github.com/RHElements/rhelements-documentation/pull/12) is open for updating the documentation, and should be merged whenever this PR is merged.  Follow [this link](https://github.com/RHElements/rhelements-documentation/blob/ab1c682dde86682f35fb36532fa64b40d0d828dc/content/develop/step-1a.md#develop) to see the docs for this feature.

PR #274, opened 2019-01-07
add easy syntax for building/testing certain elements

mwcz:
This extends #256 to `test` and `build`.  Allows running builds and tests like:

```
npm run build pfe-card pfe-cta
```

Or

```
npm test pfe-datetime
```


castastrophe:
I also tested `npm run build` and `npm test` and they both run the full suite by default +1

PR #277, opened 2019-01-09
Pt 2: Storybook updates

castastrophe:
This branch updates the storybook features to be a little smarter - adds defaults and required field support.


castastrophe:
This branch updates:

- pfe-card: Since "standard" is the default value, only "small" is a valid attribute value.  Updates the dropdown to null or "small"
- pfe-cta: Removes default value from dropdown and allows "null" to fill in for that instead.
- pfe-datetime: Sets up the required value on datetime and type and removes defaults for non-required fields.
- pfe-health-index: Builds a dynamic story for the Health Index.

elements/pfe-card/src/pfe-card.story.js
```diff
@@ -40,8 +40,7 @@ stories.add(PfeCard.tag, () => {
     size: {
       title: "Padding size",
       type: "string",
-      enum: ["small", "standard"],
-      default: "standard"
+      enum: ["small"]
```
kylebuch8: @castastrophe - Is there an option for anything other than small?

elements/pfe-cta/src/pfe-cta.story.js
```diff
@@ -23,14 +23,13 @@ stories.add(PfeCta.tag, () => {
     priority: {
       title: "Priority",
       type: "string",
-      enum: ["default", "primary", "secondary"],
+      enum: ["primary", "secondary"],
```
kylebuch8: @castastrophe - Why are we getting rid of "default"?

elements/pfe-card/src/pfe-card.story.js
castastrophe: This is an opt-in attribute so there's "default" or "small"

elements/pfe-cta/src/pfe-cta.story.js
castastrophe: We're not getting rid of default, we're letting the storybook util add in a default value automatically instead of hardcoding a value of default (which is technically not a valid input for that attribute anyway).

castastrophe: There was an issue with the util function that wasn't injecting the default value consistently so I just updated that with my last commit!  Thanks for noting this.

.storybook/utils.js
```diff
@@ -169,12 +191,17 @@ export function autoPropKnobs(properties, bridge) {
       // If an array of options exists, create a select list
       if (options.length > 0) {
         let opts = {};
+
+        // If this is not a required field, add a null option
+        if (!required) {
```
castastrophe: ^ @kylebuch8

castastrophe:
@kylebuch8 @starryeyez024 Bump on this branch when you have time.  I responded to the questions posted - let me know if I can clarify.

castastrophe:
@kylebuch8 Bump, I think this one is ready to merge!

kylebuch8:
@castastrophe There are a couple of merge conflicts. I'm not sure whether we should be using the changes from master or not.

elements/pfe-accordion/src/pfe-accordion.json
```diff
@@ -0,0 +1,49 @@
+{
```
castastrophe: @chrisdo1 Would you mind taking a look at this schema and the updates I made to the file above (pfe-accordion.js) to ensure they match what you know about the component.  I didn't see any styles or use of `color` or `theme` so I removed those from the component.

chrisdo1: @castastrophe Thanks for the heads up. This looks correct although it does contain the theme settings (on="dark") which you have accounted for in the schema.

castastrophe:
@kylebuch8 Merge conflicts are now resolved.  Thank you!

Issue #278, opened 2019-01-10
Root README out of date


markcaron:
There are many references to RH Elements, the RH Elements logo, and the wrong repo to clone in the root README. Also, we should consider replacing the `npm install` command with `npm run bootstrap` since the former produces errors.

mwcz:
What errors does `npm install` cause?  It is the correct command and `npm run bootstrap` should never be required.

castastrophe:
This should be resolved by https://github.com/patternfly/patternfly-elements/pull/279

PR #279, opened 2019-01-10
Update README.md

castastrophe:
Fixes issue https://github.com/patternfly/patternfly-elements/issues/278


markcaron:
I think we should just remove the RH Elements logo.

markcaron:
Nice! I see it was removed.

markcaron:
Limber Goats Traverse Mountainside

PR #283, opened 2019-01-11
Bundle adjacent modules

mwcz:
This is a tiny one-character change that causes adjacent modules to be bundled into elements.  Here's what I mean by adjacent, by example:

With these files...

```
├── pfelement.js
└── reveal.js
```

I'm calling reveal an adjacent module, because it sits beside (`./`) pfelement.

`pfelement.js` contains `import "./reveal.js"`, which in the UMD build will cause reveal's content to be embedded in `pfelement.umd.js`.  But in the ESM build, so far, it is not embedded (because when ESM is in use it can simply be loaded by the browser), but this PR causes reveal to be embedded in `pfelement.js`.

The main reason is performance.  Even with http/2 it's still faster to bundle.  A side benefit is that `reveal.js` won't appear in the network tab, which is a nice way to clean things up.  No one will ever wonder "What is this reveal.js file?!".


scripts/rollup.config.factory.js
```diff
@@ -14,7 +14,7 @@ function esmConfig({ elementName, className } = {}) {
       sourcemap: true
     },
     plugins: [resolve(), commonjs()],
-    external: id => id.startsWith(".")
+    external: id => id.startsWith("..")
```
castastrophe: @mwcz Does this cover both `.` _and_ `..`?  What about situations like pfe-tabs where it might be loading `./pfe-tab` and `./pfe-tab-panel`?  Would this allow for both syntaxes to work as expected? (I have no idea, just trying to think of possible use-cases that could break this update).

mwcz: This change doesn't affect whether those syntaxes work, it just changes where and when the imported modules are loaded.  `external` determines whether a module should be embedded into the bundle, or loaded separately at runtime.  If `pfe-tabs` imports `./pfe-tab.js`, it will now be embedded into `pfe-tabs.js` instead of loading `pfe-tab.js` separately.

castastrophe: And that is only for the `.umd` files?  Or does it embed them in all `js` compilations?

mwcz: This change only affects ESM; the UMD bundles already have `..`.  The reasoning for UMD having it from the beginning is, for example: if `pfe-avatar.js` imports `./hex2rgb.js`, it's very important that `hex2rgb.js` be embedded in the `pfe-avatar` bundle, because although RequireJS/AMD mode would be able to automatically load `hex2rgb.js`, anyone (like RHDC!) using global mode would be forced to add a script tag for `hex2rgb.js` just in order to use pfe-avatar.  That would be quite bad developer UX ("I have to add a script tag for _what_?"), so I decided to bundle them.  I decided _not_ to bundle them in ESM because the developer UX is much better (`<script type="module" src="pfe-avatar.js">` automatically loads `hex2rgb.js`), but now I think it's worth bringing them in line with each other for the reasons listed in the issue above.

PR #307, opened 2019-01-24
Move cp-theme from the old NPM org and into @patternfly

mwcz:
@markcaron @kylebuch8  Should cp-theme be in the @patternfly org on NPM?  It's still in @rhelements.  Maybe it shouldn't be in either... what do you all think?


kylebuch8:
@mwcz I don't think cp-theme should be in the PatternFly org. It should probably just live on GitLab or something else internal.

castastrophe:
@mwcz Closing this PR since the theme is going to be moved out - feel free to reopen if plans change.

PR #312, opened 2019-01-28
Card feature updates

castastrophe:
Incorporate a few new features into the card layout:

- Support for background color in the header region
- Update variable names and add additional hooks
- Add overflow setting to allow children in slots to fill the card
- Add storybook support for self-closing tags (img, br, hr, etc.)
- Updates border-radius to 3px globally


starryeyez024:
@castastrophe after running `npm run cleand && npm install`, the command `npm run storybook` returns an error at localhost:9001 that says: 
`Story of "Call to action" named "pfe-cta" already exists`

elements/pfe-card/src/pfe-card.js
```diff
@@ -82,20 +82,30 @@ class PfeCard extends PFElement {
...
   get backgroundColor() {
-    return this.getAttribute("color") || "base";
+    return (
+      this.getAttribute("pfe-color") || this.getAttribute("color") || "base"
+    );
   }
 
   static get observedAttributes() {
-    return ["color"];
+    return ["pfe-color", "pfe-img-src", "pfe-size"];
```
chrisdo1: @castastrophe I'm curious about the pfe-size=small attribute. Why not set this as a boolean? Would we want to leave it open for other size values later?

elements/pfe-card/src/pfe-card.scss
```diff
@@ -1,127 +1,149 @@
...
+  // Start of style declarations for host element
+  display: flex;
+  flex-direction: column;
+  justify-items: flex-start;
 
-.pfe-card__body::slotted(*:nth-child(2)) { // child pseudo-selectors appear to be based on slotted code before upgrade.
-  margin-top: 0 !important;
-}
+  padding:       #{pfe-local(Padding)};
+  border:        #{pfe-local(Border)};
```
starryeyez024: I don't know that we want a border around the card by default... can the default color be transparent? It creates a rim around the card header:
screen shot 2019-02-19 at 4 43 12 pm

starryeyez024:
http://pastebin.test.redhat.com/718499

elements/pfe-card/src/pfe-card.scss
castastrophe: The border color is set to transparent already but the fact that it has width is pushing the header in.  I think I need to set the weight to 0px instead to fix this.

elements/pfe-card/src/pfe-card.js
castastrophe: Yes and I wanted to match the approach we've taken in other components too

elements/pfe-card/src/pfe-card.scss
castastrophe: Fixed this.

kylebuch8:
Alright. I think we're finally good to go with card. Just a heads up, we're making a few changes to the PFElement base class.

In the `connectedCallback()`, we're adding `this.setAttribute("pfelement", "")` and then in the `_mapSchemaToSlots` method, we're using `this.setAttribute(`has_${slot}`, "")` if the slot exists.

This is all done because set classes might be fragile.

[View this comment](https://github.com/angular/angular/issues/15399#issuecomment-318785677) to see what helped guide us to these changes.

PR #313, opened 2019-01-29
using default instead of null

kylebuch8:
@castastrophe Let me know what you think.

I prefer seeing the word default in the dropdown as opposed to an empty field.

### DIFF
--- a/.storybook/utils.js
+++ b/.storybook/utils.js
@@ -194,7 +194,7 @@ export function autoPropKnobs(properties, bridge) {
 
         // If this is not a required field, add a null option
         if (!required) {
-          opts.null = "";
+          opts[""] = "default";
         }
 
         // Convert the array into an object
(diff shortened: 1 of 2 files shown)


.storybook/utils.js
```diff
@@ -194,7 +194,7 @@ export function autoPropKnobs(properties, bridge) {
 
         // If this is not a required field, add a null option
         if (!required) {
-          opts.null = "";
+          opts[""] = "default";
```
castastrophe: @kylebuch8 I'm fine if the printed value is default but the stored value should stay `null` for the rest of the logic in the dynamic build to work.

elements/pfe-card/src/pfe-card.story.js
```diff
@@ -85,7 +85,6 @@ stories.add(PfeCard.tag, () => {
     const ctaPriorityValue = storybookBridge.select(
       "Priority",
       {
-        null: "",
```
castastrophe: This is manually building the select field and not using the storybook utils I wrote so if you want it to include the default value, this needs to stay: `null: "default"`

.storybook/utils.js
castastrophe: i.e., `opts.null = "default"`

kylebuch8:
Fixed in https://github.com/patternfly/patternfly-elements/pull/277/commits/adf90c6aca8b31fd25a24741a47af69862b70ac9

PR #317, opened 2019-01-31
Fixes #314: using group.children instead of the query selector

### LINKED ISSUE #314: Content set: bug rendering nested components inside the panel
When adding a `pfe-cta` to the panel of the `pfe-content-set`, the content inside the nested component is being pulled out and rendered twice:

Rendering issue:
screen shot 2019-01-29 at 2 22 51 pm

Screenshot of the shadow root:
screen shot 2019-01-29 at 2 23 06 pm

kylebuch8:
Unfortunately, there isn't a querySelectorAll that works for getting the immediate children of an element. :scope would work but it has limited support and I dont want to add a polyfill. Using .children and then filtering out the things that don't have a pfe-heading attribute should do the trick.

Fixes #314


elements/pfe-content-set/demo/index.html
```diff
@@ -36,6 +37,7 @@ <h3>Container max: 500px</h3>
         <pfe-content-set-group>
           <h2 pfe-heading>Heading 1</h2>
           <p>Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore </p>
+          <pfe-cta priority="primary" color="base" on="dark"><a href="#">Primary</a></pfe-cta>
```
castastrophe: @kylebuch8 Just so you know, `on="dark"` no longer needs to be manually added to CTA as the container element will add that when it's appropriate.

castastrophe: I updated this and prettified/clarified the template file.  I was updating to add a few other variations to ensure there were no regressions.

castastrophe:
👍 Lemmings going toward mountains

starryeyez024:
Any reason why this can't be merged?

castastrophe:
@starryeyez024 No, it's ready and approved!  It was just pending AT passing and it has. w00t.

PR #322, opened 2019-02-12
Removing duplicate has_slot function in PFElement

markcaron:
Removing duplicate has_slot function in PFElement by making it has_slots to return array of slots with the same name.


markcaron:
I noticed that `has_slot` only returns a single slot, but if you have multiple instances of the same slot name (like I will in `rh-footer`) it doesn't help. Since there were two `has_slot` functions for some reason, I just converted one of them to the multiple / querySelectorAll version.

elements/pfelement/src/pfelement.js
```diff
@@ -39,8 +39,8 @@ class PFElement extends HTMLElement {
     return this.querySelector(`[slot='${name}']`);
   }
 
-  has_slot(name) {
-    return this.querySelector(`[slot='${name}']`);
+  has_slots(name) {
```
kylebuch8: @markcaron I checked the entire codebase and didn't see any instances of us using `has_slots`. While I do see this being useful, if we're not using it anywhere, I think we should just remove the function.

castastrophe: @kylebuch8 pfe-band is using `has_slot` singular and I like the idea of the two being separate, `has_slot` versus `has_slots` so you can tell by the name of the function what return to expect. 👍  It sounds like @markcaron needs the `has_slots` in a soon-to-be-built component so I vote yes to this update.

elements/pfelement/src/pfelement.js
```diff
@@ -39,8 +39,8 @@ class PFElement extends HTMLElement {
...
+    return this.querySelectorAll(`[slot='${name}']`);
```
castastrophe: Should we go ahead and make this an array for convenience? `return [...this.querySelectorAll(`[slot='${name}']`)];`

kylebuch8: @castastrophe Let's go with `[...this.querySelector(`

castastrophe: Sorry just to clarify - `[...this.querySelectorAll`?

castastrophe:
@kylebuch8 I updated this to reflect the discussion above - let me know if it looks good to merge!

PR #328, opened 2019-02-14
run a build after npm install

mwcz:


package.json
```diff
@@ -8,7 +8,7 @@
     "build": "bash scripts/build.sh",
     "clean": "rm -rf node_modules package-lock.json {elements,themes}/*/{node_modules,package-lock.json}",
     "bootstrap": "lerna bootstrap --hoist",
-    "postinstall": "npm run bootstrap",
+    "postinstall": "npm run bootstrap && npm run build",
```
castastrophe: So postinstall runs after `npm install` is run but I think we also need the build to run before `npm run start` right?  To ensure users are seeing the latest in the preview mode?  I think running it before storybook is kicked off would also be helpful.

mwcz: I don't think that's necessary.  It has always been the case that builds must be run before the latest changes are visible in storybook and/or demo pages.  The `dev` task covers that need very nicely, without slowing down `start` and `storybook`.

PR #329, opened 2019-02-15
WIP: Package updates with newest Babel!

castastrophe:
- Update to latest babel version and the file updates that go with it
    - Define supported browsers for better fallback support
- Update demo pages for consistency in how elements are pulled in
- Add scaling attributes to each page for better rendering on various devices


elements/pfe-accordion/.babelrc
```diff
@@ -1,4 +1,17 @@
 {
-  "presets": [["env", { "modules": false }]],
```
mwcz: Am I right in guessing that `modules: false` is no longer necessary in the new babel version?

starryeyez024:
Is this too stale? should we close or try to update with master??

castastrophe:
> Is this too stale? should we close or try to update with master??

No I think we keep it open - I need the babel update code, it's very important; just not a high priority.  I'd rather not try to update it right now until I have time to go through it.

Issue #310, opened 2019-01-28
pfe-band issues / suggestions


starryeyez024:
### 1. height of children in body

```css
.pfe-band__body {
    display: flex;
    flex-flow: column nowrap;
    align-items: flex-start; // this prop causes this children not to fill the height of the row
}
```

![card-height](https://user-images.githubusercontent.com/456863/51810149-27c0ac80-2274-11e9-8179-338227d9e788.gif)

### 2. Inline styles in the demo 

Demo page, orange example "Aside: left body top" has inline styles causing overlap
 `style="height: -webkit-fill-available;"`

![overlap](https://user-images.githubusercontent.com/456863/51811512-44acae00-227b-11e9-9786-5b93b6a0504e.jpg)

### 3. Formatting of grid props for readability

Not a bug, but a suggestion. Can we format the CSS so that its easier to see what the regions are?
```
--pfe-band--gridTemplateArea_desktop: 
          "header  header" 
          "body    aside" 
          "footer  footer";
```

### 4. Band should set max-width instead of width

To avoid the band breaking out of any other container

![image](https://user-images.githubusercontent.com/456863/53104815-0dd65a80-34fe-11e9-9f2e-d2b5eaae7081.png)

castastrophe:
Hmm, does it make sense to spit these out into separate issues so they can be discussed/fixed individually?

starryeyez024:
Pull request to resolve: https://github.com/patternfly/patternfly-elements/pull/336

castastrophe:
I was thinking that newlines in CSS variables wouldn't work but they totally do!! +1 https://codepen.io/castastrophee/pen/WmXvvE?editors=1100

PR #336, opened 2019-02-20
Fixes issue #310: pfe-band - resolve layout  issues

starryeyez024:
https://github.com/patternfly/patternfly-elements/issues/310

- remove flex alignment prop that prevents cards from stretching
- change width to max-width
- adjust formatting of grid props
- remove errant style from demo file


elements/pfe-band/src/pfe-band.scss
```diff
@@ -148,19 +148,23 @@ $LOCAL: band;
     --pfe-band--gridTemplateArea_mobile: "body";
     // Styles for the different combination of regions
     &[pfe-has-aside] {
-      --pfe-band--gridTemplateArea_mobile: "body" "aside";
-      --pfe-band--gridTemplateArea_desktop: "body aside";
+      --pfe-band--gridTemplateArea_mobile:
```
castastrophe: These newlines are not necessary.  The layout issues are due to missing attributes, not to line breaks.  As long as grid templates use individual quotes around the layers, they don't need newlines.

elements/pfe-band/src/pfe-band.scss
```diff
@@ -170,66 +174,119 @@ $LOCAL: band;
...
+            "aside body"
+            "aside footer";
         }
       }
     }
 
     position:  relative;
     margin:    0 auto;
-    width: #{pfe-local(Width)};
+    max-width: #{pfe-local(Width)};
```
castastrophe: Doesn't max-width put the band at risk of being too short if the content doesn't fit?  And since it's centered we'd end up with inconsistently sized bands.  This needs to be width or min-width.

elements/pfe-band/src/pfe-band.scss
starryeyez024: agreed, they don't change the resulting CSS. The line breaks are for readability, to view the regions as they would appear in a grid format.

elements/pfe-band/src/pfe-band.scss
starryeyez024: Hmm, not sure what you mean? max-width != max-height
It does need `width:100%; max-width: #{pfe-local(Width)};` to ensure it fills out the available space though. good catch!

starryeyez024: updated b6b00181c23d195252f2b17fa0a5dd8a9f6ee9e9

castastrophe: Short horizontally not vertically so still related to width.  The width: 100% fixes the issue though.  Just noting it needs both width and max-width or min-width.  +1

elements/pfe-band/src/pfe-band.scss
castastrophe: +1, I wasn't sure they would work with custom properties but they DO! 🎉

PR #334, opened 2019-02-18
Add autoprefixer to the repo

castastrophe:
Issue #330

PR #340, opened 2019-02-22
ESM .min.js

mwcz:
This is a change by @castastrophe, extracted from the #334 branch.  It causes the ESM builds to generate two files, `pfe-foo.js` is unminified ESM and `pfe-foo.min.js` is minified ESM.

### DIFF
--- a/elements/pfe-avatar/demo/index.html
+++ b/elements/pfe-avatar/demo/index.html
@@ -13,16 +13,15 @@
     <script src="https://cdnjs.cloudflare.com/ajax/libs/vue/2.5.2/vue.min.js"></script>
 
     <!-- uncomment the es5-adapter if you're using the umd version -->
-    <script src="/@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js"></script>
-    <script src="/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.5/require.min.js"></script>
-    <script>require(['../../pfelement/pfelement.umd.min.js', '../pfe-avatar.umd.min.js'], function(PFElement) {
-      PFElement.debugLog(true);
-    })</script>
+    <!-- <script src="/@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js"></script> -->
+    <!-- <script src="/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script> -->
+    <!-- <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.5/require.min.js"></script> -->
+    <!-- <script>require(['../../pfelement/pfelement.umd.js', '../pfe-avatar.umd.js'], function(PFElement) { -->
+    <!--   PFElement.debugLog(true); -->
+    <!--   })</script> -->
 
-    <!-- <script type="module" src="../pfe-avatar.min.js"></script> -->
-    <!-- <script type="module" src="../pfe-avatar.js"></script> -->
-    <!-- <script src="../../../themes/cp-theme/cp-theme.umd.js"></script> -->
+    <script type="module" src="../pfe-avatar.js"></script>
+    <script src="../../../themes/cp-theme/cp-theme.umd.js"></script>
 
     <style type="text/css" media="screen">
       /* :root {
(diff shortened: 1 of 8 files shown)


mwcz:
Sourcemaps for non-minified files are now working (mapping `foo.js` &rarr; `src/foo.js`) but minified sourcemaps are not (mapping `foo.min.js`  &rarr; `foo.js`  &rarr; `src/foo.js`).

mwcz:
One problem is that the minified files are importing non-minified files.

For example, `pfe-avatar.min.js` is importing  `../pfelement/pfelement.js` but it needs to import `../pfelement/pfelement.min.js`.  We do a similar transform in the gulpfile `merge` task when generating the `umd.js` files.

castastrophe:
@mwcz Do you want to make the update on the importing of non-minified files to fix the sourcemaps in this branch?  Or would you prefer to make that update in a separate branch?

mwcz:
I'd like to fix both importing min.js and sourcemaps in this branch.  I'll try to tackle that today.

mwcz:
This is done and ready for review!

elements/pfe-avatar/demo/index.html
```diff
@@ -13,15 +13,16 @@
...
-    <!-- <script src="/@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js"></script> -->
-    <!-- <script src="/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script> -->
-    <!-- <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.5/require.min.js"></script> -->
-    <!-- <script>require(['../../pfelement/pfelement.umd.js', '../pfe-avatar.umd.js'], function(PFElement) { -->
-    <!--   PFElement.debugLog(true); -->
-    <!--   })</script> -->
+    <script src="/@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js"></script>
+    <script src="/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.5/require.min.js"></script>
+    <script>require(['../../pfelement/pfelement.umd.min.js', '../pfe-avatar.umd.min.js'], function(PFElement) {
```
castastrophe: @mwcz Do you still need this debug statement?

mwcz:
Yeah, those should be there.  Those are the sourcemaps, without them we can't debug.  If they are ever missing, raise the alarm!

castastrophe:
Yeah, I knew they were sourcemaps, just didn't realize we needed one for every format we compile.  Makes sense though.  Yay for file explosion.  Perhaps another potential argument for a dist/compiled directory :)

elements/pfelement/src/pfelement.js
```diff
@@ -2,6 +2,10 @@ import { autoReveal } from "./reveal.js";
 const prefix = "pfe-";
 
 class PFElement extends HTMLElement {
+  static get version() {
+    return "$version";
```
castastrophe: Is this related to the minification?

package.json
```diff
@@ -77,6 +77,7 @@
     "rollup-plugin-babel": "^3.0.7",
     "rollup-plugin-commonjs": "^9.1.5",
     "rollup-plugin-node-resolve": "^3.3.0",
+    "rollup-plugin-re": "^1.0.7",
```
castastrophe: Should we remove gulp-rename?  Is that being used elsewhere?

mwcz: It's used in the gulpfiles for icon, layouts, pfelement, and cp-theme.

elements/pfelement/src/pfelement.js
mwcz: Whoops, no, I was just playing around with the gulp-plugin-re.  I'll remove it from this PR.  I would like to add the version number somewhere, either as an element property or maybe in an `@preserve` comment at the top of each file ([like](https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.11/lodash.min.js) [many](https://cdnjs.cloudflare.com/ajax/libs/jquery/3.3.1/jquery.min.js) [libs](https://cdnjs.cloudflare.com/ajax/libs/vue/2.6.7/vue.min.js) [do](https://cdnjs.cloudflare.com/ajax/libs/react/16.8.3/umd/react.production.min.js)), so if we're ever looking at a source file we can determine what version it is.

castastrophe:
Pending travis, this is 👍

elements/pfe-avatar/demo/index.html
mwcz: It's not needed exactly, in fact none of the changes to that file are needed (I was using this file to test out the various bundle types), so I reverted it back to how it is in master.

PR #343, opened 2019-02-27
WIP use pfe-band in navigation tray

starryeyez024:
@castastrophe can you tell me more about why this approach won't work?


castastrophe:
It's not that it won't work - there are just pros and cons.  I think the issue of the markup (i.e., pfe-band) controlling the bounds of the container for the tray is a pretty big lock-in.  What if band evolves?  What if this pattern evolves?  Now we're stuck with having to update markup to accomplish updates where styles could have handled it.  I think we use the same placeholder concept (turned mixin for this repo) as webrh where we have a `container` that we can extend.

castastrophe:
@starryeyez024 Do you mind if we close this one?

starryeyez024:
Sure, you can close this.

Issue #345, opened 2019-02-28
pfe-content-set doubles up content in light dom; doesn't respect light dom styles for things in panel


starryeyez024:
![image](https://user-images.githubusercontent.com/456863/53593938-cc7e2480-3b67-11e9-9085-1d61963e9e2a.png)

image.png
![image](https://user-images.githubusercontent.com/456863/53593922-c5efad00-3b67-11e9-9dd1-7d7256251d4f.png)

![image](https://user-images.githubusercontent.com/456863/53593907-bbcdae80-3b67-11e9-9ae2-248500308254.png)

kylebuch8:
@starryeyez024 This should be fixed in #346

PR #346, opened 2019-03-04
Fixes issue #345: pfe-content-set 

kylebuch8:
… dom


elements/pfe-content-set/README.md
```diff
@@ -3,35 +3,27 @@
...
 
 Requires both the `pfe-accordion` and `pfe-tabs` components, as well as the base `pfelement`.
 
-
 ## Usage
 
-You must wrap each chunk of information (heading + content) in a `<pfe-content-set-group>` tag. 
-
-You must also apply the `pfe-heading` to the HTML heading element, like an H2 or H3. Otherwise content within the component will not be rendered.
+<!-- You must wrap each chunk of information (heading + content) in a `<pfe-content-set-group>` tag. -->
```
castastrophe: Can this comment be deleted now?  It appears not to be needed anymore.

elements/pfe-content-set/demo/index.html
```diff
@@ -2,219 +2,225 @@
...
+            <a href="https://redhat.com">Primary CTA</a>
+          </div>
+        </div>
+        <h2 pfe-content-set--header>Heading 2</h2>
+        <div pfe-content-set--panel>Content for heading 2.</div>
+        <h2 pfe-content-set--header>Heading 3</h2>
+        <div pfe-content-set--panel>Content for heading 3.</div>
       </pfe-content-set>
     </div>
+
```
castastrophe: Can we remove this commented out code?

elements/pfe-content-set/src/pfe-content-set.js
```diff
@@ -31,69 +42,84 @@ class PfeContentSet extends PFElement {
...
-      return " vertical";
-    }
+  constructor() {
+    super(PfeContentSet, { delayRender: true });
   }
 
-  static get observedAttributes() {
-    return ["pfe-vertical", "selected-index", "pfe-variant", "theme", "color"];
-  }
+  connectedCallback() {
```
castastrophe: This is so freaking clean and beautiful and simple. Many +1s.  Much happy.

elements/pfe-content-set/src/pfe-content-set.js
```diff
@@ -31,69 +42,84 @@ class PfeContentSet extends PFElement {
...
-  constructor() {
-    super(PfeContentSet, {
-      delayRender: true
+  _buildAccordion() {
+    const fragment = document.createDocumentFragment();
+    const accordion = document.createElement("pfe-accordion");
+
+    [...this.children].forEach(child => {
+      if (child.hasAttribute("pfe-content-set--header")) {
+        const header = document.createElement("pfe-accordion-header");
```
castastrophe: Wow.  So nice.  This is my new go-to coding example.

elements/pfe-content-set/test/pfe-content-set_test.html
```diff
@@ -6,94 +6,67 @@
...
+          <h2 pfe-content-set--header>heading 2</h2>
+          <p pfe-content-set--panel>Consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus.</p>
```
castastrophe: Should there be an intentionally borked example in the test case?  One we know will fail to test that it fails gracefully?

elements/pfe-tabs/src/pfe-tab.scss
```diff
@@ -182,8 +182,14 @@ $pfe-tabs__tab--PaddingBottom-factor:  1.5; // 54px at 16px base.
   }
 }
 
-::slotted(h2) {
+::slotted(h1),
```
castastrophe: Suggest:
```
%tab-headings {
    font-size: #{pfe-var(font-size)};
    font-weight: #{pfe-var(normal, font-weight)};
    margin: 0;
    user-select: none;
}

@for $i from 1 through 6 {
    ::slotted(h#{$i}) {
        @extend %tab-headings;
    }
}
```

elements/pfe-tabs/src/pfe-tabs.js
```diff
@@ -152,7 +153,7 @@ class PfeTabs extends PFElement {
     super.connectedCallback();
 
     this.addEventListener("keydown", this._onKeyDown);
-    this.addEventListener("click", this._onClick);
+    // this.addEventListener("click", this._onClick);
```
castastrophe: Does this need to be uncommented?  Do we not want to use this click event?

elements/pfe-tabs/src/pfe-tabs.js
```diff
@@ -434,16 +437,16 @@ class RhTab extends PFElement {
...
-    if (!this.id) {
-      this.id = `${RhTab.tag}-${generateId()}`;
-    }
+    super(PfeTab);
   }
 
   connectedCallback() {
     super.connectedCallback();
 
+    if (!this.id) {
```
castastrophe: Can we update this to attach a `pfe-id` instead of using ids?  This causes some issues for us when building A/B tests with the Adobe Test & Target tool because it will always see and try to use the ID (it's a GUI) and fail because this is randomly generated.

elements/pfe-tabs/src/pfe-tab.scss
starryeyez024: Hmm,  listing the headings might not seem DRY, but it is easier to scan quickly to find what you're looking for, because you can see the h1 or h2, etc. 
Since they both net out to 11 lines, can we leave the easier looking one for simplicity's sake?

elements/pfe-tabs/src/pfe-tabs.js
kylebuch8: @castastrophe It won't be randomly generated if the id attribute is already present on the element. Is that still a problem if you give it an ID ahead of time?

elements/pfe-tabs/src/pfe-tab.scss
castastrophe: Sure, it's easier to read to me but I'm fine leaving it expanded if that's preferred.  Note that I also cleaned up the var references though too.  Could we pull that update in even if we don't use the loop?

elements/pfe-tabs/src/pfe-tabs.js
castastrophe: That means every time we use the pattern we'd have to include instructions that it has to have an ID on it.  Is the ID important to the connection?  What would it hurt to use a more internal way of connecting the parts?

elements/pfe-content-set/package.json
```diff
@@ -28,8 +28,5 @@
     "@patternfly/pfe-tabs": "^1.0.0-prerelease.9",
     "@patternfly/pfelement": "^1.0.0-prerelease.9"
   },
-  "devDependencies": {
-    "@patternfly/pfe-sass": "^1.0.0-prerelease.9"
```
castastrophe: Why remove pfe-sass as a dev dependency?

kylebuch8: @castastrophe We're not using it. We're just using

```
:host {
  display: block;
}

:host([hidden]) {
  display: none;
}
```

castastrophe: Oh of course.  This is a combo. #duh

castastrophe:
@kylebuch8 Thanks for the updates!  This is looking great.  I'll pull it down and test.  Have you done browser testing already?

elements/pfe-tabs/src/pfe-tab.scss
chrisdo1: I will leave this as is then.

elements/pfe-content-set/demo/index.html
chrisdo1: @castastrophe I'm not seeing any commented code in this file. Perhaps it has already been removed?

elements/pfe-content-set/README.md
chrisdo1: @castastrophe I have removed that.

elements/pfe-content-set/test/pfe-content-set_test.html
chrisdo1: @castastrophe I added a broken content set example (at the very bottom of the page) in which I included no attributes on the header or panel.

elements/pfe-tabs/src/pfe-tabs.js
chrisdo1: @castastrophe @kylebuch8 I have uncommented that line.

PR #349, opened 2019-03-05
move docs into monorepo

mwcz:
Move documentation into the main repo.  The two major benefits of this are:

Pull requests can contain both code and related docs, together at last.  Docs can even be required before approving a PR, similar to how tests are/should be required.

With docs as a first-class member of the main repo, everyone will have closer proximity to them and a better understanding of how to update; it's currently an oft-forgotten repo that few people use.


scripts/dev.sh
```diff
@@ -6,4 +6,13 @@ for el in "$@"; do
   CMD="$CMD --scope \"*/$el\""
 done
 
+source scripts/hugo-check.sh
+if hugoCheck; then
+  pushd docs > /dev/null
+  hugo server &
+  popd > /dev/null
+else
+  echo "Not running Hugo server (docs site) becuase hugo is not installed."
```
castastrophe: minor typo: because

castastrophe:
`npm run start` failed with the following:

```
patternfly-elements  (docs-docs-docs) $ npm run start

> @patternfly/patternfly-elements@ start /Users/carobert/repos/patternfly-elements
> npm run doc-listing-inject && spandx


> @patternfly/patternfly-elements@ doc-listing-inject /Users/carobert/repos/patternfly-elements
> node scripts/doc-listing-inject.js

/Users/carobert/repos/patternfly-elements/scripts/doc-listing-inject.js:29
  if (err) throw err;
           ^

Error: ENOENT: no such file or directory, open '/Users/carobert/repos/patternfly-elements/doc/wrapper.html'
```

mwcz:
Thanks, fixed!

scripts/deploy-docs.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+source hugo-check.sh
```
castastrophe: Would it be potentially easier to set up hugo to install with npm using https://www.npmjs.com/package/hugo-installer?

castastrophe: There's also a hugo-cli that might be useful https://www.npmjs.com/package/hugo-cli

mwcz:
Yes that is awesome!!!

scripts/deploy-docs.sh
mwcz: Yes that is awesome!!!

mwcz: I won't have time to add those packages this sprint, so I made #352 to keep track of it.  I'd like to merge this PR as-is and add those next sprint if no one objects.

scripts/dev.sh
mwcz: Thanks, fixed!

scripts/deploy-docs.sh
castastrophe: 👍

castastrophe:
Just for posterity: confirming that I tested `npm run start`, `npm run build`, `npm run storybook`, and `npm run dev` to ensure no regressions.

Issue #298, opened 2019-01-24
Update pfe-cta to use new attribute naming conventions


starryeyez024:
per https://github.com/patternfly/patternfly-elements/issues/220

Issue #294, opened 2019-01-20
pfe-cta needs to use broadcast variable(s)


starryeyez024:
`--pfe-broadcasted--color--ui-link`

https://rally1.rallydev.com/#/detail/userstory/305092942296?fdp=true

Issue #236, opened 2018-11-06
pfe-cta: Rounded corners should not be the default


kylebuch8:
There is already a CSS variable ui--border-radius but we need to expose a variable in the rh-cta.scss: 

--rh-cta--ui--border-radius: ui--border-radius;   

Can we set the theme variable to default  to 0px?
--ui--border-radius: 0px

markcaron:
Default for what? Red Hat or the community?

Currently the default is for the base RH Elements, which is community-based. So, I don't agree that we should make the fallback 0px.

A theming developer can easily change this with CSS vars to square corners. This would be the proper approach to making a site, like Red Hat's, have square corners (which is our Brand's preference).

markcaron:
If you're wanting to expose a custom, component-level var, we can do that. The border-radius statement would need to look like:

```
border-radius: calc(var(--rh-cta--ui--border-radius, #{rh-var(ui--border-radius)}) * 20);
```

This would give us a more broad stroke to turn ONLY CTAs square/round. But, I don't think we should set the default (fallback in rh-sass) to be 0px.

castastrophe:
+1 @markcaron, I like your solution.  What would you think of slightly reducing the rounded corner in the community project though?  I'm also curious why it's multiplied by 20?

Issue #220, opened 2018-10-17
Conventions for attributes and custom properties


starryeyez024:
If it's not a problem with the JavaScript, these are my proposed naming standards:

## Requested standard attributes
- `type="    "`
    - Values:  `content`, `container`, `combo`
- `format="  "`
    - Example values, would vary per component:  `vertical`, `horizontal`
- `variant="  "`
    - Purpose: to group various styles together into one batch of settings
    - Values:  `primary` , `secondary`, `tertiary`
- `color="  "`
    - Purpose: Invoke palette color override
    - Values: `base`, `complement`, `accent`

## Rules of thumb: 
 - Use attributes with values instead of boolean attributes 
    - instead of  `<rh-pagination show-jump show-page>`
    -  do this: `<rh-pagination show="jump page whatev"`  

## Remove
- `priority="  "`
    - remove this, change to variant
    - instead we could use it to adjust semantic markup in a heading or something

### Custom properties
 - Used for small "flair" type styles, like the color of an underline, or rounded corner radius, etc.

mwcz:
One minor change, `type` should be `rh-type`, since `type` is a standard HTML attribute.  See #52

starryeyez024:
ack thx!

starryeyez024:
@mwcz @kylebuch8  is it harder to target attributes like `<rh-pagination show="arrow">` than `<rh-pagination show-arrow>`?

castastrophe:
@starryeyez024 I don't think show by itself makes sense as an attribute though.  I could see not wanting the word arrow specifically but what about `show-indicator`.  I see a lot of value and logic in the boolean attribute and prefer we keep it in certain situations like this one.  If we name them more generically so we can evolve the design beneath them, do you still oppose?

Homa:
control-buttons="arrow" 
control-buttons="caret"

Better not to use show and hide attributes without prefix. They usually are used in js libs to hide and show elements.

castastrophe:
I like the theory that attributes should be human-readable.  For example, `<rh-cta show-pagination style-variant="1" indicator-style="caret">`

starryeyez024:
Combos are bringing together existing web components

- allows user to use more succinct markup 
- editors would ideally use combos for specific use cases, i.e. 
- collection of settings
- doesn't contain its own CSS

starryeyez024:
From Dec 14th FED architecture meeting:

# Requested standard attributes

## Why we should prefix attributes

[...]

PR #353, opened 2019-03-08
pfe-cta bug fixes + open brand styles

castastrophe:
https://rally1.rallydev.com/#/detail/userstory/305092942296?fdp=true

- https://github.com/patternfly/patternfly-elements/issues/294
- https://github.com/patternfly/patternfly-elements/issues/298
- https://github.com/patternfly/patternfly-elements/issues/236


starryeyez024:
apologies in advance b/c i dont have time to checkout this branch & look, but wanted to capture somewhere that the current CTA component in the master branch has CSS such that the arrow breaks to a new line by itself. Not sure if you already solved for that or not #forgiveme 
image

castastrophe:
@starryeyez024 I'm confused - the CTA buttons don't have the right arrow on them...

package.json
```diff
@@ -81,7 +81,7 @@
     "gulp-trim": "^1.0.2",
     "gulp-uglify": "^3.0.1",
     "husky": "^1.1.1",
-    "lerna": "^3.0.0-rc.0",
+    "lerna": "^3.14.1",
```
mwcz: Upgrading lerna and browser-sync sounds good, but can we pull these into a separate PR?

elements/pfe-cta/src/pfe-cta.scss
```diff
@@ -1,200 +1,344 @@
...
+  background-color: #{pfe-local(background)};
+
+  padding:       #{pfe-local(Padding)};
+  border-radius: #{pfe-local(BorderRadius)};
+  border:        #{pfe-var(ui--border-width)} #{pfe-var(ui--border-style)} #{pfe-local(border)};
+
+  // Update the broadcasted variables, map to main color
+  --pfe-broadcasted--color--ui-link: #{pfe-local(Color)};
+  --pfe-broadcasted--color--ui-link--hover: #{pfe-local(Color--hover)};
+  --pfe-broadcasted--color--ui-link--focus: #{pfe-local(Color--focus)};
```
starryeyez024: Yay!!! this works great. :)

starryeyez024: Ok fleeting friday afternoon thought: this does not solve for this scenario:
```
    <pfe-band pfe-color="darkest"> <-- should pass through the lighter blue broadcasted var -->
        <pfe-cta ><a href="#">Default</a></pfe-cta> <--overrides it to basic link blue -->
    </pfe-cta>
```

image

elements/pfe-cta/src/pfe-cta.js
```diff
@@ -13,42 +13,94 @@ class PfeCta extends PFElement {
...
     const firstChild = this.children[0];
 
-    if (!firstChild) {
-      console.warn(
-        `${
-          PfeCta.tag
-        }:The first child in the light DOM must be an anchor tag (<a>)`
-      );
-    } else if (firstChild && firstChild.tagName.toLowerCase() !== "a") {
+    if (!firstChild || (firstChild && !["A", "BUTTON", "INPUT"].includes(firstChild.tagName))) {
```
starryeyez024: @castastrophe this is throwing errors in IE11; when I revert it back to the original code it works. Can we back this out & make it a todo for later?

elements/pfe-cta/src/pfe-cta.js
```diff
@@ -13,42 +13,94 @@ class PfeCta extends PFElement {
...
+  get defaultStyle() {
+    return this.hasAttribute("priority") ? false : true;
+  }
+
+  static get observedAttributes() {
+    return [
+      "pfe-priority",
+      "pfe-color",
+      "pfe-alt",
+      "on"
```
starryeyez024: we need to remove alt & on @chrisdo1

elements/pfe-sass/README.md
```diff
@@ -1,3 +1,68 @@
...
+4. Reset local variable values as needed for color attribute overrides.
+
+
+### CSS Example
+
+
+```
+    // 1. set up local vars equal to theme vars & fallbacks
+    :host {
+      --pfe-local--Color: var(theme--ui-link, #06c);
```
starryeyez024: make example clearer that pfe-local means somethign like pfe-cta

starryeyez024: for example if you are in CTA... its pfe-cta  @starryeyez024

elements/pfe-cta/src/pfe-cta.json
```diff
@@ -0,0 +1,85 @@
...
+            "primary",
+            "secondary"
+          ],
+          "observer": "_basicAttributeChanged"
+        },
+        "color": {
+          "title": "Color",
+          "type": "string",
+          "enum": [
+            "base",
```
starryeyez024: need to update these; remove base, add lightest

elements/pfe-cta/src/pfe-cta.scss
```diff
@@ -1,200 +1,367 @@
...
-  //  &::after {
-  //    @include pfe-arrow-color(var(--pfe-cta--main--visited));
-  //  }
-  // }
+  position: relative;
+  transition: all #{pfe-var(animation-timing)};
 
-  ::slotted(a:hover) {
-    color: var(--pfe-cta--main--hover);
+  line-height: inherit;
```
starryeyez024: on line 39, we should update `all` and change it to `padding`

elements/pfe-cta/src/pfe-cta.js
starryeyez024: list should be
```
      "priority",
      "color",
```

chrisdo1: @starryeyez024 I actually had removed both of those but had not committed yet. Changed `pfe-color` to `color` as well.

starryeyez024: or did we decide to keep `pfe-` namespacing to avoid conflicts with angular , etc?
@castastrophe @chrisdo1 @markcaron @kylebuch8

starryeyez024: this issue says keep namespacing https://github.com/patternfly/patternfly-elements/issues/220

castastrophe: I'm so torn.  Removing the namspacing is cleaner for developers.  Angular namespaces their stuff with `ng-` though.  I defer to others.  I like it without but can see the value of keeping the namespacing too.

chrisdo1: Do we need to change the attributes in the light DOM then? 
`<pfe-cta priority="primary"><a href="#">Primary</a></pfe-cta>`
Should be:
`<pfe-cta pfe-priority="primary"><a href="#">Primary</a></pfe-cta>`

elements/pfe-cta/src/pfe-cta.js
```diff
@@ -13,42 +13,94 @@ class PfeCta extends PFElement {
     return "pfe-cta.html";
   }
```
kylebuch8: Did we intentionally leave `get schemaUrl() { return "pfe-cta.json"; }` out? If so, we should remove all of the `observedAttributes` and `attributeChangedCallback` code.

elements/pfe-cta/src/pfe-cta.js
castastrophe: Actually I can't replicate any errors.  I just pulled up the demo page in IE11 on BrowserStack and see no error messages in the console.

castastrophe: I added a polyfill for includes because it's not supported in IE11 - let me know if that does not resolve the issue you are seeing.

elements/pfe-cta/src/pfe-cta.js
castastrophe: Thanks Kyle!  I added it with the latest commit.

elements/pfe-cta/src/pfe-cta.scss
castastrophe: If we put the padding on the :host element, it works in IE11 but the default padding looks a little bigger.  Do you know why?

elements/pfe-cta/src/pfe-cta.js
starryeyez024: Since we talked about it as a larger group & it's [currently in the PFE documentation](https://github.com/patternfly/patternfly-elements-documentation/blob/master/content/develop/_index.md#attribute-naming), I think we should follow the prefix recommendation in both the JS file, the light DOM examples, everywhere (find & replace).

castastrophe: I want to note that `on=` is the exception since that is being deprecated anyway

package.json
starryeyez024: @mwcz sure i'll remove from this PR, not sure why its appearing here. Do you have this captured as an issue or something elsewhere?

elements/pfe-cta/src/pfe-cta.js
starryeyez024: same, no errors anymore, thx!

elements/pfe-cta/src/pfe-cta.scss
starryeyez024: padding is on host now :)

elements/pfe-sass/variables/_colors.scss
```diff
@@ -32,7 +32,7 @@ $pfe-color--ui-link--on-dark--visited:          lighten($pfe-color--ui-link--vis
 $pfe-color--ui-link--on-dark--hover:            lighten($pfe-color--ui-link--on-dark, 10%) !default; // Was #c0ecff
 $pfe-color--ui-link--on-dark--focus:            lighten($pfe-color--ui-link--on-dark, 10%) !default; // Was #c0ecff
 
-$pfe-color--ui-link--on-saturated:              lighten($pfe-color--ui-link, 40%) !default;               // Was #97daff
+$pfe-color--ui-link--on-saturated:              $pfe-color--white;               // Was #97daff
```
starryeyez024: This is a system change, but it makes more sense to me if no one objects?

starryeyez024:
@castastrophe OK! I think this has been tidied, tested, browser tested, and is good to go!

elements/pfe-sass/variables/_colors.scss
castastrophe: 👍

starryeyez024:
I remember now why I had added the inner-border to the CTA secondary. The whole issue is surrounding the hover state on the secondary CTA when its on a dark background. What's complicated about this scenario is that the whole button is supposed to flip, where the background is white and the text is black. We need two colors then, white and black. Since we deprecated `on="dark"` , the only information the CTA has to know that it is on a dark background is the broadcast variables.  The problem is that the card only provides one real broadcast color which is white. So we don't actually have a variable available to use to set the text color with. 

We also have a mix'n'match approach to colors still, where some pfe-color attributes requesting surface colors and other attributes requesting ui colors:

image

chrisdo1:
> I remember now why I had added the inner-border to the CTA secondary. The whole issue is surrounding the hover state on the secondary CTA when its on a dark background. What's complicated about this scenario is that the whole button is supposed to flip, where the background is white and the text is black. We need two colors then, white and black. Since we deprecated `on="dark"` , the only information the CTA has to know that it is on a dark background is the broadcast variables. The problem is that the card only provides one real broadcast color which is white. So we don't actually have a variable available to use to set the text color with.
> 
> We also have a mix'n'match approach to colors still, where some pfe-color attributes requesting surface colors and other attributes requesting ui colors, which might be unintuitive for people applying the attributes and guessing what color they are going to get.
> 
Thanks for reviewing, @starryeyez024. The first issue is how to address the secondary on dark backgrounds, which is the oddball of the cta world since the colors flip. Seems like the only solution would be to add new broadcast custom properties that would only affect secondary CTAs? It strikes me as a lot of overhead but what other choice do we have? The second issue is around the secondary cta color variables. We have most of our cta variables pointing to the `ui` color families. In the case of the secondary CTAs we are using the `surface` family to achieve the black color, since ui doesn't quite have anything close. Do we map the secondary color palette to the `ui-complement` family (and then change that in our theme repo)? `ui-complement` is quite a bit lighter than the specified color for secondary cta text and borders (#464646 compared to #151515).

elements/pfe-cta/src/pfe-cta.scss
```diff
@@ -1,200 +1,359 @@
...
+  --pfe-cta--BackgroundColor:           transparent;
+  --pfe-cta--BorderColor:               var(--pfe-broadcasted--color--text, #{pfe-color(ui-complement)});
+  --pfe-cta__inner--BorderColor:        transparent;
+  --pfe-cta--Color:                     var(--pfe-broadcasted--color--text, #{pfe-color(ui-complement)});
+
+  --pfe-cta--BackgroundColor--hover:    #{pfe-color(ui-complement)};
+  --pfe-cta--BorderColor--hover:        var(--pfe-broadcasted--color--text, #{pfe-color(ui-complement)}); //white, then black
+  --pfe-cta--Color--hover:              #{pfe-color(ui-complement--text)};
+
+  --pfe-cta--BackgroundColor--focus:    #{pfe-color(surface--lighter)};
```
chrisdo1: @starryeyez024 this is the only color (it's a pale gray, #f0f0f0) for which we don't have anything close in the `ui-complement` family. Should I add a `ui-complement--focus` variable for this?

starryeyez024: I'm feeling like we could merge this as-is since it works, and then go back and update color variables here?

chrisdo1: Okay, works for me.

package.json
mwcz: #90

examples/index.html
```diff
@@ -22,7 +22,9 @@ <h1>Demos</h1>
 					<li><a href="../elements/pfe-icon/demo">pfe-icon</a></li>
 					<li><a href="../elements/pfe-icon-panel/demo">pfe-icon-panel</a></li>
 					<li><a href="../elements/pfe-layouts/demo">pfe-layouts</a></li>
+					<li><a href="../elements/pfe-link-list/demo">pfe-link-list</a></li>
 					<li><a href="../elements/pfe-markdown/demo">pfe-markdown</a></li>
+					<li><a href="../elements/pfe-navigation/demo">pfe-navigation</a></li>
```
mwcz: The changes in this file look like they're holdovers from another branch.

elements/pfe-cta/src/pfe-cta.js
```diff
@@ -13,42 +13,120 @@ class PfeCta extends PFElement {
...
+
+  _basicAttributeChanged(attr, oldValue, newValue) {
+    this[attr].value = newValue;
+  }
 }
 
 PFElement.create(PfeCta);
 
-export default PfeCta;
+export { PfeCta as default };
```
mwcz: Just curious, why this change?

examples/index.html
chrisdo1: @mwcz Thanks for reviewing. Removed pfe-navigation and pfe-link-list from the examples index file.

elements/pfe-cta/src/pfe-cta.js
starryeyez024: not sure; @kylebuch8 ?

kylebuch8: @KamiQuasi is requesting that we do this. And I've seen some posts on why this is better.

kylebuch8: That said, should we just leave this change out of this PR and do one big change where we update them all at the same time?

starryeyez024: removed

mwcz: Cool.  I'm not that familiar with arguments for the `export { Foo as default }` style.  I am ready to be swayed.  I just like the sweet simple syntax of `export default Foo`.

PR #354, opened 2019-03-08
Change Spandx port to auto

alexmarple:
After running into an issue booting up a repo with the hardcoded localhost port that Spandx uses (due to Docker blocking it), it was suggested that the port be set to auto in an effort to find _any_ free localhost port.


castastrophe:
@alexmarple Can you add a little more in your description area about the driver behind the PR for me?  Thanks!

alexmarple:
@castastrophe Let me know if that looks right? I suspect it'll take me a few tries to make sure I have the writing conventions down for this team (or group of teams)!

castastrophe:
Thanks @alexmarple!  Have you already tested that `npm run start`, `npm run dev`, and `npm run storybook` are still working as expected with this change?

Issue #360, opened 2019-03-14
Band body + grid + collapsing margins


starryeyez024:
The band section `.pfe-band__body` uses `display:grid`, with a `grid-column-gap` to space out items inside that section. The problem is that we cannot reliably assume anything about the way the paragraph (or other elements) are controlling their margins. 

Even if there are no extra styles on the page, basic paragraphs come with their own margins from the browser, so this doubles up the spacing on regular paragraphs. Normally the top and bottom margins are cancelled out by [collapsing margins](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing), but the effect goes away when grid is introduced. So you end up with a bottom margin from the first paragraph, a grid gap, and a top margin from the second paragraph.

![paragraphs in the band](https://user-images.githubusercontent.com/456863/54325481-c6ad2800-45d8-11e9-9af0-28c91fdf1ba2.gif)

The band demo/index.html file has some inline styles that negate this issue:
```
    p (and headers) {
      margin-top: 0;
    }

    p:last-child (and headers) {
      margin-bottom: 0;
    }
```
so its not visible on that page, but I don't think we can rely on these styles being present in a pre-existing page:

I think the solution is easy enough though, if we remove `display:grid`, `display:flex` (and related properties) from the `.pfe-band__body` section, it will go back to being neutral.

castastrophe:
Updating this issue with the resolution we came up with - we actually didn't remove grid from the band body region.  The guidance states that the body region should use an `article` or `div` tag to contain text elements.

PR #361, opened 2019-03-14
#issue-360: resolve grid gap issue wtth band body

starryeyez024:
When trying to unravel issue #360, I noticed that in the band component, the compiled css contained this style:
```
::slotted(*) {
    margin-bottom: 0;
}
::slotted(*) {
  margin-bottom: var(--pfe-band--gutter--vertical);
}
```
which (I think?) is coming from this sass:
```  
  @include pfe-slot($slot-name, "> *") {
    margin: 0;
  }
```
which lead me to wonder, is there much to gain from abstracting selectors? I was searching and searching for `slotted` but couldn't find it because it was   `@include pfe-slot()` instead.  So I thought maybe this is one of the levels of abstraction we can simplify? I understand the looping was to reduce replicating code, but it was also generating css selectors that don't work: 
 `::slotted([slot="pfe-band--header"] > *:not(:last-child)) `. Side note for the future, this [selector](https://codepen.io/starryeyez024/pen/BbJPMO?editors=1010) (line 59) does work: `[name="title"]::slotted(*:not(:last-of-type))` 

So, the explanation is almost longer than the code, but I wanted to explain the why for this approach, and why I removed the slot mixin. :)


castastrophe:
I was already planning to deprecate the slot mixin.  The goal behind was not to simplify the code but rather, slots are really confusing about how to structure the selector so the goal was to abstract that so that if we find a better way to structure our slotted references, we could update that in the mixin and have it propagate everywhere.  That said, I just couldn't make the mixin work well.

castastrophe:
So +1 slot mixin deprecation but I'd push back on removing all the loops from our sass like this.  It's not that much abstraction to say, for the following regions - header, footer, body, aside - do this same thing.

starryeyez024:
@castastrophe when I was trying to add the fallback margins and test in IE11, I noticed two things:

1.  moving the layout styles to shadowDOM classes means IE11 cant see any of those styles (there is no shadow dom? at least the inspector doesn't think so)
2. I can view the pfe-band demo page in IE11, in this branch, but not in master (can you reproduce this?) I wonder if the slot styles are what's breaking the band in master?

#1 has major implications in terms of how we style things and how they work (or don't in IE11) if that is true. cc @markcaron @kylebuch8

starryeyez024:
> @castastrophe when I was trying to add the fallback margins and test in IE11, I noticed two things:
> 
> 1. moving the layout styles to shadowDOM classes means IE11 cant see any of those styles (there is no shadow dom? at least the inspector doesn't think so)
> 2. I can view the pfe-band demo page in IE11, in this branch, but not in master (can you reproduce this?) I wonder if the slot styles are what's breaking the band in master?
> 
> #1 has major implications in terms of how we style things and how they work (or don't in IE11) if that is true. cc @markcaron @kylebuch8

this is what the band looks like in IE11, when i can get it to load (after i added the band, the page throws errors about a script taking too long to load)
https://www.evernote.com/l/AA62Bbjjq3JGFL5r1h2zp9DmmPqfknGC58k

elements/pfe-band/src/pfe-band.scss
```diff
@@ -99,46 +99,44 @@ $LOCAL: band;
...
-    $slot-name: "pfe-band--#{$slot-name}";
-    $grid-area: $slot;
-  }
-  .pfe-band__#{$grid-area} {
-    // Set up the flex styles for each region
-    display:     flex;
-    flex-flow:   column nowrap;
+.pfe-band__body {
+    display: block;
+    margin-bottom: #{pfe-local(gutter--vertical)};
```
starryeyez024: @starryeyez024  @todo
update readme
add back display:grid to body

elements/pfe-band/README.md
```diff
@@ -7,10 +7,35 @@ This container element provides a set of slots in which to render banded content
...
 
 - `pfe-band--header`: This slot renders at the top of the container and generally contains the title, headline, and or subheadline content.  Other possible candidates include a set of social sharing links or tags that describe the content below. The recommended tag for this slot is the `header` tag with h-level or p tags contained within it.
-- **Default slot**: This is the default slot and contains the bulk of the content in this element.  Paragraph text or a grid of cards or images might be rendered in this region.  Recommended tag for this slot is the `article` tag. Any content not assigned to a slot will be rendered here.
+- **Default slot**: This unnamed slot should contain the bulk of the content in this element.  The recommended wrapper within this slot is the `article` tag, but a div could also work to contain text elements like paragraphs. If you are rendering non-text components such as cards or images, no wrapper is necessary, and these items will make use of the grid layout.  Any content not assigned to a named slot will be rendered here.
```
castastrophe: Since article is in code syntax, should we add the tick marks around `div` too for consistency?

elements/pfe-band/README.md
```diff
@@ -7,10 +7,35 @@ This container element provides a set of slots in which to render banded content
...
+    <article>
+        <p>Consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata dolor sit amet.</p>
+    </article>
+
+    <pfe-card slot="pfe-band--aside">
+        <h3 slot="header">Aside: left body bottom</h3>
+        <p>Ut wisi enim ad minim veniam.</p>
+    </pfe-card>
+
+    <footer slot="pfe-band--footer" class="custom-footnote">
```
castastrophe: Can we remove the custom class from the code example?

elements/pfe-band/src/pfe-band.scss
```diff
@@ -99,46 +99,43 @@ $LOCAL: band;
...
-      margin-bottom: 0;
+}
+
+@supports (display: grid) {
+    .pfe-band__header,
+    .pfe-band__body,
+    .pfe-band__aside,
+    .pfe-band__footer {
+        display: grid;
+        grid-row-gap: var(--pfe-band--gutter--vertical);
```
castastrophe: Why not use the `pfe-local` mixin?

```
grid-row-gap: #{pfe-local(gutter--vertical)};	
grid-column-gap: #{pfe-local(gutter--horizontal)};
```

elements/pfe-band/README.md
starryeyez024: good catch

elements/pfe-band/README.md
starryeyez024: added

Issue #355, opened 2019-03-11
Add the `docs/resources/_gen` folder to gitignore


castastrophe:
Seems like running Hugo is causing some compiled assets to diff:

```
	modified:   docs/resources/_gen/assets/scss/src/sass/kube.scss_48b060fe05b0a273d182ef83c0605941.content
	modified:   docs/resources/_gen/assets/scss/src/sass/kube.scss_48b060fe05b0a273d182ef83c0605941.json
```

Need to ignore these or flag them as binaries.

castastrophe:
This is solved by https://github.com/patternfly/patternfly-elements/pull/373

PR #373, opened 2019-03-21
silence the constant diffs from Kube!

starryeyez024:


castastrophe:
This solves https://github.com/patternfly/patternfly-elements/issues/355

PR #379, opened 2019-03-27
Documenting release steps

castastrophe:
I know there was some talk of not documenting these steps but since you have to have credentials to push to our repos, I think we're safe.  Always nice to have a solid reference document too!


docs/content/develop/step-4.md
```diff
@@ -9,6 +9,31 @@ menu = "develop"
...
+    - Choose the appropriate version bump type for the release you're publishing:
+        - if bumping a prerelease version (example: from 1.0.0-prerelease.2 to 1.0.0-prerelease.3), choose *Custom Prerelease*
+4. Assign that version to this variable for use below: `$NEW_VERSION: 1.0.0-prerelease.3`.
+5. Create a new branch: `git checkout -b release/$NEW_VERSION` (example: release/1.0.0-prerelease.3).
+5. Run a fresh install, build all the assets, and then force add those assets for the tag release:
+    - `npm install && npm run build && git add elements/*/*.{js,map,css} -f`
+6. Commit the updates: `git commit -am "$NEW_VERSION"`
+7. Roll the tag: `git tag "v$NEW_VERSION"`
+8. Remove the bundle files:
+    - `for e in elements/*; do git rm -f elements/$(basename $e)/$(basename $e).{js,css,map}; done`
```
castastrophe: @mwcz This doesn't quite work for us anymore since some of our assets don't have the same name as the base element - do you have a suggested update for this string?

docs/content/develop/step-4.md
```diff
@@ -9,6 +9,31 @@ menu = "develop"
 tags = [ "develop" ]
 +++
 
+We've been publishing our PatternFly Elements to the [@patternfly organization](https://www.npmjs.com/org/patternfly) on npm.
 
+To roll a new release, use the following steps:
 
-We've been publishing our PatternFly Elements to the [@patternfly organization](https://www.npmjs.com/org/patternfly) on npm.
+1. Start in the root directory of the patternfly-elements project.
+2. Check out the master branch and pull down the latest: `git checkout master && git pull`.
```
starryeyez024: should we include some other safeguards here, such as: 
` git reset --hard && git clean -df`

docs/content/develop/step-4.md
```diff
@@ -9,6 +9,31 @@ menu = "develop"
...
+7. Roll the tag: `git tag "v$NEW_VERSION"`
+8. Remove the bundle files:
+    - `for e in elements/*; do git rm -f elements/$(basename $e)/$(basename $e).{js,css,map}; done`
+9. Commit: `git commit -am "Remove bundles from $NEW_VERSION"`
+10. Push up to origin: `git push origin release/$NEW_VERSION -u`
+11. Push up the tag: `git push --tags`
+12. Checkout master and reset:
+    - `git checkout master`
+    - `git reset --hard origin/master`
+13. Create a PR for the branch you just created: [https://github.com/patternfly/patternfly-elements/compare](https://github.com/patternfly/patternfly-elements/compare)
```
starryeyez024:
```suggestion
13. [Create a pull request](https://github.com/patternfly/patternfly-elements/compare) for the branch you just created: 
```

docs/content/develop/step-4.md
```diff
@@ -9,6 +9,31 @@ menu = "develop"
...
+14. Delete branch after merging PR.
```
starryeyez024:
```suggestion
14. Delete branch after merging PR. 
`git branch -d release/$NEW_VERSION`
```

docs/content/develop/step-4.md
```diff
@@ -9,6 +9,31 @@ menu = "develop"
...
+15. To publish to npm (you will have to have the right credentials to do this):
+    - `git checkout v$NEW_VERSION`
```
starryeyez024: I don't understand this part, would this be `git checkout -b release/$NEW_VERSION` ? Would the branch still be there if we just told them to delete it in step 4?

castastrophe: You're checking out the tag and not the branch so this is correct.

docs/content/develop/step-4.md
castastrophe: Applied this with a little styling tweak so didn't commit as-is.

docs/content/develop/step-4.md
castastrophe: Same as above - applied but with a style fix.

docs/content/develop/step-4.md
castastrophe: Applied but with style fix

docs/content/develop/step-4.md
castastrophe: I added a note to clarify this in the docs too!

docs/content/develop/step-4.md
castastrophe: That for loop you sent me in chat doesn't run on my system either -- it doesn't escape so it doesn't execute 🤷‍♂️ Could you edit this and put in what you think we should use here?

mwcz: Can you try this? `for e in elements/*; do find $e -maxdepth 1 \( -not -name "gulpfile.js" -not -name "rollup.config.js" \) \( -name "*.js" -or -name "*.css" -or -name "*.map" \) -exec git rm -f {} \; ;done`

So ugly, but it works well, at least on Linux.  Very curious if it works on Macs.

mwcz:
I added a bit at the top about the new `npm run release` script.

PR #381, opened 2019-03-28
WIP: Clean up that duplicate sass layer

castastrophe:
- Removes the extra layer of abstraction in the _color variable file by incorporating colors directly into the map
    - Added/updated a color function which makes calling those values directly easier
- Removed deprecated variables since they are not in use in the repo (yay!)
- Reduced variables being printed by abstracting text/ui-link values from each surface color to being connected to themes (light, dark, saturated)
    - Bonus, this means if a user overrides the accent variable in their theme and it's no longer saturated, they can just update the theme property and pick up the right text values.
- Added a new mixin which iterates over surface colors and prints them to a component (this has been redone a lot on multiple components and seems very beneficial)


castastrophe:
My main driver behind the PR is that I think there's too much confusion between surface color and it's theme or the context in which a text or link color is invoked.  We have certain colors supported in our system that our known values (more or less, obviously they can be overwritten), then we support attributes for toggling text and link colors for unknown values (contexts/colors being set outside our system).  These are best done with generic terms like "light", "dark", and "saturated" (we can add more contexts too as we find a need).

We're bloating our printed variables by printing text and link values for every theme AND every color supported when all of the colors match up with a theme at this time.  Again, if we need to add more themes, we can totally do this later.

starryeyez024:
I'm into it! I started a branch like this too, lol. #greatminds 

I wonder though, is there reason to keep a handful of the sass variables that we see repeated over & over, like ui-link? Devil's advocate could say that keeping #06c as a variable preserves the relationship between those variables, so that if someone decides to change the link color (as we have been times), the relationship of the lighter or darker variations remains. 

Other css variables could lean on the PFE "crayon" type sass vars:

image

starryeyez024:
@castastrophe just noting something is missing in terms of broadcast variables:
image

starryeyez024:
> @castastrophe just noting something is missing in terms of broadcast variables:
> image

shoot, nevermind, this is a bug in master :(

starryeyez024:
> shoot, nevermind, this is a bug in master :(

I'm sorry for putting all this in this MR, but just to follow up on this "bug", if I load the redhat-theme stylesheet in the card demo page, it resolves the problem. This could be good to discuss in the FED architecture meeting too. :) 
`    <link href="https://static.redhat.com/libs/redhat/redhat-theme/1.0.0/advanced-theme.css" rel="stylesheet"></link>`

image

elements/pfe-sass/README.md
```diff
@@ -1,3 +1,73 @@
...
+    $colors: (lightest, darkest),
+    $properties: (BackgroundColor)
+);
+```
+
+This would render the following in your component:
+
+```css
+:host([pfe-color="lightest"]) {
+    --pfe-foo--BackgroundColor: var(--pfe-theme--color--surface--lightest, #fff);
```
starryeyez024: foo ^

elements/pfe-sass/README.md
```diff
@@ -1,3 +1,73 @@
...
+  --pfe-broadcasted--color--text: var(--pfe-theme--text, #333);
+  --pfe-broadcasted--color--ui-link: var(--pfe-theme--ui-link, #06c);
+  --pfe-broadcasted--color--ui-link--visited: var(--pfe-theme--ui-link--visited, rebeccapurple);
+  --pfe-broadcasted--color--ui-link--hover: var(--pfe-theme--ui-link--hover, #003366);
+  --pfe-broadcasted--color--ui-link--focus: var(--pfe-theme--ui-link--focus, #003366);
+}
+
+:host([pfe-color="darkest"]) {
+    --pfe-foo--BackgroundColor: var(--pfe-theme--color--surface--darkest, #131313);
```
starryeyez024: foo ^

elements/pfe-sass/README.md
castastrophe: It's documentation so foo is on purpose :)

elements/pfe-sass/mixins/_mixins.scss
```diff
@@ -55,49 +55,12 @@
...
-    --pfe-broadcasted--color--ui-link:            #{pfe-color(surface--lightest--link)};
-    --pfe-broadcasted--color--ui-link--visited:   #{pfe-color(surface--lightest--link--visited)};
-    --pfe-broadcasted--color--ui-link--hover:     #{pfe-color(surface--lightest--link--hover)};
-    --pfe-broadcasted--color--ui-link--focus:     #{pfe-color(surface--lightest--link--focus)};
+  @else if $theme == saturated {
+    --pfe-broadcasted--color--text:               #{pfe-color(text--on-saturated)};
+    --pfe-broadcasted--color--ui-link:            #{pfe-color(ui-link--on-saturated)};
+    --pfe-broadcasted--color--ui-link--visited:   #{pfe-color(ui-link--on-saturated--visited)};
+    --pfe-broadcasted--color--ui-link--hover:     #{pfe-color(ui-link--on-saturated--hover)};
+    --pfe-broadcasted--color--ui-link--focus:     #{pfe-color(ui-link--on-saturated--focus)};
```
starryeyez024: @castastrophe this is what I was thinking about re: my question of this MR & the broadcast colors MR being somewhat intertwingled.  Just verifying: you still prefer to merge the broadcast-colors MR first?

castastrophe: I still want to merge the broadcasted story first.

PR #387, opened 2019-04-02
PFE Markdown fork | Adds schema and storybook utils to demo

castastrophe:
@kylebuch8 I created a fork that adds a schema, updates the storybook file to use it, and adds support for the context variables which can influence the text color if the markdown appears on a dark background.

### DIFF
--- a/.storybook/test-theme.css
+++ b/.storybook/test-theme.css
@@ -1,0 +1,31 @@
+.PFElement {
+    font-family: var(--pfe-theme--font-family);
+    color: var(--pfe-broadcasted--color--text);
+}
+.PFElement a {
+    color: var(--pfe-broadcasted--color--ui-link);
+}
+.PFElement a:visited {
+    color: var(--pfe-broadcasted--color--ui-link--visited);
+}
+.PFElement a:hover {
+    color: var(--pfe-broadcasted--color--ui-link--hover);
+}
+.PFElement a:focus {
+    color: var(--pfe-broadcasted--color--ui-link--focus);
+}
+.PFElement > *:first-child {
+    margin-top: 0;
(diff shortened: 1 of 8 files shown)


.storybook/test-theme.css
```diff
@@ -0,0 +1,31 @@
+.PFElement {
```
starryeyez024: can this become 
`@import "https://static.redhat.com/libs/redhat/redhat-theme/latest/advanced-theme.css";`

castastrophe: I think we discussed this in person already but we want to showcase the theme concept but not link to the Red Hat theme in the open source project.

castastrophe: This code is the minimum to make broadcast work without overriding local styles.

elements/pfe-markdown/demo/index.html
```diff
@@ -9,6 +9,9 @@
     </noscript>
 
     <link href="../../pfelement/pfelement.min.css" rel="stylesheet">
+    
+    <!-- Test using the Red Hat theme -->
+    <link href="https://static.redhat.com/libs/redhat/redhat-theme/latest/advanced-theme.css" rel="stylesheet">
```
castastrophe: @castastrophe Note to self to remove this.

elements/pfe-markdown/src/pfe-markdown.scss
```diff
@@ -6,3 +6,11 @@
 :host([hidden]) {
   display: none;
 }
+
+:host([on="dark"]) {
```
kylebuch8: @castastrophe Are we still using `on` attributes?

PR #388, opened 2019-04-02
release 1.0.0-prerelease.11

mwcz:
This merges in the version number updates from release 1.0.0-prerelease.11


elements/pfelement/reveal.js
```diff
@@ -1,51 +0,0 @@
-/*
```
starryeyez024: @mwcz did we intend to remove reveal.js ?

mwcz: Yes, it's a compiled asset that shouldn't have been there to begin with.  @castastrophe noticed that it was a stowaway. :)

mwcz: Oh and it isn't gone for good, it lives in `elements/pfelement/src/reveal.js`.  This just removed the compiled version.

starryeyez024: Listen, Gandalf Taps Musically  👍

PR #394, opened 2019-04-15
on branch checkout, look for leftover elements from other branches

mwcz:
This PR adds a script that is run every time a branch is checked out.  The script looks for any elements which were created in another branch but do not exist in the current branch.

I think this scenario has happened to all of us.

 - look at `elements/`
 - see some element you created a month ago and abandoned
 - wonder *WHY IS THAT STILL HERE?!*

The reason element directories stay around is that within them is a node_modules directory that is not tracked by git.  When you `git checkout my-branch`, it will delete any files that aren't in `my-branch`.  The catch is that it only deletes tracked files.  Untracked files are totally unknown by git and it refuses to delete files you didn't tell it to pay attention to.

This PR will resolve that issue by notifying you of any leftover elements.

## Testing this PR

Here's a good way to test it:

```
# get the PR branch
git checkout master
git fetch origin
git checkout -b leftover-element-check origin/leftover-element-check

# npm install, this will set up the post-checkout git hook which runs the script
npm install

# check out a new branch (you may or may not see leftover element warnings here.  if you have leftover elements already you will see them here)
git checkout -b pfe-foo

# create a new element on your new branch and name it pfe-foo (other answers don't matter)
npm run new

# commit the new element to the pfe-foo branch
git add .
git commit -m "add pfe-foo for testing"

# now go back to the leftover-element-check branch, where pfe-foo does not exist
git checkout leftover-element-check
```

After the last step, you should see a warning that `pfe-foo` was leftover and it is safe to be deleted:

```
WARNING: Found leftover element: elements/pfe-foo

This element were likely created on another branch, but do not exist in your current branch.  Git did not delete it when you switched branches because they contain a `node_modules` directory which is not tracked by git.  They can be safely deleted if they are causing you any issues.    
```

## Auto-deletion, but not yet

IF this PR works well for everyone, we can then consider auto-deleting the leftover elements.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2624,7 +2624,7 @@
 		},
 		"@storybook/podda": {
 			"version": "1.2.3",
-			"resolved": "http://registry.npmjs.org/@storybook/podda/-/podda-1.2.3.tgz",
+			"resolved": "https://registry.npmjs.org/@storybook/podda/-/podda-1.2.3.tgz",
 			"integrity": "sha512-g7dsdsn50AhlGZ8iIDKdF8bi7Am++iFOq+QN+hNKz3FvgLuf8Dz+mpC/BFl90eE9bEYxXqXKeMf87399Ec5Qhw==",
 			"dev": true,
 			"requires": {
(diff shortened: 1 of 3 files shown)


kylebuch8:
@mwcz - I followed your directions and did not get a warning. :(

kylebuch8:
You can see in the screenshot that the script did run. It just didn't give me the warning.

Screen Shot 2019-04-15 at 5 22 00 PM

mwcz:
Hmm, interesting.  Let's take a look tomorrow.

PR #402, opened 2019-04-29
adding a mutationobserver to pfe-tabs

### LINKED ISSUE #332: pfe-content-set and/or pfe-accordion both need a mutation observer
Since we are moving the contents of the Light DOM into the Shadow DOM when we upgrade the component to either pfe-accordion or pfe-tabs, we need to put something in place that watches the light DOM for changes so the shadow DOM is updated accordingly.

kylebuch8:
Adding a mutation observer so the component will continue working properly if new tabs and panels are added.

Fixes #332

### DIFF
--- a/elements/pfe-tabs/src/pfe-tabs.js
+++ b/elements/pfe-tabs/src/pfe-tabs.js
@@ -142,15 +142,10 @@ class PfeTabs extends PFElement {
     super(PfeTabs);
 
     this._linked = false;
-
-    this._onSlotChange = this._onSlotChange.bind(this);
+    this._init = this._init.bind(this);
     this._onClick = this._onClick.bind(this);
-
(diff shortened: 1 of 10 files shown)


elements/pfe-tabs/src/pfe-tabs.js
```diff
@@ -264,7 +258,15 @@ class PfeTabs extends PFElement {
     this._selectTab(tab);
   }
 
-  _onSlotChange() {
+  _init() {
+    if (!this.hasAttribute("role")) {
+      this.setAttribute("role", "tablist");
```
mwcz: Maybe force this to be tablist instead of allowing custom?

kylebuch8: Fixed!

mwcz:
Leaf Gives The Munchies

PR #405, opened 2019-05-02
WIP: Adjustments to abstract heading values

castastrophe:


elements/pfe-text/demo/index.html
```diff
@@ -33,8 +33,8 @@ <h1>&lt;pfe-text&gt;</h1>
 
        <pfe-band pfe-color="lighter">
         <pfe-text slot="pfe-band--header">
-          <p pfe-text="headline-2">Band title: lorem ipsum</strong></p>
```
castastrophe: @starryeyez024 I wanted to get your take on this idea.  What if we abstract the attributes away from numbers to avoid confusion with markup and keep with the alpha, beta, gamma, naming?

elements/pfe-text/src/pfe-text.scss
```diff
@@ -1,6 +1,5 @@
-@import "../../pfe-sass/red-hat-font";
```
castastrophe: @starryeyez024 I think it's important the upstream not invoke Red Hat too much so I think it would be good to leave the pfe theme absent of the Red Hat font (our theme variables will invoke it anyway so we're covered).

elements/pfe-text/src/pfe-text.scss
```diff
@@ -43,50 +47,69 @@
...
     font-family: #{pfe-var(font-family--heading)};
-    line-height: #{pfe-var(line-height--heading)};
     margin-bottom: calc(#{pfe-var(content-spacer)}  * 1.1);
 }
 
 /* size */
+::slotted([pfe-heading="xxl"]) {
+    font-size: #{pfe-var(font-size--heading--xxl)};
+    font-weight: #{pfe-var(font-weight--light)};
+    line-height: #{pfe-var(line-height--heading)};
```
castastrophe: The line-height from the design system is a bit more nuanced - the alpha has 1.3 but every other heading is 1.2.

elements/pfe-sass/variables/_maps.scss
```diff
@@ -161,13 +161,15 @@ $pfe-vars: (
...
-  line-height--heading:          1.3,
-  line-height--heading-xl:       1.2,
+  line-height--heading:          1.2,
+  line-height--heading--alpha:   1.3,
 
   font-family:                   '\"Red Hat Text\", \"Overpass\", Overpass, Helvetica, helvetica, arial, sans-serif',
   font-family--heading:          '\"Red Hat Display\", \"Overpass\", Overpass, Helvetica, helvetica, arial, sans-serif',
   font-family--code:             '\"Overpass Mono\", Consolas, Monaco, \"Andale Mono\", monospace',
 
+  font-size--heading--xxl:       48px,
```
castastrophe: I added the xxl and xl sizes

elements/pfe-sass/variables/_maps.scss
```diff
@@ -161,13 +161,15 @@ $pfe-vars: (
 
   font-size:                     16px,
   line-height:                   1.5,
-  line-height--heading:          1.3,
-  line-height--heading-xl:       1.2,
+  line-height--heading:          1.2,
+  line-height--heading--alpha:   1.3,
```
castastrophe: Not sure about this naming, maybe we keep it as xl, I was just trying to sort it out first in the code.

elements/pfe-text/demo/index.html
starryeyez024: I can jive with that :)

elements/pfe-text/src/pfe-text.scss
starryeyez024: ah yes, didn't mean to commit this, just wanted to test with it. I'll leave this comment open so I remember to remove that

PR #406, opened 2019-05-03
adding a mutationobserver to pfe-content-set

kylebuch8:


kylebuch8:
@castastrophe In commit https://github.com/patternfly/patternfly-elements/pull/406/commits/dbbc1d3aa84af165b34cdfc7d88803e40ee14eaf I added most of the comments back. I left just a few of the self explanatory comments out.

mwcz:
Somehow prettier was not run on `pfe-content-set.js`.

elements/pfe-content-set/src/pfe-content-set.js
```diff
@@ -130,13 +150,14 @@ class PfeContentSet extends PFElement {
       tabs.setAttribute("pfe-variant", this.variant.value);
```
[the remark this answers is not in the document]
kylebuch8: @mwcz - This was fixed here: https://github.com/patternfly/patternfly-elements/pull/406/commits/29ed87b52e53d4ce6c87ace781069e990608bff6

elements/pfe-content-set/src/pfe-content-set.js
```diff
@@ -31,94 +31,114 @@ class PfeContentSet extends PFElement {
 
   constructor() {
     super(PfeContentSet, { delayRender: true });
+
+    this._init = this._init.bind(this);
```
castastrophe: Perhaps this needs to be in pfelement with a setting like `delayRender: true, mutationObserver: true`?

kylebuch8: I think it's fine where it is for now. MutationObserver is something we should address post 1.0.

PR #411, opened 2019-05-16
Update Storybook utils to support self-closing tag

castastrophe:
Current storybook tools parse html in order to inject the slot tags appropriately and allow us to use an object definition for tags.  This PR adds support for self-closing tags to those utilities such as ` <br/> <hr/>` etc. that do not contain "content".

### DIFF
--- a/.storybook/utils.js
+++ b/.storybook/utils.js
@@ -68,26 +68,25 @@ const listProperties = (obj) =>
 // -- attributes: passed through the listProperties function
 // -- content: Accepts html or plain text or renders default content
 export function customTag(obj) {
+  // Most common self-closing tags = br, hr, img, input, link
+  let selfClosing = ["br", "hr", "img", "input", "link"];
   let start = "";
   let end = "";
 
   // If a tag is defined, or it has slots or attributes to apply
   // render an open and close tag
   if (obj.tag || obj.slot || obj.attributes) {
-    start += "<";
-    end += "</";
     // If a tag is defined, use that, else use a div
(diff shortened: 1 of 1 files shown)


.storybook/utils.js
```diff
@@ -108,9 +107,9 @@ const parseMarkup = string => {
     // If results remain in the array, get the attributes
     if (result.length > 1 && typeof result[2] === "string") {
       // Break the attributes apart using the spaces
-      let attr = result[2].trim().split(" ");
+      let attr = result[2].trim().match(/[\w|-]+="[^"]+"/g);
```
castastrophe: Note: this split was losing attributes that had more than one value.  So `foo="bar bat baz"` would be split to`["foo=\"bar", "bat", "baz\""].  No good!

.storybook/utils.js
```diff
@@ -150,9 +150,10 @@ const renderSlots = (slots = []) =>
         typeof slot.attributes !== "undefined" &&
         Object.keys(slot.attributes).length > 0;
       if (!has_tag && (has_slot || has_attr)) {
-        Object.assign(slot, parseMarkup(slot.content));
+        let parsed = parseMarkup(slot.content);
+        Object.assign(slot, parsed);
       }
-      return slot.content
+      return slot.content || slot.tag && selfClosing.includes(slot.tag)
```
castastrophe: Self-closing tags do not have "content" so we check for content *or* if a tag is defined and it's a self-closing one.

PR #415, opened 2019-05-24
add 'viewport' meta tag to pfe-content-set demo

mwcz:
This adds `<meta name="viewport" value="...">` to the pfe-content-set demo page.  Without it, the content set fails to detect the actual width of the screen in some situations (like devtools responsive mode).


mwcz:
Screenshot from 2019-05-24 14-08-55

PR #417, opened 2019-05-29
Fix image distortion in cards

markcaron:
When images are stretched with both Flexbox (inside the card) and margins, images get distorted. Adding `object-fit` to allow the image to fill itself proportionally. Also, when images don't have `pfe-overflow` they stretch due to the Flexbox on the parent—when in some cases the author may not want it stretch 100%. So, adding `align-self: flex-start` to regular `img`s in the slot. When `pfe-overflow` attribute is present we can reset `align-self` with `auto` to give it back the 100% Flexbox stretch.


starryeyez024:
@markcaron  Just asking the edge case question -- if the image doesn't fit the available space in the card, is it better to stretch and crop the image? Or to not enlarge it and just center it in the available space?  This is a little less edge case than we might think, often devices just below the 768px width stack everything, so cards really would be this wide.
image

markcaron:
@starryeyez024, I'd think it's best to not adjust the size and center it. But, if the dev/user wanted it stretched (as an attribute or something), then we should stretch it.

PR #419, opened 2019-05-30
moving this.button initialization to the constructor

### LINKED ISSUE #418: pfe-accordion DOM APIs don't work in Angular's ngAfterViewInit callback
In an Angular app, if you try to call one of the public methods on pfe-accordion in the ngAfterViewInit callback, you get an error. For example, calling `this.accordion.nativeElement.toggle(0);` produces the following error in the console:

ERROR TypeError: Cannot read property 'setAttribute' of undefined

It appears that `this.button` is undefined in the expanded setter.
https://github.com/patternfly/patternfly-elements/blob/master/elements/pfe-accordion/src/pfe-accordion.js#L430

I think this can be fixed by moving the initialization of `this.button` to the constructor instead of being in the `_init` method.

https://github.com/patternfly/patternfly-elements/blob/master/elements/pfe-accordion/src/pfe-accordion.js#L450

kylebuch8:
Fixes #418

### DIFF
--- a/elements/pfe-accordion/package-lock.json
+++ b/elements/pfe-accordion/package-lock.json
@@ -1,18 +1,18 @@
 {
   "name": "@patternfly/pfe-accordion",
-  "version": "1.0.0-prerelease.11",
+  "version": "1.0.0-prerelease.15",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
     "@patternfly/pfe-sass": {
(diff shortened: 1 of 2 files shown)


kylebuch8:
@mwcz The only way to test this is for you to take a look at an Angular app on my local.

mwcz:
Limb Gone Thigh Missing

PR #421, opened 2019-06-05
pfe-navigation component

castastrophe:
This creates a new component for primary navigation (for now, we could extend it for secondary navigation styles in future).  This includes a region for logo, main navigation, utility navigation (specifically, search, login, and language).

https://rally1.rallydev.com/#/detail/userstory/285742576332?fdp=true


elements/pfe-navigation/src/pfe-navigation.html
```diff
@@ -0,0 +1,55 @@
+<div class="pfe-navigation__wrapper">
+  <div class="pfe-navigation__container">
+    <slot name="logo" class="pfe-navigation__logo"></slot>
+    <slot name="skip" class="pfe-navigation__skip"></slot>
+    <slot name="main" class="pfe-navigation__main"></slot>
+    <div class="pfe-navigation__utility">
+      <pfe-navigation-item pfe-icon="menu">
+        <h2 slot="trigger">Menu</h2>
+        <div slot="tray" hidden>
```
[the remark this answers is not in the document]
castastrophe: When we did discovery, we said that if JavaScript was not turned on, the navigation will render a direct link to the landing page.  Top-level navigation only.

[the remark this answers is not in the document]
castastrophe: There's no way to make all the navigation trays visible if the JS isn't loaded that wouldn't overwhelm the user.

elements/pfe-navigation/src/pfe-navigation.scss
```diff
@@ -0,0 +1,195 @@
...
+  padding-right: 16px !important;
+  padding-bottom: #{pfe-local(Padding--vertical)} !important;
+
+  // min-width: 80px;
+  max-height: 72px;
+}
+
+::slotted([slot="skip"]) {
+  @include pfe-sr-only;
+}
```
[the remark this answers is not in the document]
castastrophe: Skip nav is not for sighted users so it should never be visible.

elements/pfe-navigation/src/pfe-navigation-item.scss
```diff
@@ -0,0 +1,148 @@
...
+    position: relative;
+
+    display: flex;
+    flex-flow: column nowrap;
+    align-items: center;
+    justify-content: center;
+    flex-grow: 1;
+
+    background-color: #{pfe-local(BackgroundColor)};
+    padding: 20px 16px !important;
```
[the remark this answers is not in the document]
castastrophe: The mock-ups also show 20px of padding above and below the nav items so would need clarity on which is more important.

elements/pfe-navigation/src/pfe-navigation.scss
```diff
@@ -0,0 +1,195 @@
...
+
+  // Darkest is the default background color
+  --pfe-navigation--BackgroundColor: #{pfe-color(surface--darkest)};
+  --pfe-navigation--Color: #{pfe-color(surface--darkest--text)};
+  --pfe-navigation--Border: #{pfe-var(surface--border-width)} #{pfe-var(surface--border-style)} transparent;
+  --pfe-navigation--Width: auto;
+  --pfe-navigation--Height: auto;
+
+  // Loop through the other breakpoints
+  @each $size in (md, lg, xl) {
```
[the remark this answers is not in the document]
castastrophe: This is not ready for code review - I'm still working on styles.

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -0,0 +1,431 @@
...
+  _stickyHandler() {
+    if(window.pageYOffset >= this.top) {
+      this.classList.add("sticky");
+    } else {
+      this.classList.remove("sticky");
+    }
+  }
+
+  _outsideListener(event) {
+    if ((event.target !== this && event.target.closest("pfe-navigation") === null) || event.path[0] === this.overlay) {
```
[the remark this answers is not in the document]
castastrophe: I was still coding on this - I did say it was WIP.  You pulled it down as I was coding this section.

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -0,0 +1,431 @@
...
+
+        accordion.appendChild(header);
+        accordion.appendChild(panel);
+      });
+
+      fragment.appendChild(accordion);
+
+      this._menuSlotMobile.innerHTML = "";
+      this._menuSlotMobile.appendChild(fragment);
+    }
```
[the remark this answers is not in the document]
castastrophe: This is still WIP - this is not ready for code review.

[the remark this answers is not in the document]
castastrophe: I only un-WIP'ed it to see if the tests would pass.

elements/pfe-navigation/src/pfe-navigation.scss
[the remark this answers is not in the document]
castastrophe: This should be corrected now - I hadn't implemented the tablet view yet.

elements/pfe-navigation/src/pfe-navigation.js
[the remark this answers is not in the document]
castastrophe: You pulled it down mid-coding so it wasn't ready yet.  It should be in a good place now.

elements/pfe-navigation/demo/index.html
```diff
@@ -0,0 +1,368 @@
...
+      <pfe-navigation-item slot="login" pfe-icon="user">
+        <h3 slot="trigger"><a href="/login">Login</a></h3>
+        <div slot="tray" hidden>
+          <div class="container">
+            <p>Login things</p>
+          </div>
+        </div>
+      </pfe-navigation-item>
+      <pfe-navigation-item slot="site-switcher" pfe-icon="bento" pfe-top-link>
+        <h3 slot="trigger"><a href="/url-to-rh-websites-page">Apps &amp; Websites</a></h3>
```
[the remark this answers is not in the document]
castastrophe: Updated

elements/pfe-navigation/src/pfe-navigation.js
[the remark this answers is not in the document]
castastrophe: This needs to be implemented in the accordion component and I could use an assist on that.

elements/pfe-navigation/src/pfe-navigation.scss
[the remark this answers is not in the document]
starryeyez024: It is also useful for sighted users who are not able to use a mouse, but [only navigate with a keyboard](https://webaim.org/techniques/skipnav/#creating). That way they don't have to tab through 9 menu links on every single page. :)

elements/pfe-navigation/src/pfe-navigation-item.scss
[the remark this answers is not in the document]
starryeyez024: Ah, the 20px spacer actually overlaps the icons if stretched across (will note this with Corey). Since the icon+links are the tallest items, we'll probably want to set padding based on those. 
image

If I change the spacer to about 15px, and subtract the border-top of 4px, its about 10 or 11px.

elements/pfe-navigation/src/pfe-navigation.js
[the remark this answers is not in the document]
starryeyez024: Yep, you noted it was ready for some preliminary branch testing; I was just noting UX issues where I thought it would be applicable in the MR (I tried not to add any comments specific to how code was written). I'll use a google doc next time

elements/pfe-navigation/src/pfe-navigation.scss
```diff
@@ -0,0 +1,194 @@
...
+        width: calc(#{pfe-breakpoint(#{$size})} - #{map-get($pfe-vars, container-spacer) * 2});
+      }
+    }
+
+    display: flex;
+    flex-flow: row nowrap;
+    align-items: stretch;
+    justify-content: space-between;
+
+    padding: 0 calc(#{pfe-var(container-padding)} * 2);
```
starryeyez024: Mockups show 24px of spacing on edges of mobile menu:
image

currently demo page shows 32px on edges:
image

elements/pfe-navigation/src/pfe-navigation.js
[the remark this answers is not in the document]
starryeyez024: Per your note in google chat, this is ready for code review now? Would you like to remove the WIP? thanks!

elements/pfe-navigation/src/pfe-navigation-item.scss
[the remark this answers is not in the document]
castastrophe: So update the nav-item padding top and bottom to....

elements/pfe-navigation/src/pfe-navigation.scss
castastrophe: It does but we don't have a variable for 20px in the design system.  I opted for container-spacer * 2 because it was close.  It looks pretty close to me.  I also think you're looking at an outdated version because your navigation element shows a `slot="[object Object]"`.

starryeyez024: @castastrophe I still see `slot="[object Object]"` 
Can you send me instructions on what steps I might be missing? Here's what I'm doing:
```
git checkout pfe-navigation
git fetch origin && git pull origin && git status  
git show --oneline -s
4e9cd55b (HEAD -> pfe-navigation, origin/pfe-navigation) Replace Red Hat logo with generic placeholder
npm install
npm run start
```
(new terminal window)
```
npm run dev pfe-navigation
```

elements/pfe-navigation/src/pfe-navigation-item.scss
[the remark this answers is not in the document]
starryeyez024: `padding: 10px 16px`

elements/pfe-navigation/demo/demo.css
```diff
@@ -0,0 +1,74 @@
...
+.container {
+  grid-template-columns: 1fr;
+  max-width: var(--pfe-navigation--Width);
+  margin: 0 auto;
+  display: grid;
+  text-align: left;
+}
+
+.test-overflow {
+  height: 100vh;
```
starryeyez024: Is this to show that if theres more content than vertical space for the tray, its still scrollable?

castastrophe: Exactly!  :) I wanted to have a good-enough solution for that edge case because we've seen it happen before with navigation

elements/pfe-navigation/src/pfe-navigation.scss
```diff
@@ -0,0 +1,194 @@
...
+  &__container {
+    margin: 0 auto;
+    width: #{pfe-local(Width)};
+    color: #{pfe-local(Color)};
+    z-index: 2;
+
+    // need to do something here for IE11 since we can't redefine the variable
+    @each $size in (md, lg, xl) {
+      @media screen and (min-width: pfe-breakpoint(#{$size})) {
+        width: calc(#{pfe-breakpoint(#{$size})} - #{map-get($pfe-vars, container-spacer) * 2});
```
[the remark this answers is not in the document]
castastrophe: I believe this is correct now - can you confirm?

elements/pfe-navigation/src/pfe-navigation.scss
castastrophe: This is resolved now

elements/pfe-navigation/src/pfe-navigation-item.scss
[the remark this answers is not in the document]
castastrophe: What about `padding: 8px 16px`?  Based on the DSC chat, they are working towards all metrics being multiples of 8.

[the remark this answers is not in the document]
castastrophe: So it would equate to: variable -> (container-padding / 2) and container-padding

elements/pfe-navigation/src/pfe-navigation.scss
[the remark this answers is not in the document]
castastrophe: I shared this with DSC but since it's hidden on dotCom today - let's leave it hidden and do some discovery on how to approach it generally

elements/pfe-navigation/demo/pfe-navigation.story.js
```diff
@@ -0,0 +1,210 @@
...
+  <pfe-navigation-main>
+    <ul>
+      <li>
+        <pfe-navigation-item>
+          <h2 slot="trigger"><a href="#">Products</a></h2>
+          <div slot="tray" hidden>
+              <div class="container test-overflow">
+                <div columns="3">
+                  <div stacked>
+                    <pfe-link-list>
```
chrisdo1: Does this need to be updated since we removed the pfe-link-list custom element?

elements/pfe-navigation/src/pfe-navigation-item.scss
```diff
@@ -0,0 +1,189 @@
...
+    outline: none;
+
+    background-color: #{pfe-local(BackgroundColor)};
+    padding: var(--pfe-navigation--Padding--vertical) var(--pfe-navigation-item--Padding--horizontal) !important;
+
+    border-top: #{pfe-var(surface--border-width--heavy)} #{pfe-var(surface--border-style)} #{pfe-local(BorderTopColor)};
+    border-right: #{pfe-var(surface--border-width)} dashed #{pfe-local(BorderColor)};
+    border-bottom: #{pfe-var(surface--border-width)} dashed #{pfe-local(BorderColor)};
+    border-left: #{pfe-var(surface--border-width)} dashed #{pfe-local(BorderColor)};
+    color: #{pfe-local(Color)};
```
chrisdo1: This is repeated from line 108.

elements/pfe-navigation/src/pfe-navigation-item.scss
```diff
@@ -0,0 +1,189 @@
...
+        @include pfe-arrow(down, #{pfe-color(feedback--default)}, $width: 6px, $height: 6px) {
+          position: absolute;
+          bottom: 13px;
+        }
+      }
+
+      :host([pfe-icon]) & {
+        margin-bottom: calc(#{pfe-var(container-spacer)} / 2);
+        width: 21px;
+        height: 21px;
```
chrisdo1: Should we set a local variable for these?

castastrophe: Let's leave them hardcoded for now and we can revisit hooks later.

elements/pfe-navigation/src/pfe-navigation-item.scss
castastrophe: +1!  Updated in latest push

elements/pfe-navigation/src/pfe-navigation.scss
```diff
@@ -0,0 +1,182 @@
...
+  z-index: 2;
+  width: 100%;
+}
+
+:host(.sticky) {
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100vw;
+  z-index: 99;
```
chrisdo1: Would it be easy to override this z-index value in a situation where something is higher on the page like an alert?

elements/pfe-navigation/src/pfe-navigation.html
```diff
@@ -0,0 +1,66 @@
...
+    </div>
+    <div class="pfe-navigation__utility">
+      <pfe-navigation-item pfe-icon="menu">
+        <h2 slot="trigger">Menu</h2>
+        <div slot="tray" hidden>
+          <div class="pfe-navigation__utility--mobile">
+            <div class="pfe-navigation__utility--mobile--top">
+              <slot name="mobile-search"></slot>
+            </div>
+            <slot name="mobile-menu"></slot>
```
chrisdo1: I don't see this slot referenced in the schema file.

castastrophe: It's not a slot for users, it's a slot to inject content into dynamically during navigation initialization.  Only the web component puts stuff in there.

elements/pfe-navigation/src/pfe-navigation--lightdom.scss
```diff
@@ -0,0 +1,96 @@
...
+            }
+        }
+    }
+}
+
+pfe-navigation [slot="logo"] > a {
+    color: #{pfe-local($cssvar: Color, $fallback: #{map-get($pfe-colors, text--on-dark)})};
+    text-decoration: none;
+    &:focus {
+        outline: 1px dashed ;
```
chrisdo1: minor: extra space `dashed ;`

elements/pfe-navigation/src/pfe-navigation.scss
chrisdo1: `z-index: var(--pfe-z-index--nav, 99);`

elements/pfe-navigation/src/pfe-navigation.scss
```diff
@@ -0,0 +1,189 @@
...
+@each $icon in (user, globe) {
+  ::slotted([icon="#{$icon}"]) {
+    --pfe-navigation__icon--Background: #{map-get($icons, #{$icon})} no-repeat;
+  }
+}
+
+.pfe-navigation {
+  &__wrapper {
+    position: relative;
+    z-index: 2;
```
chrisdo1: Same question here about setting a variable with a fallback.

elements/pfe-navigation/demo/pfe-navigation.story.js
castastrophe: Yes, it does!

elements/pfe-navigation/src/pfe-navigation.scss
castastrophe: Maybe we add z-indexing as a map in our pfe-sass like we do for webrh?

castastrophe: Added this!  Thanks!

elements/pfe-navigation/src/pfe-navigation.scss
castastrophe: Good idea, I've updated these and added a z-index map

elements/pfe-navigation/src/pfe-navigation--lightdom.scss
castastrophe: Pushing this fix up next.

elements/pfe-navigation/README.md
```diff
@@ -0,0 +1,65 @@
...
+- `logo`: This slot should contain the site's main logo.
+- `search`: This slot, when assigned to a navigation-item component, renders a dropdown in the navigation for the search functionality. Inside the tray for the slot, we recommend tagging the search form or search functionality that includes the input and submit button with the attribute `pfe-navigation--mobile-search`. The element inside the search slot that is tagged with this attribute will be copied into the mobile menu and appear above the accordion for the main navigation.
+- `language`: This slot, when assigned to a navigation-item component, renders the dropdown for the user to select the site language.
+- `mobile-language`: A simple link to a landing page containing language switching information.  Example: `<a href="/url-to-language-page" slot="mobile-language" pfe-icon="user" hidden>English</a>`.
+- `login`: This slot, when assigned to a navigation-item component, renders the dropdown for the user to log into the site.
+- `mobile-login`: A simple link to a landing page containing language switching information.  Example: `<a href="/login" slot="mobile-login" pfe-icon="user" hidden>Login/Register</a>`.
```
chrisdo1: Maybe this should be: 
```
mobile-login: A simple link to a page containing a login form. Example:...
```

elements/pfe-navigation/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Slots
+
+- `trigger`: Accepts a link tag with a fallback URL (used if the JS doesn't upgrade). This slot should be applied to an h-level tag. Example: `<h3 slot="trigger"><a href="http://www.redhat.com">Home</a></h3>`.
+- `tray`: Accepts any light DOM content to be rendered inside the tray. Example: `<div slot="tray" hidden></div>`.  We recommend using the `pfe-navigation-item__tray--container` class on the wrapper element inside the tray. This will give the tray a max-width that matches the navigation tray.
+
+### Attributes
+
+- `pfe-icon`: Currently available icons are "bento", "user", "globe", "search", and "menu".  Adding this attribute renders that icon above the trigger.
+
```
chrisdo1: Maybe it's overkill, but should we include the `hidden` attribute here where we explain that it should always be attached to the tray slot?

elements/pfe-navigation/README.md
```diff
@@ -0,0 +1,65 @@
...
+- `logo`: This slot should contain the site's main logo.
+- `search`: This slot, when assigned to a navigation-item component, renders a dropdown in the navigation for the search functionality. Inside the tray for the slot, we recommend tagging the search form or search functionality that includes the input and submit button with the attribute `pfe-navigation--mobile-search`. The element inside the search slot that is tagged with this attribute will be copied into the mobile menu and appear above the accordion for the main navigation.
+- `language`: This slot, when assigned to a navigation-item component, renders the dropdown for the user to select the site language.
+- `mobile-language`: A simple link to a landing page containing language switching information.  Example: `<a href="/url-to-language-page" slot="mobile-language" pfe-icon="user" hidden>English</a>`.
```
chrisdo1: Do we need the `hidden` attribute here?

elements/pfe-navigation/README.md
chrisdo1: Same question about the `hidden` attribute here.

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -0,0 +1,322 @@
...
+      };
+
+      // Kick off the initialization of the light DOM elements
+      this._init();
+
+      // Listen for tab events
+      this.addEventListener("keydown", this._keydownHandler);
+
+      // Watch for screen resizing
+      window.addEventListener("resize", this._resizeHandler);
```
kylebuch8: Instead of using a "resize" listener, it would be more efficient to use `matchMedia` and `.addListener`. The listener will only fire when the media query conditions are met which is more performant than the callback function firing on every change of the screen size.

https://enmascript.com/articles/2018/10/31/the-underestimated-power-behind-window-matchmedia

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -0,0 +1,322 @@
...
+    this.overlay = false;
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    // If this element contains light DOM, initialize it
+    if (this.children.length) {
+      // If only one value exists in the array, it starts at that size and goes up
+      this.breakpoints = {
```
kylebuch8: If you use `matchMedia`, you could use something like 
```
this.breakpoints = {
  main: window.matchMedia("max-width: 1199px"),
  search: window.matchMedia("min-width: 768px"),
  ...
}
```

And then when you need to check the breakpoint, you just need to use `this.breakpoints.main.matches` which will return true or false.

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -0,0 +1,322 @@
...
+        "mobile-language": [0, 767],
+        login: [768],
+        "mobile-login": [0, 767]
+      };
+
+      // Kick off the initialization of the light DOM elements
+      this._init();
+
+      // Listen for tab events
+      this.addEventListener("keydown", this._keydownHandler);
```
kylebuch8: This listener needs to be disconnected in the `disconnectedCallback`

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -0,0 +1,322 @@
...
+      console.error("This component does not have any light DOM children.  Please check documentation for requirements.");
+    }
+    
+    PFElement._debugLog = false;
+  }
+
+  disconnectedCallback() {
+    // Remove the scroll, resize, and outside click event listeners
+    window.removeEventListener("resize", this._resizeHandler);
+    window.removeEventListener("scroll", this._stickyHandler);
```
kylebuch8: The listener for scroll is added in a condition so we should only remove the event listener if the event was listened for in the first place.

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -0,0 +1,322 @@
...
+    document.removeEventListener("click", this._outsideListener);
```
kylebuch8: The listener for click is added in a condition so we should only remove the event listener if the event was listened for in the first place.

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -0,0 +1,322 @@
...
+    // Initialize active navigation item to empty array
+    this._activeNavigationItems = [];
+    this.overlay = false;
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    // If this element contains light DOM, initialize it
+    if (this.children.length) {
```
kylebuch8: The fact that `this._init()` is only called if this condition is met is going to be problematic with Angular. My experience with Angular applications is that when the `connectedCallback` fires in Angular applications, the children are not present so your call to `_init` would never happen in an Angular app.

Look at https://github.com/patternfly/patternfly-elements/blob/master/elements/pfe-tabs/src/pfe-tabs.js#L161 for the approach we've taken on other components.

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -0,0 +1,335 @@
...
+    };
+
+    // Initialize active navigation item to empty array
+    this._activeNavigationItems = [];
+    this.overlay = false;
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+
```
kylebuch8: We should probably make sure the other components, `pfe-navigation-item` and `pfe-navigation-main`, have been created before we get into the `this.children.length` check. 

Follow `pfe-tabs` as an example.
https://github.com/patternfly/patternfly-elements/blob/master/elements/pfe-tabs/src/pfe-tabs.js#L157

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -0,0 +1,335 @@
...
+  _outsideListener(event) {
+    if ((event.target !== this && event.target.closest("pfe-navigation") === null) || event.path.length > 0 && event.path[0] === this._overlay) {
+      this._activeNavigationItems.map(item => item.close());
+    }
+  }
+
+  _setVisibility(width) {
+    Object.keys(this.breakpoints).forEach(label => {
+      let bps = this.breakpoints[label];
```
kylebuch8: I think if you use matchMedia with `this.breakpoints`, this could probably be simplified.

elements/pfe-navigation/src/pfe-navigation-item.js
```diff
@@ -0,0 +1,429 @@
...
+        if (path.indexOf(window) === -1 && path.indexOf(document) === -1)
+          path.push(document);
+        if (path.indexOf(window) === -1)
+          path.push(window);
+        return path;
+      }
+    });
+}
+
+if (!Array.prototype.filter){
```
kylebuch8: I'm guessing we're making these files larger than they need to be with this `Array.prototype.filter` check. The same check is in pfe-navigation.js so this code is going to be duplicated. Can we remove this one and just rely on the one in pfe-navigation.js?

elements/pfe-navigation/src/pfe-navigation-item.js
```diff
@@ -0,0 +1,429 @@
...
+      isVisible = Boolean(isVisible);
+  
+      if (isVisible) {
+        this.removeAttribute("hidden");
+      } else {
+        this.setAttribute("hidden", "");
+      }
+    }
+  
+    open(event) {
```
kylebuch8: This is super nit picky but can we move these public API methods after the `disconnectedCallback`? It just makes things consistent with the other components.

elements/pfe-navigation/src/pfe-navigation-item.js
```diff
@@ -0,0 +1,429 @@
...
+  
+      this._init__trigger();
+      this._init__tray();
+  
+      // Add a slotchange listeners to the lightDOM elements
+      if (this.trigger) this.trigger.addEventListener("slotchange", this._init__trigger);
+      if (this.tray) this.tray.addEventListener("slotchange", this._init__tray);
+    }
+  
+    attributeChangedCallback(attr, oldValue, newValue) {
```
kylebuch8: Is this necessary if we're not doing anything inside the `attributeChangedCallback`?

elements/pfe-navigation/src/pfe-navigation-item.js
```diff
@@ -0,0 +1,429 @@
...
+        })
+      );
+    }
+  
+    close(event) {
+      if (event) event.preventDefault();
+  
+      // Close the children elements
+      this.navigationWrapper._activeNavigationItems = this.navigationWrapper._activeNavigationItems.filter(item => {
+        let close = this.nestedItems && this.nestedItems.includes(item);
```
kylebuch8: `.includes` is not available in IE11

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

elements/pfe-navigation/src/pfe-navigation-item.js
```diff
@@ -0,0 +1,429 @@
...
+        this._trigger.removeEventListener("keyup", this._keyupHandler);
+      } else {
+        this._trigger.removeEventListener("click", this._navigateToUrl);
+        this._trigger.removeEventListener("keyup", this._directLinkHandler);
+      }
+    }
+  
+    _init__trigger() {
+      // If no slots have been assigned, assign it to the trigger slot
+      const unassigned = [...this.children].filter(child => !child.hasAttribute("slot"));
```
kylebuch8: I'm pretty sure this is going to fail in Angular but I haven't tested it yet. `_init__trigger` is being called in the `connectedCallback` and I'm pretty sure that `this.children` is going to be empty. We may need to look at using a MutationObserver.

kylebuch8: I confirmed that Angular isn't able to find the children.

Screen Shot 2019-09-05 at 6 20 15 PM

elements/pfe-navigation/demo/index.html
```diff
@@ -0,0 +1,414 @@
...
+      <!-- Skip navigation -->
+      <div slot="skip">
+        <ul>
+          <li><a href="#rh-main-content">Skip to content</a></li>
+        </ul>
+      </div>
+      <!-- Site logo -->
+      <h2 slot="logo">
+        <a href="https://redhat.com">
+          <img class="redhat-logo" src="https://via.placeholder.com/150x50.png" class="screen" title="Red Hat"/>
```
kylebuch8: The `class` attribute is duplicated on this line.

elements/pfe-navigation/src/pfe-navigation-item.js
castastrophe: It will not upgrade without the polyfill manually added.  If you can fix the polyfill tool, that would be rad but in the meantime, without these in here, it won't upgrade in Edge/IE11.

elements/pfe-navigation/src/pfe-navigation-item.js
castastrophe: We can but I think it makes sense to keep public-facing functions at the top because those are the ones developers will be looking at most frequently during debugging.

elements/pfe-navigation/src/pfe-navigation-item.js
castastrophe: Probably not - I'll remove.

elements/pfe-navigation/src/pfe-navigation-item.js
castastrophe: I'll add the polyfill

elements/pfe-navigation/src/pfe-navigation.js
castastrophe: :+1: Added this.

elements/pfe-navigation/src/pfe-navigation.js
castastrophe: Updated

elements/pfe-navigation/src/pfe-navigation.js
castastrophe: Updated

elements/pfe-navigation/src/pfe-navigation.js
castastrophe: Cool!  I'd love to learn more about matchMedia for a future update to this component!

.gitignore
```diff
@@ -3,20 +3,14 @@ node_modules
 tmp
 
 # ignore js, map, and css in element dirs
+elements/*/_temp
```
mwcz: May I ask why `_temp` instead of `.temp`?  It seems to me that making the dir hidden would be cleaner.

elements/pfelement/gulpfile.js
```diff
@@ -1,92 +1,4 @@
...
-    "compile",
-    "minify-css",
-    "bundle"
-  )
-);
-
-gulp.task("default", gulp.series("build"));
-
-gulp.task("dev", gulp.series("build", "watch"));
+gulpFactory(pfelementPackage);
```
mwcz: When I created the gulpFactory, it didn't work with pfelement.  I take this diff to mean that it does work now?  YAY!

.gitignore
castastrophe: I think there's value in seeing those folders and the dot naming structure is harder to engage with in a GUI because they're hidden from the user's view.

castastrophe: Finder for example, a user won't even see that the folder exists on a Mac; what if I'm debugging and want to look at the compiled asset.  The only way to do that then becomes the command line and I don't want to lock out users.

elements/pfe-navigation/demo/index.html
castastrophe: Looks like this is outdated now

scripts/gulpfile.factory.js
```diff
@@ -271,18 +244,24 @@ ${fs
...
+      .pipe(dest(paths.temp));
   });
 
-  task("copy", () => {
-    return src(["*.js", `!${elementName}.js`], {
+  task("copy:src", () => {
+    return src(["*.js", `!${elementName}*.js`], {
       cwd: paths.source
-    }).pipe(dest(paths.compiled));
+    }).pipe(gulpif((file) => files.length > 0 && gulpmatch(file, files) || files.length === 0, dest(paths.compiled), dest(paths.temp)));
```
kylebuch8: `pfe-avatar` is not building because of this line. Why wouldn't we want the other js files to also go to the temp directory?

kylebuch8:
> It would be great if the temp dir change was its own PR. It will be a much smaller and easier to review, and can be merged ahead of nav. You can merge it into nav immediately to continue nav work unblocked.

This change is causing the build to break.

scripts/gulpfile.factory.js
kylebuch8: @castastrophe Can you help me understand what this ternary is doing?

elements/pfe-navigation/README.md
castastrophe: +1 updating now

elements/pfe-navigation/README.md
castastrophe: +1 updating now

elements/pfe-navigation/demo/index.html
```diff
@@ -18,9 +18,9 @@
   <link href="../pfe-navigation--lightdom.css" rel="stylesheet">
 
   <!-- uncomment the es5-adapter if you're using the umd version -->
-  <script src="/@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js"></script>
-  <script src="/@webcomponents/webcomponentsjs/webcomponents-bundle.js"></script>
-  <script src="/requirejs/require.js"></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/2.2.10/custom-elements-es5-adapter.js"></script>
```
castastrophe: Why would we use the CDN for the polyfills instead of our locally version-controlled copies?

castastrophe: I would prefer this be reverted to using the local copies.  It ensures we are all using the same version as we test our components and communicates to consumers what versions work with our components.

kylebuch8: @castastrophe - We need this for Netlify. Until we figure out how to get Netlify to use the local web components files, this is the only work around that I can come up with. If I don't have those changes, then this doesn't work.

https://deploy-preview-421--happy-galileo-ea79c4.netlify.com/elements/pfe-navigation/demo/

PR #429, opened 2019-06-17
pfe-accordion | Fix collapse animation

TylerMartin5:
The panel currently appears to collapse before its content disappears. This should feel a little smoother.


castastrophe:
@TylerMartin5 The code for this looks really solid.  I'd like to build this on a sandbox and test it in a few different browsers but after that it's likely +1.  Thanks for doing this!

Issue #435, opened 2019-06-21
variable name mismatch for broadcasted


starryeyez024:
in _functions.scss, we need `--color` added to the function. Currently its
```
@function pfe-radio($cssvar...) {
    @return var(--#{$repo}-broadcasted--#{to-string($cssvar, '--')});
}
```
and files like mixins.scss and card.scss ask for the variables like  ` --pfe-broadcasted--color--text`

Also card.scss should use the new pfe-theme mixin

starryeyez024:
WIP https://github.com/patternfly/patternfly-elements/pull/392/files

castastrophe:
There are broadcasted variables other than color available so I'm not sure I agree with this issue.  If we embed color in the mixin then we can't set text-decoration for example.

starryeyez024:
PR 392 was merged, closing this out

PR #392, opened 2019-04-07
feat: Updating broadcast and how variables are applied

starryeyez024:
## Automated broadcasting of theme context
* Link(s) to demo pages where this element can be viewed:

---

### What has changed and why
Summarize files edited as part of this MR along with a brief description of what was changed/why.

* Broadcast variable names were simplified (`--color` and `ui-` removed)
  - New broadcast variable added for `link-decoration` (this is helpful on saturated themes where the presence of an underline can help indicate a link)
* Demo page updates to preview theme contexts and inheritance
* Removes theme variables that tied text and link colors to specific surface values (these were abstracted to theme-specific text and link values)
* Updates to the following components to support the theme inheritance:
  - Accordion
  - Autocomplete
  - Band (this update uses the new functions and creates a graceful IE11 fallback with white background and black text)
  - Card (this update uses the new functions and creates a graceful IE11 fallback with white background, black text, gray border, with a light gray header background color if the region exists)
  - CTA
  - Icon and Icon panel
  - Modal (note that the shadow template window updates the theme variable to light but the trigger inherits from it's context)
  - Tabs
* Rename pfe-layouts to pfe-styles to reflect the addition of the context stylesheet
* Restructured sass functions and mixins:
  - Added error messaging if value does not exist in the map
  - `pfe-color`, `pfe-var`, `pfe-zindex`, `pfe-broadcasted` all use `pfe-get-from-map` as their source of truth
  - `pfe-local` and `pfe-apply-broadcast` functions updated, bugs fixed
  - New `broadcasted` map, includes `link-decoration` which is typically mapped to `text-decoration` property
  - Duplicate variables removed and values added directly to maps
  - Accordion mixins rewritten and moved to a components file
  - Mixins: `pfe-set-broadcast-theme`, `pfe-print-variables`, `pfe-surfaces`, and `pfe-theme-contexts`
  - Global variable lists added for `$BROADCAST-VARS`, `$THEMES`, `$SURFACE`
  - Default for USE-FALLBACK updated to true
* `pfe-styles`
  - A new contexts css file added to allow use of helper classes `on-dark`, `on-light`, `on-saturated`
  - A new theme-test preview page added to demo folder (this uses pfe-color.css which is a local-only compiled asset, not for publication)
* Documentation updates

### Testing instructions
Be sure to include detailed instructions on how your update can be tested by another developer.

Goals accomplished by this branch:
- Accurate theme variable should be set by data attribute pfe-color
[...]

PR #436, opened 2019-06-21
Fixes Issue #435, broadcast variable name

starryeyez024:
https://github.com/patternfly/patternfly-elements/issues/435


starryeyez024:
this function is re-written in the larger broadcast MR  https://github.com/patternfly/patternfly-elements/pull/392/files#diff-9f65d1cab6eec08b9077732fad5c2573R326

PR #445, opened 2019-06-26
fix pfe-avatar in ie11

mwcz:
Fixes two problems with IE11:

 - `PFelement.var()`, used to get the current value of CSS variables, doesn't work in IE11 (or anywhere with shadyCSS I would bet).  The size of avatars is controlled with a CSS variable so the size wasn't available in IE11.  This adds a fallback of 128x128.
 - IE11 doesn't accept decimals in rgb values (at least not for canvas fillStyle).  To fix, I floored the values.  `rgb(128.555, 16, 71.0)` will now be `rgb(128, 16, 71)`.

IE11 screenshot

win10_ie_11 0


castastrophe:
Browser tested: https://www.browserstack.com/screenshots/ddd5eae180de29d5a523eb63437c6e1a1107a52e ✅

PR #459, opened 2019-07-03
US206196  modal web component

chrisdo1:
https://rally1.rallydev.com/#/270861059696d/detail/userstory/293332135224

https://web-dev-drupal-cdoherty.dev.a1.vary.redhat.com/en/modal-demo


mwcz:
@TylerMartin5 did some work in the RHElements days on a modal.  I think he was using the new-ish [`<dialog>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) element.  Just mentioning it in case there's something to be gained by looking at what Tyler started.

chrisdo1:
> @TylerMartin5 did some work in the RHElements days on a modal. I think he was using the new-ish [`<dialog>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) element. Just mentioning it in case there's something to be gained by looking at what Tyler started.

Thanks for the heads up, Michael! We reviewed the dialog element and decided against it because it's not very well supported at this time: https://caniuse.com/#feat=dialog

elements/pfe-modal/src/pfe-modal.html
```diff
@@ -0,0 +1,14 @@
+<slot name="pfe-modal--trigger"></slot>
+<section class="pfe-modal__overlay" hidden>
```
chrisdo1: @castastrophe @starryeyez024 I applied the class name `pfe-modal__overlay` which is technically correct since it contains the overlay styles.

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,125 @@
...
+		border-radius: #{pfe-var(ui--border-radius)};
+		overflow-y: auto;
+		overscroll-behavior: contain;
+
+		@media screen and (min-width: $pfe-modal--breakpoint--medium) {
+			max-width: pfe-local(MaxWidth);
+			max-height: pfe-local(MaxHeight);	
+		}
+	}
+	&__container {
```
starryeyez024: If you add one more div inside the container, then the container could have the border radius, and the inner container could have the scroll bars. This would allow the border radius to apply to the scrollbar side. Right now it has a hard edge on the right side:

image


```html
<div class="pfe-modal__container">
   <div class="pfe-modal__container-inner"
   ...
   </div>
</div>
```

```css

--pfe-modal--MaxHeight: 90vh;

.pfe-modal__wrapper {
    z-index: 1;
    min-width: var(--pfe-modal--MinWidth);
    min-height: var(--pfe-modal--MinHeight);
    margin: 0 auto;
    position: absolute;
    top: calc( (100vh - var(--pfe-modal--MaxHeight)) / 2);  // also suggested to set some top/bottom spacers based on the maxHeight var
    bottom: calc( (100vh - var(--pfe-modal--MaxHeight)) / 2); 
    max-width: var(--pfe-modal--MaxWidth);
    max-height: var(--pfe-modal--MaxHeight);
    border-radius: #{pfe-var(ui--border-radius)};
 }

.pfe-modal__container {
    padding: 0;
    max-height: 100%;
    position: relative;
    border: white 1px solid;
    border-radius: #{pfe-var(ui--border-radius)};
    box-sizing: border-box;
    overflow: hidden;
    height: 100vh;
 }

.pfe-modal__container-inner {
    overflow-y: scroll;
    overscroll-behavior: contain;
    position: relative;
    display: block;
    height: calc(100% - var(--pfe-modal--Padding));
    padding: var(--pfe-modal--Padding);
    padding-right: 62px !important;
    margin-bottom: var(--pfe--modal--Padding);
    box-sizing: border-box;
}
```
image

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,125 @@
+@import "../../pfe-sass/pfe-sass";
+$LOCAL: modal;
+
+$pfe-modal--breakpoint--medium: 640px;
+
+:host {
+	--pfe-modal--MaxWidth: 70vw;
+	--pfe-modal--MinWidth: 50vw;
+	--pfe-modal--MaxWidth--mobile: 94vw;
+	--pfe-modal--MaxHeight: none;
```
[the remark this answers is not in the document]
chrisdo1: @starryeyez024 good point, then we don't need to worry about adding a margin at the top.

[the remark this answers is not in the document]
chrisdo1: @starryeyez024 Good suggestion! Set `max-height: 90vh`

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,125 @@
...
+		z-index: 2;
+		position: relative;
+		max-width: pfe-local(MaxWidth--mobile);
+		min-width: pfe-local(MinWidth);
+		min-height: pfe-local(MinHeight);
+		margin: 0 auto;
+		box-shadow: pfe-var(box-shadow--lg);
+		background-color: pfe-color(surface--lightest);
+		border-radius: #{pfe-var(ui--border-radius)};
+		overflow-y: auto;
```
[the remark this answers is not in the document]
chrisdo1: @starryeyez024 thanks for reviewing. Applied this suggestion.

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: Perfect. Works great now.

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,121 @@
...
+	--pfe-modal--Padding: calc(#{pfe-var(container-padding)} * 2) calc(#{pfe-var(container-padding)} * 4) calc(#{pfe-var(container-padding)} * 2) calc(#{pfe-var(container-padding)} * 2);
+}
+
+.pfe-modal {
+	&__outer {
+		display: flex;
+		position: fixed;
+		height: 100%;
+		width: 100%;
+		top: 0;
```
starryeyez024:
```suggestion
		top: calc( (100vh - var(--pfe-modal--MaxHeight)) / 2);  // also suggested to set some top/bottom spacers based on the maxHeight var
		bottom: calc( (100vh - var(--pfe-modal--MaxHeight)) / 2); 
```

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,121 @@
...
+		border-radius: #{pfe-var(ui--border-radius)};
+
+		@media screen and (min-width: $pfe-modal--breakpoint--medium) {
+			max-width: pfe-local(MaxWidth);
+		}
+	}
+	&__container {
+		position: relative;
+		max-width: calc(100% - #{pfe-var(ui--element--size)});
+		padding: var(--pfe-modal--Padding);
```
starryeyez024:
```suggestion
    margin: 0;
    overflow: hidden;
    width: 100%;
```

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,121 @@
...
+			margin-top: 0 !important;
+			width: 90%;
+		}
+	}
+
+	&__content {
+		overflow-y: auto;
+		overscroll-behavior: contain;
+		max-height: pfe-local(MaxHeight);	
+		padding-right: calc(#{pfe-var(container-padding)} * 2);
```
starryeyez024:
```suggestion
		    padding: var(--pfe-modal--Padding);
		    padding-right: calc(var(--pfe-theme--container-padding,1rem) * 2);
}
```

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	&__close {
+		@include pfe-reset-button;
+		position: absolute;
+		top: calc(#{pfe-var(container-padding)} * .25);
```
starryeyez024:
```suggestion
    top: 0;
    right: var(--pfe-theme--container-padding,1rem);
    background-color: white;
    border-radius: #{pfe-var(ui--border-radius)};
```

elements/pfe-modal/src/pfe-modal.scss
castastrophe: This line won't work but you can do this which I believe works well: `top: calc(50vh - var(--pfe-modal--MaxHeight));` or this: `top: calc(100vh - calc(var(--pfe-modal--MaxHeight) / 2));` however nested calcs don't work in IE11 so you need a manual fallback for it if you do this.

castastrophe: ^ Reason being calc in IE11 runs into issues with nested calculations

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,142 @@
...
+
+$pfe-modal--breakpoint--medium: 640px;
+
+:host {
+	--pfe-modal--MaxWidth: 70vw;
+	--pfe-modal--MinWidth: 50vw;
+	--pfe-modal--MaxWidth--mobile: 94vw;
+	--pfe-modal--MaxHeight: 90vh;
+	--pfe-modal--Padding: calc(#{pfe-var(container-padding)} * 2) 0 calc(#{pfe-var(container-padding)} * 2) calc(#{pfe-var(container-padding)} * 2);
+	--pfe-modal--Padding--mobile: 0 0 0 calc(#{pfe-var(container-padding)} * 2);
```
castastrophe: Instead of creating a separate mobile variable, could we update the original padding variable inside a media query in the host selector?  So...
```
:host {
     --pfe-modal--Padding: calc(#{pfe-var(container-padding)} * 2) 0 calc(#{pfe-var(container-padding)} * 2) calc(#{pfe-var(container-padding)} * 2);
     @media screen and (max-height: $pfe-modal--breakpoint--medium) {
          --pfe-modal--Padding: 0 0 0 calc(#{pfe-var(container-padding)} * 2);
     }
}
```

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,142 @@
...
+	--pfe-modal__overlay--BackgroundColor: #{pfe-color(overlay)};
+	--pfe-modal--CloseButtonSize: calc(#{pfe-var(ui--element--size)} - 4px);
```
castastrophe: Can we rename this to: `--pfe-modal--CloseButtonSize` to `--pfe-modal__close--size`?  In BEM naming this would be `__{region}--{abstract name}`.  The only time BEM uses camel case is for CSS properties like `BackgroundColor` or `Padding`.

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,142 @@
...
+		display: flex;
+		position: fixed;
+		height: 100%;
+		width: 100%;
+		top: 0;
+		left: 0;
+		flex-direction: column;
+		align-items: center;
+		justify-content: center;
+		z-index: var(--pfe-z--modal, 2000);
```
castastrophe: Let's add the z-index map to this branch too.  This line would become `z-index: #{pfe-var($cssvar: modal, $map: $z-index)};` and in pfe-sass in the `_maps` file, can you add:
```
$z-index: (
  modal: 99,
  navigation: 98,
  overlay: 97,
  content: 0
);
```

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,142 @@
...
+
+		&[hidden] {
+			display: none;
+			z-index: 0;
```
castastrophe: Can we update this to: `z-index: #{pfe-var($cssvar: content, $map: $z-index)};`

elements/pfe-modal/src/pfe-modal.scss
castastrophe: I like the idea of all z-index's starting below 100 if we can and building our layers from there.  If needed, someone could customize their z-indexes in their theme attributes.

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,142 @@
...
+			padding: var(--pfe-modal--Padding--mobile);
+		}
+
+		&[hidden] {
+			display: none;
+		}
+
+		:host(:not([has_header])) & {
+			@media screen and (min-width: $pfe-modal--breakpoint--medium) {
+				padding: calc(#{pfe-var(container-padding)} * 2) calc(#{pfe-var(container-padding)} * 6) calc(#{pfe-var(container-padding)} * 2) calc(#{pfe-var(container-padding)} * 2);
```
castastrophe: Can these customizations be applied to the padding variable instead of directly to the padding attribute?

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,142 @@
...
+				padding: calc(#{pfe-var(container-padding)} * 2) calc(#{pfe-var(container-padding)} * 6) calc(#{pfe-var(container-padding)} * 2) calc(#{pfe-var(container-padding)} * 2);			
+			}
+
+			@media screen and (max-height: $pfe-modal--breakpoint--medium) and (min-width: $pfe-modal--breakpoint--medium) {
+				padding: 0 0 calc(#{pfe-var(container-padding)} * 2) calc(#{pfe-var(container-padding)} * 2);			
+			}
+
+			// Remove margin-top on the first slotted element that is not the header.
+			::slotted(*:nth-child(1)),
+			::slotted(*:nth-child(2)) {
```
castastrophe: Why first and second child?

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,142 @@
...
+				margin-top: 0 !important;
+			}
+		}
+		::slotted([slot="pfe-modal--header"]) {
+			margin-top: 0 !important;
+			width: 90%;
```
castastrophe: Should this be max-width?

castastrophe: Would it be more accurate to do: `width: calc(100% - #{pfe-local($cssvar: size, $region: close)} - #{pfe-var(content-spacer)});`?  Basically subtract the size of the close button from the width of the header plus a little content-spacer breathing room between the header and the button?

elements/pfe-modal/src/pfe-modal.scss
castastrophe: May I suggest:
```
padding: pfe-local(Padding);
padding-right: calc(#{pfe-var(container-padding)} * 2);
```

elements/pfe-modal/src/pfe-modal.scss
castastrophe: I'm not sure if the background color or border-radius make sense for the close button - is this comment in the right place of the code?

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,142 @@
+@import "../../pfe-sass/pfe-sass";
+$LOCAL: modal;
+
+$pfe-modal--breakpoint--medium: 640px;
+
+:host {
+	--pfe-modal--MaxWidth: 70vw;
```
castastrophe: Should we add a `--pfe-modal--BackgroundColor` variable to allow a hook for designers who want to change the color of the modal?

elements/pfe-modal/test/pfe-modal_test.html
```diff
@@ -0,0 +1,27 @@
...
+      This is the element content.
+    </pfe-modal>
+
+    <script>
+      suite('<pfe-modal>', () => {
+        test('it should upgrade', () => {
+          assert.instanceOf(document.querySelector('pfe-modal'), customElements.get("pfe-modal", 'pfe-modal should be an instance of pfeModal'));
+        });
+
+        // you need to write more tests!!!
```
castastrophe: @chrisdo1 Do you feel comfortable taking a stab at the tests or would you like me to?

examples/index.html
```diff
@@ -23,9 +23,11 @@ <h1>Demos</h1>
 					<li><a href="../elements/pfe-icon-panel/demo">pfe-icon-panel</a></li>
 					<li><a href="../elements/pfe-layouts/demo">pfe-layouts</a></li>
 					<li><a href="../elements/pfe-markdown/demo">pfe-markdown</a></li>
+					<li><a href="../elements/pfe-modal/demo">pfe-modal</a></li>
 					<li><a href="../elements/pfe-number/demo">pfe-number</a></li>
 					<li><a href="../elements/pfe-page-status/demo">pfe-page-status</a></li>
 					<li><a href="../elements/pfe-progress-indicator/demo">pfe-progress-indicator</a></li>
+					<li><a href="../elements/pfe-styles/demo">pfe-styles</a></li>
```
castastrophe: I think the pfe-styles item can be removed

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: I think it's a reference to the `pfe-modal__window` 
We refactored the shadowdom a bit since this comment I think so it may not actually be referencing the same element any more.

elements/pfe-modal/README.md
```diff
@@ -0,0 +1,66 @@
...
+  <pfe-cta>
+    <a href="#bar">Learn more</a>
+  </pfe-cta>
+</pfe-modal>
+```
+
+## Slots
+
+### Trigger
+The only part visible on page load, the trigger opens the modal window. The trigger can be a button, a cta or a link.
```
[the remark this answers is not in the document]
chrisdo1: Yes! Added that.

elements/pfe-modal/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Header
+The header is an optional slot that appears at the top of the modal window. It should be a header tag (h2-h6). 
+
+### Default slot
+The default slot can contain any type of content. When the header is not present this unnamed slot appear at the top of the modal window (to the left of the close button). Otherwise it will appear beneath the header.
+
+## Events
+
+### openModal
```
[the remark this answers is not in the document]
chrisdo1: Good point. Added that verbiage to both events.

elements/pfe-modal/README.md
```diff
@@ -0,0 +1,66 @@
...
+Fires when a user clicks on the trigger.
+
+### closeModal
+Fires when either a user clicks on either the close button or the overlay.
+
+
+Make sure you have [Polyserve][polyserve] and [Web Component Tester][web-component-tester] installed.
```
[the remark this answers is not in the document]
chrisdo1: Yes! Done.

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: Done. Much cleaner now. Thanks.

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: Good one. I altered the variable name slightly `-pfe-modal--close--size` so that it would work with the `pfe-local` function. Let me know if I should change it back.

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: Hmm. I think that makes this overly complicated since we'll get into the issue of text color as well if someone opts for a modal on a dark background. What do you think?

castastrophe: The text should be light DOM so it will pick up the theme stylesheet which will applying the broadcast--text variable which can be updated at the same time as the background color.  That does point out that we would need to make the fill color on the close icon pick up the broadcast color though.  Let's do this update as a future enhancement instead of now. 👍

elements/pfe-modal/src/pfe-modal.scss
castastrophe: the `__close` is the region which is what the underscore highlights.  You can do this in the pfe-local function like this: `pfe-local($cssvar: size, $region: close)`.

elements/pfe-modal/src/pfe-modal.scss
starryeyez024: ++

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: Sounds good!

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: Ahh, thanks. Done.

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: Yes! I was going to ask about using the z-index map from the nav. I wonder if we should set it higher than 99 since the sticky nav on rh.com is at 1000.

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: Yes, done.

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: Yep. Done. 
Moved all the padding declarations from the `__container` to the `__content` section and it's working now.

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: Good point. Removed the first-child.

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: done.

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: @castastrophe maybe it makes more sense to remove the z-index map and add it back later (when the nav is actually at z-index: 98)?

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: @castastrophe Rather than adjust the header width I set the `pfe-modal__content` region to get narrower so that the close button doesn't overlap. Testing this on the cdoherty sandbox but it seems to be working correctly.

elements/pfe-modal/test/pfe-modal_test.html
chrisdo1: I'll give it a shot.

elements/pfe-modal/src/pfe-modal.scss
castastrophe: Added this issue for tracking that work later: https://github.com/patternfly/patternfly-elements/issues/483

elements/pfe-modal/package.json
```diff
@@ -0,0 +1,43 @@
...
+  },
+  "contributors": [
+    {
+      "name": "castastrophe",
+      "url": "https://github.com/castastrophe"
+    }
+  ],
+  "license": "MIT",
+  "dependencies": {
+    "@patternfly/pfe-sass": "1.0.0-prerelease.17",
```
castastrophe: These will need to be bumped to prerelease.19

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: I have moved this and `::slotted([slot="pfe-modal--header"])` so that it's under `&__content`
The header styles weren't working in Edge. I also readded the `::slotted(*:nth-child(1))`. For some reason Edge would remove the margin on top without that.

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: I have removed this and left a comment that we should revisit this once the nav is merged in and in production.

castastrophe: We can set it higher in our dotCom theme but I think for the general PFE library, starting at 100 should be plenty of levels.

castastrophe: Yours will be merged in first and since you need the z-index for it to work, go ahead and copy in the z-index map to your project.

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: @starryeyez024 @castastrophe I think that the original intent of this code was to vertically center the modal with some margins above and below. We achieved this with flexbox and set a max-height on the window itself. Marking this resolved if it's okay with everyone?

examples/index.html
chrisdo1: ahh, pesky index file. Removed.

elements/pfe-modal/package.json
chrisdo1: Thanks! Made that fix.

examples/index.html
chrisdo1: Removed. Thanks.

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: @castastrophe added this back.

elements/pfe-modal/test/pfe-modal_test.html
chrisdo1: @castastrophe I added a few tests. Let me know if there are additional ones you'd like to see.

starryeyez024:
@kylebuch8 did you back out and fork your changes?

kylebuch8:
@starryeyez024 We worked though things in a code review and didn't need to back things out so we're good.

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,132 @@
...
+		&[hidden] {
+			display: none;
+		}
+	}
+
+	&__content {
+		overflow-y: auto;
+		overscroll-behavior: contain;
+		max-height: pfe-local(MaxHeight);
+		padding: pfe-local(Padding);
```
starryeyez024:
```suggestion
		padding: pfe-local(Padding);
		box-sizing: border-box; /*added*/
```

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,132 @@
...
+
+		@media screen and (max-height: $pfe-modal--breakpoint--medium) {
+			box-sizing: border-box;
+			--pfe-modal--Padding: #{pfe-var(container-padding)} calc(#{pfe-var(container-padding)} * 2) #{pfe-var(container-padding)} #{pfe-var(container-padding)};
```
starryeyez024:
```suggestion
			--pfe-modal--Padding: #{pfe-var(container-padding)} calc(#{pfe-var(container-padding)} * 3) #{pfe-var(container-padding)} #{pfe-var(container-padding)};
```

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	&__content {
+		overflow-y: auto;
+		overscroll-behavior: contain;
+		max-height: pfe-local(MaxHeight);
+		padding: pfe-local(Padding);
+
+		@media screen and (max-height: $pfe-modal--breakpoint--medium) {
+			box-sizing: border-box;
```
starryeyez024:
```suggestion
			/* box-sizing: border-box;   added above */
```

starryeyez024: https://www.evernote.com/l/AA4sT3IF7rZHaq9pOwxX6O6edit3-zS7NVs

elements/pfe-modal/src/pfe-modal.scss
starryeyez024: https://www.evernote.com/l/AA4mpPNI_etKwrCiKs8mNyHj6Kg5HmxkCns

elements/pfe-modal/src/pfe-modal.scss
```diff
@@ -0,0 +1,132 @@
...
+
+		&[hidden] {
+			display: none;
+		}
+	}
+	&__window {
+		position: relative;
+		max-width: pfe-local(MaxWidth--mobile);
+		min-width: pfe-local(MinWidth);
+		max-height: calc(100% + -32px);
```
starryeyez024: why this calculation instead of `var(--pfe-modal--MaxHeight)`?

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: @starryeyez024 added this (and removed the other two instances of it).

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: Yes! Added this. Was able to remove the other media query since the padding is now identical.

elements/pfe-modal/src/pfe-modal.scss
chrisdo1: Ahh, good point. Fixed this.

elements/pfe-modal/src/pfe-modal.scss
starryeyez024: +1

PR #460, opened 2019-07-03
pfe-cta| broadcast var updates, arrow positioning

starryeyez024:
- removes function from band & card which was putting on=dark attributes on CTAs
- uses local vars inside of CTA as the source of truth, leaving room for broadcast vars to show up and have an opinion // [added these notes to markdown file](https://docs.google.com/document/d/1P6ohcr13y0DTca_aow6Lm5qSnUcIdM56wu8Fc4SwE00/edit#heading=h.3w8wgkgnq9ov)
- removes alt attribute, we are just using color=complement
- adds variety of tests to the demo file
- adjusts arrow so that positioning is inline, and it does not nudge content to the right on over

Demo pages:
- https://web-dev-drupal-webux3.dev.a1.vary.redhat.com/es/technologies/all-products
- https://web-dev-drupal-webux3.dev.a1.vary.redhat.com/en/web-components-demo

IE11 is throwing errors related to band / card color function removal

### DIFF
--- a/elements/pfe-cta/src/pfe-cta.scss
+++ b/elements/pfe-cta/src/pfe-cta.scss
@@ -178,10 +178,12 @@ $pfe-cta__arrow-size: 0.313em;
 /// ===========================================================================
 // %^&$#@%^&*$#@&$*#@^&$(#@^$&*#@^$&*@#^$&*(@#^$&*#@^$&*#@^$&*#@$#))
 
-/* Broadcast interception! Tune into whatever colors the containers are offering up, per priority level */
-// primary sets its own background colors, so it ignores broadcast 
-// if there is no broadcast, these are *still* seen by the browser, so set fallbacks here!
-// fallback values should match whats above 
+// Broadcast interception! Tune into whatever colors the containers are offering up, per priority level 
+
+// - The location of these styles in the doc is important
+// - priority="primary" sets its own background colors, so it ignores broadcast 
+// - if there is no broadcast variable available, these are still seen by the browser, so we also set fallbacks here!
+// - Fallback values should match whats above
 
 //default
 :host(:not([priority])) {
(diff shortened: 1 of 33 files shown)


starryeyez024:
@castastrophe previously, the `pfe-cta` extended the full width of the container and the link inside it was inline-block so it was only as wide as the text:
image

https://redhat.com/en/web-components-demo

Now that we are utilizing the :host for background colors, the `pfe-cta` extends the full width because of the display:grid on the `.pfe-band__footer`. In order to fix this, I put the CTA in a group. Is this an acceptable/desirable workaround?
image

https://web-dev-drupal-webux3.dev.a1.vary.redhat.com/en/web-components-demo

elements/pfe-cta/src/pfe-cta.scss
```diff
@@ -109,16 +102,16 @@ $pfe-cta__arrow-size: 0.313em;
...
   text-decoration: none;
 }
 
 ::slotted(a:focus) {
   outline: none !important;
-  color: #{pfe-local(Color--focus)};
+  color: #{pfe-local(Color--focus)} !important;
 }
 
 :host(:focus-within) {
```
starryeyez024: @castastrophe @chrisdo1  the polyfill for IE11 trips over the `focus-within` selector. If its present, the whole component does not upgrade. Kyle suggested we remove it and instead put the arrow in the light dom, POC 
```
class PfeCTa extends HTMLElement {
	constructor() {
  	this._hasArrow = false;	
  }
  
	connectedCallback() {
  	super.connectedCallback()
    
    const arrow = `<svg whatever>`;
    const fragment = document.createDocumentFragment();
    fragment.innerHTML = arrow;
    
    if (!this._hasArrow) {
    	this.appendChild(fragment);
      this._hasArrow = true
    }
  }
}
```

PR #462, opened 2019-07-12
Pfe progress indicator

Djfaucette:
## Feature Branch: Pfe Progress Indicator

* Appendix doc:
    - 
* Link(s) to demo pages_ where this element can be viewed:
    - 

### Ready-for-merge Checklist

- [x] Expected files: all files in the merge request related to one feature request?
- [x] Is your branch up-to-date with the master branch?
- [x] Did automated tests pass?
- [x] Was it [browser-tested] Not sure how to test in web store? (https://docs.google.com/spreadsheets/d/1luriPYswlGcKdcbKhQfW6kxIgj7tjNjnBVP7yZ4onwA/edit#gid=1660273076)?
- [ ] Was this feature demo'd and the design review task marked as complete?
- [ ] Did it get a LGTM *after the last commit*? (changelog updates are exempt from this rule)
- [ ] Changelog: consider this your postage stamp, it can’t go out without one! Stories should be in alphabetical/numerical order under the tag release.

---

### What's in this branch?


elements/pfe-progress-indicator/README.md
```diff
@@ -0,0 +1,32 @@
+# PatternFly Element | Progress Element
+
```
chrisdo1: I think we should add `## Overview` and `## Usage` documentation here.

elements/pfe-progress-indicator/src/pfe-progress-indicator.js
```diff
@@ -0,0 +1,54 @@
...
+
+  get templateUrl() {
+    return "pfe-progress-indicator.html";
+  }
+
+  get styleUrl() {
+    return "pfe-progress-indicator.scss";
+  }
+
+  // static get observedAttributes() {
```
chrisdo1: Can we remove commented code?

elements/pfe-progress-indicator/src/pfe-progress-indicator.js
```diff
@@ -0,0 +1,54 @@
...
+        `${
+        PfeProgressIndicator.tag
+        }: You do not have a backup loading message.`
+      );
+    } else {
+      // do nothing
+    }
+  }
+
+  // attributeChangedCallback(attr, oldValue, newValue) {}
```
chrisdo1: Remove?

elements/pfe-progress-indicator/src/pfe-progress-indicator.scss
```diff
@@ -0,0 +1,65 @@
...
+:host([hidden]) {
+  display: none;
+}
+
+:host([pfe-indeterminate]) {
+    animation: spin 1s linear infinite;
+    -webkit-animation:spin 1s linear infinite;
+    -moz-animation:spin 1s linear infinite;
+
+    border-bottom: 4px solid rgba(0,0,0,0.25);
```
chrisdo1: Can we set local variables for these so they aren't repeated? Maybe something like:
```
$pfe-progress-indicator--border-light: 4px solid rgba(0,0,0,0.25);
$pfe-progress-indicator--border-dark: 4px solid rgba(0,0,0,0.75);
```

elements/pfe-progress-indicator/src/pfe-progress-indicator.scss
```diff
@@ -0,0 +1,65 @@
...
+    display: inline-block;
+    height: 2rem;
+    margin: 0 auto;
+    position: relative;
+    width: 2rem;
+    vertical-align: middle;
+    visibility: visible;
+}
+
+@-ms-keyframes spin {
```
chrisdo1: Since a lot of these values are repeated I wonder if we could set some local variables for these or possibly include them in a mixin with an @for to loop through the different vendor prefixes?

Djfaucette: I think once we have auto-prefixer that will be a moot point. At the moment it's four rules, but I think in the future it will just be one.

elements/pfe-progress-indicator/README.md
```diff
@@ -0,0 +1,44 @@
+# PFElements Progress Indicator Element
+
+`pfe-progress-indicator` is a call-to-action (CTA) element, that stands out from regular hypertext links, and is used for linking users to webpages.
```
castastrophe: This description seems potentially out of date?  Is it considered a call-to-action?

elements/pfe-progress-indicator/README.md
```diff
@@ -0,0 +1,44 @@
...
+## Usage
+```html
+<pfe-progress-indicator pfe-indeterminate>
+    <h1>My fallback loading message</h1>
+</pfe-progress-indicator>
+```
+
+## Slots
+
+Progress Indicator has one slot only. It is named `progress-message`.
```
castastrophe: This might be out of date too - I believe this there is only one unnamed slot, correct?

elements/pfe-progress-indicator/README.md
```diff
@@ -0,0 +1,44 @@
+# PFElements Progress Indicator Element
+
+`pfe-progress-indicator` is a call-to-action (CTA) element, that stands out from regular hypertext links, and is used for linking users to webpages.
+
+## Usage
+```html
+<pfe-progress-indicator pfe-indeterminate>
```
castastrophe: Can you add notes to the doc about what pfe-indeterminate means?

elements/pfe-progress-indicator/src/pfe-progress-indicator.html
```diff
@@ -0,0 +1 @@
+<slot class="sr-only" name="progress-message"></slot>
```
castastrophe: Should this be an unnamed slot?

elements/pfe-progress-indicator/src/pfe-progress-indicator.json
```diff
@@ -0,0 +1,51 @@
...
+  "properties": {
+    "slots": {
+      "title": "Slots",
+      "description": "Definition of the supported slots",
+      "type": "object",
+      "properties": {
+        "content": {
+          "title": "Content",
+          "type": "array",
+          "namedSlot": true,
```
castastrophe: This should be set to false and the oneOf should be "raw" instead of "progress-message".  The array inside content indicates what components or code is allowed inside the slot.  Since all HTML is allowed inside the slot, we can just use raw.

elements/pfe-progress-indicator/demo/index.html
```diff
@@ -0,0 +1,28 @@
...
+  <!-- uncomment the es5-adapter if you're using the umd version -->
+  <script src="/@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js"></script>
+  <script src="/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.5/require.min.js"></script>
+  <script>require(['../pfe-progress-indicator.umd.js'])</script>
+</head>
+
+<body unresolved>
+  <h1>&lt;pfe-progress-indicator&gt;</h1>
+  <pfe-progress-indicator pfe-indeterminate>
```
castastrophe: Instead of requiring the light-dom to provide the initial state, can the web component set it?

castastrophe: What about using aria attributes instead of a custom attribute? https://stackoverflow.com/questions/31716425/how-do-you-make-the-loading-icon-accessible-for-screen-readers-like-jaws

elements/pfe-progress-indicator/src/pfe-progress-indicator.scss
```diff
@@ -0,0 +1,68 @@
...
+$pfe-progress-indicator--border-light: 4px solid rgba(0, 0, 0, 0.25);
+$pfe-progress-indicator--border-dark: 4px solid rgba(0, 0, 0, 0.75);
+
+:host([hidden]) {
+  display: none;
+}
+
+:host([pfe-indeterminate]) {
+    animation: spin 1s linear infinite;
+    -webkit-animation:spin 1s linear infinite;
```
castastrophe: Don't worry about manually coding vendor prefixes - we use autoprefixer and it will add that for you.  You can confirm in the rendered .css file that this is being done correctly.

elements/pfe-progress-indicator/src/pfe-progress-indicator.scss
```diff
@@ -0,0 +1,68 @@
+@import "../../pfe-sass/pfe-sass";
+
+$pfe-progress-indicator--border-light: 4px solid rgba(0, 0, 0, 0.25);
```
castastrophe: Instead of hardcoding these values, can you use the variables?  

`$pfe-progress-indicator--border-light: #{pfe-var(surface--border-width--heavy)} #{pfe-var(surface--border-style)} rgba(0, 0, 0, .25);
$pfe-progress-indicator--border-dark: #{pfe-var(surface--border-width--heavy)} #{pfe-var(surface--border-style)} rgba(0, 0, 0, .75);`

elements/pfe-progress-indicator/src/pfe-progress-indicator.scss
```diff
@@ -0,0 +1,68 @@
...
+    -moz-animation:spin 1s linear infinite;
+
+    border-bottom: $pfe-progress-indicator--border-light;
+    border-left: $pfe-progress-indicator--border-light;
+    border-right: $pfe-progress-indicator--border-light;
+    border-top: $pfe-progress-indicator--border-dark;
+    border-radius: 100%;
+    
+    display: inline-block;
+    height: 2rem;
```
castastrophe: Could we set this as a variable to allow customization if necessary?
```
height: var(--pfe-progress-indicator--Height, 2rem);
```

elements/pfe-progress-indicator/src/pfe-progress-indicator.scss
```diff
@@ -0,0 +1,68 @@
...
+    margin: 0 auto;
+    position: relative;
+    width: 2rem;
```
castastrophe: Same here or use the same variable and name it something more generic like `--pfe-progress-indicator--size`
```
width: var(--pfe-progress-indicator--Width, 2rem);
```

elements/pfe-progress-indicator/src/pfe-progress-indicator.scss
castastrophe: Autoprefixer is already added so you can remove this code :)

elements/pfe-progress-indicator/src/pfe-progress-indicator.scss
```diff
@@ -0,0 +1,68 @@
...
+}
+@keyframes spin {
+    from {
+        transform: rotate(0deg);
+    } to {
+        transform: rotate(360deg);
+    }
+}
+
+  .sr-only {
```
castastrophe: There's already an sr-only mixin, you can replace this with:
```
.sr-only {
   @include pfe-sr-only;
}
```

elements/pfe-progress-indicator/src/pfe-progress-indicator.story.js
```diff
@@ -0,0 +1,11 @@
+import { storiesOf } from "@storybook/polymer";
+import "../pfe-progress-indicator";
+
+storiesOf("Progress", module).add(
```
castastrophe: This file needs to be moved to the demo folder for consistency with other components.  Also, for the implementation, try this instead (it's more dynamic):
```
import { storiesOf } from "@storybook/polymer";
import * as storybookBridge from "@storybook/addon-knobs/polymer";
import * as tools from "../../../.storybook/utils.js";

import PfeProgressIndicator from "../pfe-progress-indicator";

const stories = storiesOf("Progress indicator", module);

// Define the template to be used
const template = (data = {}) => {
  return tools.component(PfeProgressIndicator.tag, data.prop, data.slots);
};

stories.addDecorator(storybookBridge.withKnobs);

stories.add(PfeProgressIndicator.tag, () => {

  let config = {};
  const props = PfeProgressIndicator.properties;

  // Trigger the auto generation of the knobs for attributes
  config.prop = tools.autoPropKnobs(props, storybookBridge);

  const slots = PfeProgressIndicator.slots;

  // Trigger the auto generation of the knobs for slots
  config.has = tools.autoContentKnobs(slots, storybookBridge);

  const rendered = template(config);
  return tools.preview(rendered);
});
```

elements/pfe-progress-indicator/src/pfe-progress-indicator.scss
Djfaucette: Awesome, incorporated.

elements/pfe-progress-indicator/src/pfe-progress-indicator.html
Djfaucette: Yes, updated.

elements/pfe-progress-indicator/demo/index.html
Djfaucette: I actually looked into aria-busy, but when I looked at what it does it didn't seem like the right choice. The official write up about it says something "Users of assistive technology will not be able to read anything while the content is getting updated." This is a message that a screen reader could actually see and read.

elements/pfe-progress-indicator/src/pfe-progress-indicator.scss
Djfaucette: Def!

elements/pfe-progress-indicator/demo/index.html
Djfaucette: @castastrophe Let me know if we want to toss this idea around some more. For now I'm marking resolved.

elements/pfe-progress-indicator/src/pfe-progress-indicator.js
```diff
@@ -0,0 +1,53 @@
...
+    const firstChild = this.children[0];
+
+    if (!firstChild) {
+      console.warn(
+        `${
+        PfeProgressIndicator.tag
+        }: You do not have a backup loading message.`
+      );
+    } else {
+      // do nothing
```
castastrophe: If no action, you can remove the else statement here

elements/pfe-progress-indicator/README.md
```diff
@@ -9,6 +9,8 @@
 </pfe-progress-indicator>
 ```
 
+At the time of writing there is only one style variant, `pfe-indeterminate` that spins without informing the user of where there are in the waiting process, only that http activity has not been resolved. This may change in the future as more style variants become available.
```
castastrophe: Minor typo: where **they** are in the waiting process

elements/pfe-progress-indicator/README.md
castastrophe: This still indicates that the component has 1 named slot.  Is that accurate?

elements/pfe-progress-indicator/src/pfe-progress-indicator.scss
```diff
@@ -19,50 +22,18 @@ $pfe-progress-indicator--border-dark: 4px solid rgba(0, 0, 0, 0.75);
     border-radius: 100%;
     
     display: inline-block;
-    height: 2rem;
+    height: var(--pfe-progress-indicator--height, 2rem);
```
castastrophe: For BEM approach this should use uppercase for the name of css properties so: `--pfe-progress-indicator--Height` and `--pfe-progress-indicator--Width`

PR #502, opened 2019-08-19
pfe-collapse

kylebuch8:


markcaron:
@kylebuch8,

I tested the demo with VoiceOver in Chrome; NVDA in Firefox and Edge; and JAWS in Firefox and Edge. Everything works as expected.

Even though we expected the "Standalone Panel" with the simple "Toggle panel" button to not work well on screen readers (given the lack of informative ARIA attributes, like aria-expanded and aria-controls), it worked as expected: toggles fine, but provides nothing to screen readers. I wanted to note this, so that moving forward, _we encourage others_ to properly markup such "custom buttons".

PR #525, opened 2019-08-29
add emitEvent and .events

mwcz:
### What has changed and why

This PR adds two things we discussed in the meeting today.  An `events` object which stores event names, and an `emitEvent` function on the base class which is a tidy wrapper around the event-firing boilerplate.

#### `events` object

This object should be declared on any component that fires custom events.  It looks like this:

```js
static get events() {
  return {
    open: `${this.tag}:open`,
    close: `${this.tag}:close`
  };
}
```

This provides a single spot to store event names, and allows referencing the event names from outside the component.

Special note about components that contain several classes inside one file: it's not always clear whether the sub-components should prefix events with their own tag, or not.  The events object is compatible with both approaches, because the object can be referenced across component lines.  Use your best judgement.  For existing elements I kept the same prefixes as their preexisting events to avoid breaking changes.

#### `emitEvent` function

This adds a base class function called `emitEvent`.  It accepts the same arguments as the [CustomEvent](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent) constructor (ie, bubbles, cancelable, composed, and detail).  The only deviation from CustomEvent arguments is that bubbles defaults to `true` in PFE, whereas it defaults to `false` in CustomEvent.

Here's an example from accordion.

```js
this.emitEvent(PfeAccordion.events.change, {
  detail: {
      expanded: !this.expanded
  }
});
```

No docs added yet.  I want to collect feedback before committing time to docs.

### Testing instructions

`npm test`

### Ready-for-merge Checklist

- [ ] Did you update or add any necessary documentation (README.md, WHY.md, etc.)?


elements/pfelement/src/pfelement.js
```diff
@@ -277,6 +279,20 @@ class PFElement extends HTMLElement {
   log(...msgs) {
     PFElement.log(`[${this.tag}]`, ...msgs);
   }
+
+  emitEvent(
+    name,
+    { bubbles = true, cancelable = false, composed = false, detail = {} } = {}
```
castastrophe:
```suggestion
    { bubbles = true, cancelable = false, composed = false, detail = {} } = {},
    tag = this.tag
```

Can we set the tag as a default and allow overrides optionally?

elements/pfelement/src/pfelement.js
```diff
@@ -277,6 +279,20 @@ class PFElement extends HTMLElement {
...
+  ) {
+    this.dispatchEvent(
+      new CustomEvent(name, {
```
castastrophe:
```suggestion
      new CustomEvent(`${tag}:${name}`, {
```

elements/pfe-accordion/src/pfe-accordion.js
```diff
@@ -96,7 +102,7 @@ class PfeAccordion extends PFElement {
     this.setAttribute("role", "presentation");
     this.setAttribute("defined", "");
 
-    this.addEventListener(`${PfeAccordion.tag}:change`, this._changeHandler);
+    this.addEventListener(PfeAccordion.events.change, this._changeHandler);
```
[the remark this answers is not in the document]
mwcz: Yeah, static properties apply to the class, not the instance of the class.

Rule of thumb:

 - inside a static function, `this` is the class (PfeAccordion)
 - inside a non-static function, `this` is the instance of the class (a specific `<pfe-accordion>`)

elements/pfe-modal/demo/index.html
```diff
@@ -18,6 +19,10 @@
     <script>require(['../pfe-modal.umd.js'])</script>
   </head>
   <body unresolved>
+    <script charset="utf-8">
+      document.body.addEventListener("pfe-modal:open", () => console.log('modal opened') )
+      document.body.addEventListener("pfe-modal:close", () => console.log('modal closeed') )
```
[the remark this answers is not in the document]
mwcz: Whoooops, this whole script wasn't supposed to be committed. :fearful:

elements/pfelement/src/pfelement.js
mwcz: It's definitely possible but I this the current approach is more straightforward.  One argument for the current approach is that the `events` object contains the names of the events.  With the suggested approach, `events` would no longer be as useful because the true event names wouldn't be created until `emitEvent` is called.  Outside code would be forced to hardcode event names.  Not the end of the world, but I would like to offer the option to `import PfeAccordion` and then `addEventListener(PfeAccordion.events.open, ...)` in order to tap into PfeAccordion's event names.

An argument against the current approach is that it leaves event naming conventions up to the developer to implement and diligent reviewers to ensure.  I'm usually the one arguing for automation over diligence! :stuck_out_tongue:  But in this case, I really like having the `events` object for quick lookup of event names.

castastrophe: 👍

starryeyez024:
@mwcz can you catch this branch up with master? 
@kylebuch8 want to merge?

mwcz:
Born ready :)

castastrophe:
@mwcz I think this needs some documentation around how to set up events in the components going forward.

castastrophe:
@mwcz I took a stab at the readme but it probably needs some notes added to the docs section of the site.

PR #544, opened 2019-09-16
remove 'pfe-' from the names of pfe-icon's internal modules

mwcz:
So they aren't picked up by webrh's "pfe-*umd.js" matcher.


castastrophe:
👍 Looks good!  This will be helpful for others using similar build processes too.  Let's brainstorm soon on a more systematic fix as you suggested.

PR #556, opened 2019-09-27
Moved crucial how-to documentation

starryeyez024:
### Testing instructions

```
$ cd docs/
$ hugo server -D
```
then preview at http://localhost:1313/patternfly-elements/


docs/content/getting-started/_index.md
```diff
@@ -9,49 +9,82 @@ menu = "start"
...
+	1.  All elements are based off of pfelement.js so including the polyfills with this one file would mean you only need to include the pfelement.js file before you include anything else.
+4. Bundle all of the scripts together into one rollup, and include that.
+
 
 
 
 ## 3. Use PatternFly Elements markup
 
-You can use PatternFly Elements with other standard HTML markup in your app or page:
+Different components have different intended uses. We tend to thing of them in 3 distinct groups:
```
alazzara: typo thing should be think

docs/content/getting-started/_index.md
```diff
@@ -92,17 +125,155 @@ class App extends Component {
...
-  <pfe-card color="darkest">
-    <h2 slot="header">Dark card</h2>
-    <p>Lorem ipsum.</p>
-    <pfe-cta color="complement" priority="primary" on="dark" slot="footer">
-        <a href="#">Important call-to-action</a>
-    </pfe-cta>
-  </pfe-card>
+#### General
+
+  Priority & variant attributes change style styles globally:
```
alazzara: I think this should read:
Priority & variant attributes change styles globally:

docs/content/getting-started/_index.md
```diff
@@ -92,17 +125,155 @@ class App extends Component {
...
+
+  **Full list:**
+
+  *   `pfe-priority`
+  *   `pfe-variant`
+
+
+#### Specific
+
+Color or align attributes change specific properties
```
alazzara: I think this should read:
Color & align attributes change specific properties

docs/content/getting-started/_index.md
```diff
@@ -92,17 +125,155 @@ class App extends Component {
...
+
+<hr/>
+
+
+## FAQ
+
+### Should I use on=dark or color=darkest on my container? What's the difference?
+
+*   `on=dark` is being deprecated. Instead, custom classes already on the page should set broadcast values.
+*   The original goal for `on=dark` was all about context. But instead of having to provide another attribute somewhere, the card or a band or another container will inform any nested components that the text color needs to change through the set of broadcast variables. Existing broadcast vars (defined in `$BROADCAST-VARS`)  include: `text`, `ui-link`, `ui-link--hover`, `ui-link--visited`, `ui-link--focus`.
```
alazzara: Think this should read:
The original goal for `on=dark` was all about context, but instead of having to provide another attribute somewhere, the card, a band, or another container will inform any nested components that the text color needs to change through the set of broadcast variables.

docs/content/getting-started/_index.md
```diff
@@ -92,17 +125,155 @@ class App extends Component {
...
+  --pfe-broadcasted--color--ui-link:          var(--pfe-theme--color--ui-link--on-dark, #73bcf7);
+  --pfe-broadcasted--color--ui-link--hover:   var(--pfe-theme--color--ui-link--on-dark--hover, pink);
+  --pfe-broadcasted--color--ui-link--visited: var(--pfe-theme--color--ui-link--on-dark--visited, pink);
+  --pfe-broadcasted--color--ui-link--focus:   var(--pfe-theme--color--ui-link--on-dark--focus, pink);
+}
+```
+
+In themes (like the advanced-theme.css file from the Red Hat Theme ) we apply broadcast variables to plain links, because they are light DOM and also have default colors applied by the browser. This CSS file not only includes variables but also styles for headlines and links on the page. It sets the colors for these elements using a CSS variable, which web components can change the value of.
+
+For example, advanced-theme.css includes
```
alazzara: I think there should be a colon at the end of this sentence
For example, advanced-theme.css includes:

docs/content/getting-started/_index.md
```diff
@@ -92,17 +125,155 @@ class App extends Component {
...
+
+Instead, in the host of components, use:
+
+```
+:host {
+  color: var(--pfe-broadcasted--color--text);
+}
+```
+
+Then call theme mixin to flip colors of the on=dark on=light attributes. Default tags will use these colors. If devs implementing the component have more specific styles on their page, they will have to handle it.
```
alazzara: Seeing how it is stated earlier that `on=dark` is being deprecated should this be updated to reflect that or do you want to wait till `on=dark` is officially not supported?

docs/content/getting-started/_index.md
```diff
@@ -9,49 +9,82 @@ menu = "start"
...
+```
+
+
+
+### Container components (see also container notes below)
+
+```
+<body>
+    <pfe-card pfe-color="darkest">
+        Hello world.
```
castastrophe: Maybe wrap this in a p-tag since unwrapped text is problematic in components and it might not be good to show many demos doing that?

docs/content/getting-started/_index.md
```diff
@@ -92,17 +125,155 @@ class App extends Component {
...
 
+#### **Full list:**
+
+*   `pfe-align`
+*   `pfe-color`
+*   `vertical`
+*   `horizontal`
+
+
+## 5. Use CSS variables as needed
```
castastrophe: Suggest: Use CSS variables to provide hooks for customizing the component or as a means of updating styles based on attributes

It's a long title I know, maybe there's a more succinct way to communicate those two things.

castastrophe:
```suggestion
## 5. Use CSS variables to provide hooks for customization or as a means of updating internal styles
```

docs/content/getting-started/_index.md
starryeyez024: removed, thanks!

docs/content/getting-started/_index.md
starryeyez024: updated, thx!

docs/content/getting-started/_index.md
starryeyez024: updated, thx!

docs/content/getting-started/_index.md
starryeyez024: updated, thx!

docs/content/getting-started/_index.md
starryeyez024: updated, thx!

docs/content/getting-started/_index.md
```diff
@@ -92,17 +125,155 @@ class App extends Component {
...
+
+CSS variables are subject to the normal cascade, so consider where you want these overrides to propogate.
```
castastrophe:
```suggestion
CSS variables are subject to the normal cascade, so consider where you want these overrides to propagate.  We strongly recommend assigning internal variables or "local" variables to a property only once in a component and then utilizing the cascade to update the value of the variable.  For example:

```scss
:host {
     padding: var(--pfe-card--Padding, 20px);
}

:host([pfe-size="small"]) {
    --pfe-card--Padding: 10px;
}
 ```

castastrophe: Sorry the suggestion tool was getting confused by my insertion of the code block so that formatting will need to be cleaned up.

starryeyez024: This code example seems like a good addition to the theming section (please add!) but this getting-started doc is just for folks who are using the components, so they wouldn't be adding host CSS

docs/content/getting-started/_index.md
starryeyez024: thx, updated

docs/content/getting-started/_index.md
starryeyez024: thanks, updated

docs/content/getting-started/_index.md
castastrophe: +1 Will make a note to add that to the theming section instead!

PR #557, opened 2019-10-01
Documentation: support table

starryeyez024:
### What has changed and why
Moved status information about components to docs site and out of readme.

### Testing instructions

1. `cd docs`
2. `hugo server -d`
3. visit http://localhost:1313

image


starryeyez024:
Updated to public version of google doc

PR #572, opened 2019-10-17
adding a schema to pfe-icon-panel

kylebuch8:


elements/pfe-icon-panel/src/pfe-icon-panel.json
```diff
@@ -0,0 +1,99 @@
...
+        }
+      }
+    },
+    "attributes": {
+      "title": "Attributes",
+      "type": "object",
+      "properties": {
+        "icon": {
+          "title": "Icon",
+          "type": "string"
```
castastrophe: Can you add `"prefixed": false` to these attributes so it's clear they're unprefixed?

elements/pfe-icon-panel/src/pfe-icon-panel.json
```diff
@@ -0,0 +1,99 @@
...
+          "enum": [
+            "lightest",
+            "lighter",
+            "base",
+            "darker",
+            "darkest",
+            "complement",
+            "accent"
+          ],
+          "default": "base"
```
castastrophe: @starryeyez024 Do you think the default background color for a circled icon should be gray or would it make sense in this case to make it complement or accent?

elements/pfe-icon-panel/src/pfe-icon-panel.json
```diff
@@ -0,0 +1,99 @@
...
+        },
+        "centered": {
+          "title": "Centered",
+          "type": "Boolean",
```
castastrophe: Types in schema are weirdly lowercase 🤷‍♀

elements/pfe-icon-panel/src/pfe-icon-panel.json
```diff
@@ -0,0 +1,99 @@
...
+        },
+        "stacked": {
+          "title": "Stacked",
+          "type": "Boolean",
+          "default": false
+        },
+        "circled": {
+          "title": "Circled",
+          "type": "Boolean",
+          "default": false
```
castastrophe: @starryeyez024 Do you think this should this be true too for the default?

starryeyez024: Hmm, good question! I'm thinking that usually we opt into flair, so maybe the circle should be an add-on.

elements/pfe-icon-panel/src/pfe-icon-panel.json
starryeyez024: I'll echo the same idea, that bright colors are flair, so let's keep it basic (gray) by default.

elements/pfe-icon-panel/src/pfe-icon-panel.json
castastrophe: 👍

PR #596, opened 2019-10-30
feat: set PFE default accent color to red

starryeyez024:
I opened this to socialize the idea that we've talked about before which is bringing in official colors into the rep, but also because IE11 doesn't inherit CSS vars so our buttons are orange on prod in IE11.


castastrophe:
👍 I like the idea and I believe we've already discussed making the Design System standards the defaults in the project.

castastrophe:
Note, this will needs updates after broadcast is merged

PR #613, opened 2019-11-07
build: ignore compiled assets when determining version bumps

mwcz:
## ignore compiled assets when determining version bumps

Update lerna's `ignoreChanges` patterns so that committing compiled assets doesn't trigger version bumps in every element, during every release.  After this change, we should go back to the happy path of only the elements that change getting version bumps.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -6,11 +6,6 @@
   ],
   "version": "1.0.0-prerelease.29",
   "ignoreChanges": [
-    "elements/*/*.css",
-    "elements/*/*.js",
-    "elements/*/*.map",
-    "elements/*/dist/*.css",
-    "elements/*/dist/*.js",
-    "elements/*/dist/*.map"
+    "elements/*/dist/**/*"
   ]
 }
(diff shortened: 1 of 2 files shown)


lerna.json
```diff
@@ -6,11 +6,6 @@
   ],
   "version": "1.0.0-prerelease.29",
   "ignoreChanges": [
-    "elements/*/*.css",
-    "elements/*/*.js",
-    "elements/*/*.map",
-    "elements/*/dist/*.css",
-    "elements/*/dist/*.js",
-    "elements/*/dist/*.map"
+    "elements/*/dist/**/*"
```
castastrophe: Would it make sense to ignore the entire dist directory instead of just the files nested inside it?  Also, there aren't any folders in the dist directory, just files so if you did want to keep the dist but ignore the files in it, `dist/*` would be sufficient.

castastrophe:
Needs a changelog :) -- unless we're ready to start automating that part!

lerna.json
kylebuch8: I think we're being extra careful here on purpose.

PR #643, opened 2019-11-26
Disclosure accordion variant

chrisdo1:
## Disclosure accordion variant

* Link to discovery documentation (should contain all the planning and requirements for this work):
   - See the `DISCOVERY.md` template at the root of the project
* Link(s) to demo pages where this element can be viewed:

---

### What has changed and why
* When an accordion has only one header/panel coupling it displays as a `disclosure`: the entire header has a 1px border when collapsed and the chevron moves to the left side of the header.

### Testing instructions
1. Go to the `/en/search` page
2. Enter a search string
3. The customer portal results at the top of the page should display as a disclosure rather than a standard accordion. You should see the chevron on the left side, the header should have a gray 1px border on all sides. And if you look in the inspector you should see the attribute `pfe-disclosure="true"` on the accordion itself, the header element and the panel element.

#### Browser requirements

- [x] Galaxy S9 Firefox
- [x] iPhone X Safari
- [x] iPad Pro Safari
- [x] Pixel 3 Chrome

### Ready-for-merge Checklist

- [ ] Did you update the CHANGELOG.md file with a summary of this update?

### DIFF
--- a/elements/pfe-accordion/demo/index.html
+++ b/elements/pfe-accordion/demo/index.html
@@ -94,94 +94,6 @@
 
     <pfe-band pfe-color="lightest">
       <header slot="pfe-band--header">
-        <h2>Disclosure</h2>
-      </header>
-      <article>
-        <pfe-accordion>
-          <pfe-accordion-header>
-            <h3>This is a disclosure</h3>
-          </pfe-accordion-header>
-          <pfe-accordion-panel>
-            <p>Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut
-              labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores
-              et ea rebum.</p>
-            <p>Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor
-              sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore
-              magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet
-              clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.</p>
-            <p>Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut
-              labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores
-              et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.</p>
-          </pfe-accordion-panel>
-        </pfe-accordion>
-      </article>
-      <article>
-        <pfe-accordion pfe-disclosure="false">
-          <pfe-accordion-header>
-            <h3>Not a disclosure because pfe-disclosure=false</h3>
-          </pfe-accordion-header>
-          <pfe-accordion-panel>
-            <p>Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut
-              labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores
-              et ea rebum.</p>
-            <p>Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor
-              sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore
-              magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet
-              clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.</p>
-            <p>Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut
-              labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores
-              et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.</p>
-          </pfe-accordion-panel>
-        </pfe-accordion>
-      </article>
-    </pfe-band>
-
-    <pfe-band pfe-color="darkest">
-      <header slot="pfe-band--header">
-        <h2>Disclosure on dark</h2>
-      </header>
-      <article>
-        <pfe-accordion>
-          <pfe-accordion-header>
-            <h3>This is a disclosure</h3>
-          </pfe-accordion-header>
-          <pfe-accordion-panel>
-            <p>Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut
-              labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores
-              et ea rebum.</p>
-            <p>Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor
-              sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore
-              magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet
-              clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.</p>
-            <p>Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut
-              labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores
-              et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.</p>
-          </pfe-accordion-panel>
-        </pfe-accordion>
-      </article>
-      <article>
-        <pfe-accordion pfe-disclosure="false">
-          <pfe-accordion-header>
-            <h3>Not a disclosure because pfe-disclosure=false</h3>
-          </pfe-accordion-header>
-          <pfe-accordion-panel>
-            <p>Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut
-              labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores
-              et ea rebum.</p>
-            <p>Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor
-              sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore
-              magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet
-              clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.</p>
-            <p>Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut
-              labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores
-              et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.</p>
-          </pfe-accordion-panel>
-        </pfe-accordion>
-      </article>
-    </pfe-band>
-
-    <pfe-band pfe-color="lightest">
-      <header slot="pfe-band--header">
         <h2>Accordion component</h2>
       </header>
       <article>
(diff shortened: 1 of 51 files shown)


elements/pfe-accordion/src/pfe-accordion-header.scss
```diff
@@ -63,3 +64,38 @@ button[aria-expanded="true"] {
...
+/// DISCLOSURE STYLES
+/// ===========================================================================
+
+:host([pfe-disclosure="true"]) {
+  button {
+    @include pfe-trigger($align: right);
+
+    &[aria-expanded="true"] {
+      --pfe-accordion--BorderLeftWidth: 4px;
+      padding-left: 47px !important;
```
castastrophe: This definitely feels like a magic number, where are 47px and 21px coming from?  Consider how they will be impacted if someone changes the spacing variables that are impacting the padding on the header for example.  How can we use the variables like `container-spacer` to calculate a similar size?

elements/pfe-accordion/src/pfe-accordion.js
```diff
@@ -95,6 +95,8 @@ class PfeAccordion extends PFElement {
     this.addEventListener(`${PfeAccordion.tag}:change`, this._changeHandler);
     this.addEventListener("keydown", this._keydownHandler);
 
+    let numberItems = this.querySelectorAll('pfe-accordion > pfe-accordion-header');
```
castastrophe: Is this selector working at the moment?  I think the this context _is_ the pfe-accordion so perhaps it should be: `this.querySelectorAll(":scope > pfe-accordion-header")`?

castastrophe: Although it doesn't appear this is being used below so maybe we delete it?

elements/pfe-accordion/src/pfe-accordion.js
```diff
@@ -105,6 +107,11 @@ class PfeAccordion extends PFElement {
 
       this._observer.observe(this, { childList: true });
     });
+    if ((!this.hasAttribute("pfe-disclosure")) && ( this._allHeaders().length === 1 )) {
+      this.setAttribute("pfe-disclosure", "true");
+      this._firstHeader().setAttribute("pfe-disclosure", "true");
```
castastrophe: Is it necessary to apply this to the first header and first panel only since we do a check around it to ensure there are only one of each anyway?  Maybe just apply the attribute to all the headers and panels since there should only be one by the time we've gotten in here?

elements/pfe-sass/mixins/_components.scss
```diff
@@ -186,7 +186,7 @@
 @mixin pfe-chevron($state: open, $position: after, $size: .4em) {
   content: "";
   position: absolute;
-  top: calc((#{pfe-var(container-spacer)} * .5) + #{$size});
+  top: calc(#{pfe-var(container-spacer)} + #{$size});
```
castastrophe: By using the full container spacer height, doesn't that align the caret too low?

elements/pfe-accordion/src/pfe-accordion-panel.scss
```diff
@@ -48,3 +48,7 @@ $LOCAL: accordion;
     @include pfe-panel-variables($theme: #{$context});
   }
 }
+
+:host([pfe-disclosure="true"][expanded]) {
```
castastrophe: This is likely not necessary.  It should propogate down from the parent `pfe-accordion` where that variable is being set.

elements/pfe-accordion/src/pfe-accordion-header.scss
```diff
@@ -63,3 +64,38 @@ button[aria-expanded="true"] {
...
+/// ===========================================================================
+/// DISCLOSURE STYLES
+/// ===========================================================================
+
+:host([pfe-disclosure="true"]) {
+  button {
+    @include pfe-trigger($align: right);
+
+    &[aria-expanded="true"] {
+      --pfe-accordion--BorderLeftWidth: 4px;
```
castastrophe: If you set the BorderLeftWidth on the `pfe-accordion`, it will propagate down to both the header and the panel.

elements/pfe-accordion/src/pfe-accordion-header.scss
chrisdo1: It's a bit of a magic number and it has to be exact or you'll see the accordion header text shifting about 4 px to the right when opened or on hover. I went with this instead:
`padding-left: calc(( #{pfe-var(container-spacer)} * 3 ) - 1px ) !important;`

elements/pfe-accordion/src/pfe-accordion.js
chrisdo1: Shoot. Yep, meant to delete it.

elements/pfe-sass/mixins/_components.scss
chrisdo1: It actually looked too high before. This looks much more vertically aligned with the text.

elements/pfe-accordion/src/pfe-accordion-panel.scss
chrisdo1: Because the border-left color (gray to blue) and weight (1px to 4px) change when the disclosure is open I broke these properties out separately (i.e., border-left-color, border-left-width). I can try to keep the `border` shorthand property but it was getting very confusing because the colors and weights change independently. 

Also the values for those are in the `pfe-accordion-props` mixin which are in the `pfe-accordion-header.scss` and `pfe-accordion-panel.scss` so I think that this is the correct place to call these.

elements/pfe-accordion/src/pfe-accordion-header.scss
chrisdo1: I'm going to try to set a new argument in `pfe-accordion-props` mixin to contain these styles.

elements/pfe-accordion/src/pfe-accordion.js
chrisdo1: Good point. Set a `forEach` loop through all panels and headers and applied the `pfe-disclosure="true"` tag.

elements/pfe-accordion/src/pfe-accordion-panel.scss
chrisdo1: Moved the header and panel styles into the mixins. Let me know what you think.

elements/pfe-sass/mixins/_components.scss
chrisdo1: Here's the placement on my sandbox:
Screen Shot 2019-12-03 at 4 27 26 PM

elements/pfe-accordion/src/pfe-accordion-panel.scss
chrisdo1: And I should have said it before: thanks so much for reviewing especially while you're out! Always appreciate your feedback, @castastrophe !

elements/pfe-accordion/src/pfe-accordion-header.scss
```diff
@@ -63,3 +64,39 @@ button[aria-expanded="true"] {
     @include pfe-accordion-variables($state: expanded, $theme: #{$context});
   }
 }
+
+/// ===========================================================================
+/// DISCLOSURE STYLES
+/// ===========================================================================
+
+:host([pfe-disclosure="true"]) {
```
starryeyez024: @chrisdo1 I'm saying this blindly without knowing how difficult it might be to fix it, but we are duplicating a lot of styles that maybe don't need to be printed twice?
https://www.evernote.com/l/AA6a85jK0W5J1LGziQ1c6FIwI4CR56BofyU
If nothing else, let's flag this as a @todo to tidy later, but maybe see if we could just override the CSS vars instead of calling the mixin again?

chrisdo1: @starryeyez024 good catch! I removed the `pfe-trigger` mixin and re-added the padding properties. Maybe it doesn't make sense to have the left/right arguments in this mixin since we're not currently using `$align: right` anywhere but here.

elements/pfe-accordion/src/pfe-accordion-header.scss
```diff
@@ -63,3 +64,39 @@ button[aria-expanded="true"] {
...
+  button {
+    @include pfe-accordion-variables($variant: disclosure);
+    padding: #{pfe-var(container-spacer)} calc(#{pfe-var(container-spacer)} * 1.5) #{pfe-var(container-spacer)} 50px !important;
```
castastrophe: Don't forget to update variable and not reset properties. 👍

elements/pfe-accordion/src/pfe-accordion-panel.scss
castastrophe: What I'm suggesting though is that you can set those variables in the pfe-accordion and they will get used inside the panel. Does the accordion parent have a selector you can target that indicates it's a disclosure? You can update the border left width variable there and use it in the accordion panel. It keeps you from having to update the same values in different files.

elements/pfe-accordion/src/pfe-accordion.js
```diff
@@ -105,6 +105,15 @@ class PfeAccordion extends PFElement {
...
     });
+    if ((!this.hasAttribute("pfe-disclosure")) && ( this._allHeaders().length === 1 )) {
+      this.setAttribute("pfe-disclosure", "true");
+      this._allHeaders().forEach(function(headers) {
+        headers.setAttribute("pfe-disclosure", "true");
+      });
+      this._allPanels().forEach(function(panels) {
+        panels.setAttribute("pfe-disclosure", "true");
+      });
+    };
```
starryeyez024: Just noting here, this needs a mutation observer so that when the value of `pfe-disclosure` is changed on the parent `<pfe-accordion>` , it propagates down to the children `<pfe-accordion-header>` and `<pfe-accordion-panel>`

castastrophe: If you update the variables on pfe-accordion and let them propagate to the children that way, you might not need the extra code. You won’t need to push disclosure to the children elements at all.

elements/pfe-accordion/src/pfe-accordion-header.scss
chrisdo1: added a `triggerPadding` variable.

elements/pfe-accordion/src/pfe-accordion.js
chrisdo1: This would require quite a bit of refactoring for things that currently do not have custom properties, such as the panel padding, the chevron position / states.

elements/pfe-accordion/demo/index.html
```diff
@@ -92,6 +92,100 @@
...
+              sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore
+              magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet
+              clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.</p>
+            <p>Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut
+              labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores
+              et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.</p>
+          </pfe-accordion-panel>
+        </pfe-accordion>
+      </article>
+      <header slot="pfe-band--header">
```
starryeyez024: @chrisdo1 can you remove this duplicate band header?

elements/pfe-accordion/demo/index.html
```diff
@@ -92,6 +92,100 @@
...
```
starryeyez024: duplicate band header

elements/pfe-accordion/src/pfe-accordion-header.scss
```diff
@@ -63,3 +67,42 @@ button[aria-expanded="true"] {
...
+
+      &::before {
+        left: calc( #{pfe-var(container-spacer)} * 1.3125 ); // 21px.
+      }
+    }
+  }
+}
+
+:host([pfe-disclosure="true"].animating) {
+  border-left-color: #{pfe-color(ui-base)};
```
starryeyez024: @chrisdo1 do we / should we have a var like `--pfe-accordion--BorderLeftColor`?

elements/pfe-accordion/src/pfe-accordion-header.scss
```diff
@@ -63,3 +67,42 @@ button[aria-expanded="true"] {
...
+  border-left-width: #{pfe-var(surface--border-width--heavy)};
```
starryeyez024: should this reset the value of `--pfe-accordion--BorderLeftWidth`

elements/pfe-accordion/src/pfe-accordion-panel.scss
```diff
@@ -48,3 +48,9 @@ $LOCAL: accordion;
     @include pfe-panel-variables($theme: #{$context});
   }
 }
+
+:host([pfe-disclosure="true"]) {
+  .container {
+    padding-left: calc((#{pfe-var(container-spacer)} * 1.5) - 4px);
```
starryeyez024: we have --pfe-accordion--triggerPadding, maybe we should mark this as a todo for the future but the trigger padding and the container padding should probably be related

chrisdo1: Just trying to figure out a simple way to do this. This is the padding currently:
```
**disclosure padding**
* triggerPadding: 16px 24px 16px 47px; _set in @mixin pfe-trigger_
* panel padding: 0 48px 16px 20px; _set in @mixin pfe-panel-container_

**accordion padding**
* triggerPadding: 16px 50px 16px 24px;
* panel padding: 0 48px 16px 24px;
```

elements/pfe-accordion/demo/index.html
chrisdo1: Removed

elements/pfe-accordion/demo/index.html
chrisdo1: Removed

elements/pfe-accordion/src/pfe-accordion-header.scss
chrisdo1: sorry, yes. Should have been using `--pfe-accordion--BorderColor--accent`

elements/pfe-accordion/src/pfe-accordion-header.scss
chrisdo1: No, changed that to the correct property `--pfe-accordion--BorderLeftWidth`

elements/pfe-accordion/src/pfe-accordion-panel.scss
chrisdo1: @starryeyez024 I set a `PaddingBase` variable that is applied to both the `Trigger--padding` and the `PanelContainer--padding` variables. If the `PaddingBase` value (currently equal to `container-spacer` changes it will change the trigger and container padding.

starryeyez024: thanks!

starryeyez024:
@chrisdo1 whoops sorry, just noticed this needs a changelog. @kylebuch8 are you good w/ merging this?

PR #671, opened 2020-01-07
Update README.md

starryeyez024:
### What has changed and why

* Added links to email sign ups and slack channel


castastrophe:
Does this need a changelog?

mwcz:
Probably not.  I'd categorize this as a "chore" which end users don't need to be informed of when looking at a new release.

PR #685, opened 2020-01-20
fix error message when 'npm run release' is run outside monorepo root

mwcz:
When you run `npm run release` from a directory other than the monorepo root directory, the release script notices and throws an error.  It also suggested a command to run instead, however the command was removed a long, long time ago.  I removed the old information from the error message.


castastrophe:
@mwcz Should the fix be listed in the changelog?

PR #688, opened 2020-01-20
Updating dev commands in the docs to reflect what Michael Clayton recommended

wesruv:


castastrophe:
@wesruv Great updates.  There's also the `npm run live-demo` command which runs both the watch and the server at the same time (except it watches all components for changes). There's also `npm run demo` which builds and launches the server but doesn't watch for changes.  Would these be worth including in this documentation update too?

wesruv:
@castastrophe updated! Let me know if the updates to the docs and README make sense and seem good.

PR #694, opened 2020-01-21
feat: Move the generator into the repo

castastrophe:
## Moving the generator into the repo

* Michael and I discussed the possibility of moving the generator from a separate repo into the monorepo, this PR accomplishes that

---

### What has changed and why

* Moving the generator from a separate repo into this one

The benefit of moving the generator into the monorepo is that when updates occur in the parent project, it's easier to ensure the generator is updated at the same time.  This creates a "source of truth" for building components.

### Testing instructions

1. `npm run new` should generate prompts

#### Requirements

Your repository infrastructure updates should work for at least:
- [x] Node v8.x
- [x] NPM v7.x

### Ready-for-merge Checklist


castastrophe:
> I like it! The `generators` directory is nice, in case we add more generators in the future.
> 
> I did hit one problem. I generated an element and the generator produced an invalid package.json for that element. Here are the answers I provided:
> 
> ```
> ? What would you like to create? component
> ? Element name (i.e. pfe-card) pfe-wasm                                                                                
> ? Author name mwcz                          
> ? Do you want to use Sass with this element? No     
> ? Describe the element's purpose or goal. Run a WebAssembly (.wasm) file.
> ? List any attributes for the element, separated by commas (i.e., color, priority) src
> ? List any named slots for the element, separated by commas (i.e., header, footer) 
> ```
> 
> And here's the invalid portion of the package.json:
> 
> ```
>   "dependencies": {
>     ,
>     "@patternfly/pfelement": "1.0.0-prerelease.36"
>   }
> ```
> 
> I wasn't sure if the problem came from this PR, but I tried providing the same answers on master (with your refactored generator), and also tried the old v1.4 generator (before your big refactoring). The invalid package.json didn't appear in either case, so I think it's something related to this PR.

Pushed the fix up!  Should be good now.

castastrophe:
@mwcz This should be ready for review

generators/element/index.js
```diff
@@ -0,0 +1,394 @@
...
+
+let isPfelement = false;
+let demoTemplatePath;
+let readmePath;
+
+module.exports = class extends Generator {
+  constructor(args, opts) {
+      super(args, opts);
+
+      // this.option("type", {
```
kylebuch8: Should we just remove this code?

kylebuch8: I would like to be able to create standalone components still.

castastrophe: You still can, this portion is setting a default value for type instead of prompting for it (line 48 below) and I'm migrating this default setting to the config file instead of being set here.  I didn't want to nix it entirely yet though in case we wanted to bring it back.

kylebuch8: I do have to create standalone components from time to time but @mwcz wants to abstract the build tools so the standalone components get the benefit of the build tools updates.

castastrophe: Right, this in no way prevents you from creating a standalone. See line 48 -- you can be prompted for it or set it as your default in your config file.

docs/content/develop/step-1a.md
```diff
@@ -19,12 +19,29 @@ npm install
...
-Using the [generator-pfelement](https://github.com/patternfly/generator-pfelement), which is installed as a dev dependency, the generator will ask you a few questions that will help with the scaffolding. Make sure you are in the root directory of the PatternFly Elements repository.
+The easiest way to build a new component from scratch is to use the generator. The generator will ask you a few questions that will help with the scaffolding. Make sure you are in the root directory of the PatternFly Elements repository.
 
 ```
 npm run new
 ```
 
+When creating your new components, you may find you are entering the same answers over and over again.  To prevent this, you can add a `project.conf.json` file with some of your preferences stored.  Here is a template for that file:
+
+```json
```
castastrophe: @mwcz @kylebuch8 Want to bring y'alls attention to this doc section I added since this will take care of being able to set defaults.  If type is not set in this file Kyle, then the prompt will ask if you want to create a standalone or a pfelement.

PR #708, opened 2020-01-29
feat: Adding ability to configure the localhost domain

wesruv:
Adding dotenv package
Defaulting localhost to 'localhost', but .env file can override that
Added example .env file
Ignoring .env file

### Testing instructions
1. `npm i` to get new additions to build.
1. Checkout branch and try to run `npm start` without adding a `.env` file, should provide a server at `http://localhost:8000/examples`
2. Save `.env.example` as `.env` and add a domain of your choosing
3. Run `npm start` and you should get `http://YOURHOSTNAME:8000/examples`

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "decomment": "^0.9.1",
     "del": "^3.0.0",
     "dialog-polyfill": "^0.4.10",
+    "dotenv": "^8.2.0",
     "generator-pfelement": "^1.4.0",
     "ghp": "^1.3.1",
     "glob": "^7.1.6",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -67,7 +67,7 @@
     "decomment": "^0.9.1",
     "del": "^3.0.0",
     "dialog-polyfill": "^0.4.10",
-    "generator-pfelement": "^1.4.0",
```
kylebuch8: @wesruv I don't think we want to remove the generator as a depedency.

wesruv: @kylebuch8 Fixed

wesruv:
@kylebuch8 it looks like all of the netlify stuff is still outstanding for some reason?

PR #730, opened 2020-02-10
fix: cascading the "on" attribute on pfe-content-set

### LINKED ISSUE #720: pfe-content-set does not cascade the on="dark" attribute
image

As such, text that is inside the content set is the wrong color:
image

kylebuch8:
Fixes #720

### Testing instructions

1. View the "Using pfe-color to control the theme" section of the demo.
2. Verify that the "on" attribute has been added to both the tabs and the accordion.
3. Verify that the newly added tests for pfe-content-set are passing

### DIFF
--- a/elements/pfe-accordion/src/pfe-accordion.js
+++ b/elements/pfe-accordion/src/pfe-accordion.js
@@ -44,7 +44,7 @@ class PfeAccordion extends PFElement {
   }
 
   static get observedAttributes() {
-    return ["pfe-disclosure", "on"]
+    return ["pfe-disclosure"]
   }
 
   constructor() {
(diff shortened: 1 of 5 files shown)


kylebuch8:
@castastrophe Great suggestion! That simplified things a bunch. I've updated the PR with your recommendation.

elements/pfe-accordion/src/pfe-accordion.js
```diff
@@ -44,7 +44,7 @@ class PfeAccordion extends PFElement {
   }
 
   static get observedAttributes() {
-    return ["pfe-disclosure"]
+    return ["pfe-disclosure", "on"]
```
castastrophe: Is this still needed after the latest updates?

kylebuch8: I fixed that too. It's not needed anymore.

Issue #679, opened 2020-01-17
pfe-cta arrow wraps to a new line by itself


starryeyez024:
![image](https://user-images.githubusercontent.com/456863/72622577-7fd34180-3911-11ea-8063-598be3191f3e.png)

starryeyez024:
https://github.com/patternfly/patternfly-elements/pull/680/files

castastrophe:
Mostly fixed in https://github.com/patternfly/patternfly-elements/pull/630

bennypowers:
Is this still relevant? https://patternflyelements.org/components/cta/demo/

methomps:
Closing this issue out. Will fix in RHDS: https://github.com/RedHat-UX/red-hat-design-system/issues/333

PR #765, opened 2020-03-02
fix: pfe-cta arrow wrap issue #679

starryeyez024:
[Resolves Issue #679](https://github.com/patternfly/patternfly-elements/issues/679)

* Link(s) to demo pages where this element can be viewed:
https://deploy-preview-765--happy-galileo-ea79c4.netlify.com//elements/pfe-cta/demo/

---

### What has changed and why

* Used nobreak space in the template + white-space:nowrap to force the arrow to stick to the last word in the CTA. Then reset the white-space on the link inside the CTA. Firefox understands the box model differently, so there is a browser specific style to limit the width of the CTA so the arrow does not get cut off.

### Testing instructions

Use the resize box on the pfe-cta demo page (halfway down) to test in various browsers. 
    - The arrow should always be on the same line as the last word in the cta
    - The arrow should stay next to the last word, and not float to the edge of the CTA area

- [x] Galaxy S9 Firefox
- [x] iPhone X Safari
- [x] iPad Pro Safari
- [x] Pixel 3 Chrome

### Ready-for-merge Checklist

### DIFF
--- a/CHANGELOG-prerelease.md
+++ b/CHANGELOG-prerelease.md
@@ -1,3 +1,10 @@
+## Prerelease 39 ( TBD )
+
+Tag: [v1.0.0-prerelease.40](https://github.com/patternfly/patternfly-elements/releases/tag/v1.0.0-prerelease.40)
+
+- [ ]( ) fix: Prevent default pfe-cta arrow from wrapping to a new line by itself (#679)
+
+
 ## Prerelease 39 ( 2020-02-19 )
 
 Tag: [v1.0.0-prerelease.39](https://github.com/patternfly/patternfly-elements/releases/tag/v1.0.0-prerelease.39)
--- a/elements/pfe-cta/demo/demo.css
+++ b/elements/pfe-cta/demo/demo.css
@@ -18,20 +18,33 @@ pfe-band[pfe-color="lightest"] pfe-card[pfe-color="lightest"] {
 }
 
 .resize {
+  display: flex;
+  flex-flow: row nowrap;
   resize: horizontal;
   overflow: auto;
   border: 1px solid #ddd;
+  width: 500px;
   padding: 10px;
 }
 
-.card-layout {
-  width: 100%;
-  display: block;
-
+.resize>pfe-cta:first-child {
+  margin-right: 20px;
 }
 
-.card-layout > * {
-  margin-bottom: 20px;
+.card-layout {
+  display: flex;
+  flex-flow: row wrap;
+  display: grid;
+  grid-gap: 10px;
+  grid-template-columns: repeat(auto-fit, minmax(100px, max-content));
+  margin-top: -10px;
+  margin-left: -20px;
+}
+
+.card-layout>pfe-cta,
+.card-layout .cta-align>pfe-cta {
+  margin-top: 10px;
+  margin-left: 20px;
 }
 
 .custom-dark-band {
--- a/elements/pfe-cta/demo/demo.css
+++ b/elements/pfe-cta/demo/demo.css
@@ -25,16 +25,13 @@ pfe-band[pfe-color="lightest"] pfe-card[pfe-color="lightest"] {
 }
 
 .card-layout {
-  display: flex;
-  flex-flow: row wrap;
-  display: grid;
-  grid-gap: 10px;
-  grid-template-columns: repeat(auto-fit, minmax(280px, max-content));
+  width: 100%;
+  display: block;
+
 }
 
-.card-layout > pfe-cta,
-.card-layout .cta-align > pfe-cta {
-  margin-top: 10px;
+.card-layout > * {
+  margin-bottom: 20px;
 }
 
 .custom-dark-band {
--- a/elements/pfe-cta/demo/demo.css
+++ b/elements/pfe-cta/demo/demo.css
@@ -18,12 +18,9 @@ pfe-band[pfe-color="lightest"] pfe-card[pfe-color="lightest"] {
 }
 
 .resize {
-  display: flex;
-  flex-flow: row nowrap;
   resize: horizontal;
   overflow: auto;
   border: 1px solid #ddd;
-  width: 500px;
   padding: 10px;
 }
 
--- a/elements/pfe-cta/demo/demo.css
+++ b/elements/pfe-cta/demo/demo.css
@@ -24,16 +24,12 @@ pfe-band[pfe-color="lightest"] pfe-card[pfe-color="lightest"] {
   padding: 10px;
 }
 
-.resize > pfe-cta:first-child {
-  margin-right: 20px;
-}
-
 .card-layout {
   display: flex;
   flex-flow: row wrap;
   display: grid;
   grid-gap: 10px;
-  grid-template-columns: repeat(auto-fit, minmax(100px, max-content));
+  grid-template-columns: repeat(auto-fit, minmax(280px, max-content));
   margin-top: -10px;
   margin-left: -20px;
 }
@@ -41,7 +37,7 @@ pfe-band[pfe-color="lightest"] pfe-card[pfe-color="lightest"] {
 .card-layout > pfe-cta,
 .card-layout .cta-align > pfe-cta {
   margin-top: 10px;
-  margin-left: 20px;
+  display: block;
 }
 
 .custom-dark-band {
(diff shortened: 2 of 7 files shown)


elements/pfe-cta/src/pfe-cta.scss
```diff
@@ -130,8 +129,18 @@ $variables: (
   cursor: default !important;
 }
 
+// Default CTA fix for Firefox + arrow
+:host( :not([pfe-priority]):not([aria-disabled="true"]) ) {
+  .pfe-cta--wrapper {
```
castastrophe: For consistency can we move this down below to the class-based styles since this is making an update to the wrapper class?

~line 153 would be a great spot to add:

```
@include browser-query(firefox) {
    :host( :not([pfe-priority]):not([aria-disabled="true"]) ) & {
        max-width: calc(100% - 8px - #{pfe-local($cssvar: size, $region: arrow)});
    }
}
```

elements/pfe-sass/mixins/_mixins.scss
```diff
@@ -1,23 +1,25 @@
...
+      }
     }
-  }
-
     @if $browser == edge {
       @supports (-ms-ime-align: auto) { /* Microsoft Edge Browser 16+ (All) */
         @content;
       }
     }
+    @if $browser==firefox {
```
castastrophe:
```suggestion
    @if $browser == firefox {
```

To match formatting of other items in the list

elements/pfe-cta/demo/demo.css
```diff
@@ -18,33 +18,26 @@ pfe-band[pfe-color="lightest"] pfe-card[pfe-color="lightest"] {
...
-  grid-template-columns: repeat(auto-fit, minmax(100px, max-content));
+  grid-template-columns: repeat(auto-fit, minmax(280px, max-content));
   margin-top: -10px;
   margin-left: -20px;
 }
 
 .card-layout > pfe-cta,
 .card-layout .cta-align > pfe-cta {
   margin-top: 10px;
-  margin-left: 20px;
```
castastrophe: This is breaking the card-layout in the custom bands below - it needs to compensate for the `margin-left: -20px` above.

castastrophe: Why set the CTAs to `display: block`?  Don't we want to test that the CTA's display mode is correct?  This would override the display mode for the host element.

CHANGELOG-prerelease.md
```diff
@@ -1,3 +1,10 @@
+## Prerelease 39 ( TBD )
```
castastrophe:
```suggestion
## Prerelease 40 ( TBD )
```

elements/pfe-cta/demo/demo.css
```diff
@@ -33,6 +33,9 @@ pfe-band[pfe-color="lightest"] pfe-card[pfe-color="lightest"] {
   margin-top: -10px;
   margin-left: -20px;
 }
+.padding {
```
castastrophe: This isn't necessary - if you see my comment, it just needs the margin-left: 20px; added back to the selectors on line 40.

starryeyez024: I'd prefer not to add margins directly to the pfe-cta components because it can be misleading as to what the native behavior and styles actually are, and makes it difficult to tell what padding is coming from the container or the component. Since our traditional components (like a CTA) should always touch the edges of the parent container, I wanted to keep with that rule.

elements/pfe-cta/demo/demo.css
starryeyez024: The second point is fair. I was trying to make the demo prettier but I think its more important to show the native behavior.

elements/pfe-cta/src/pfe-cta.scss
starryeyez024: Great suggestion, thanks! updated

elements/pfe-cta/demo/demo.css
castastrophe: I understand the concern but none of the components add margin to the host element as a rule.  I approached the layout of the page just like we do on the site and one dynamic layout approach we use is to apply negative margin to the parent and compensate for that on the components.  This shows that the CTA can play nicely in a standard layout like that.  I think it's very low risk that we'd be debugging a margin-issue.  If you feel strongly, I understand, could we remove the negative margins from the containers entirely then instead of adding padding classes?

elements/pfe-cta/src/pfe-cta.scss
```diff
@@ -141,7 +141,12 @@ $variables: (
       align-items: center;
       height: 100%;
     }
-
+    // Default CTA arrow wrap fix for Firefox
+    :host( :not([pfe-priority]):not([aria-disabled="true"])) & {
```
castastrophe:
```suggestion
    :host(:not([pfe-priority]):not([aria-disabled="true"])) & {
```

Just a spacing nit

elements/pfe-cta/src/pfe-cta.scss
```diff
@@ -141,7 +141,12 @@ $variables: (
...
+      @include browser-query(firefox) {
+        max-width: calc(100% - 8px - #{pfe-local($cssvar: size, $region: arrow)});
```
castastrophe: Where does the 8px come from?

elements/pfe-cta/src/pfe-cta.scss
```diff
@@ -2,7 +2,7 @@
 
 $LOCAL: cta;
 
-$pfe-cta__arrow-size: 0.313em;
```
chrisdo1: I don't see this referenced anywhere else? Was it actually in use?

elements/pfe-cta/demo/demo.css
chrisdo1: I think the consistent approach makes a lot of sense. Let the container control the layout.

elements/pfe-cta/src/pfe-cta.scss
starryeyez024: same, i saw no references to it and nothing complained when i removed it, so i think its old cruft

elements/pfe-cta/src/pfe-cta.scss
starryeyez024: it represents the width of the `&#160;` character

elements/pfe-cta/demo/demo.css
starryeyez024: ohh, yes, sorry i did not see the negative margins on the card-layout. will do

elements/pfe-cta/src/pfe-cta.scss
castastrophe: Ah, I think `1ch` will serve this better then because if you change font size, the non breaking space size changes.

elements/pfe-cta/demo/demo.css
starryeyez024: updated

elements/pfe-cta/src/pfe-cta.scss
starryeyez024: LOL, I promise I actually tried this before, and I thought it didn't work. However I think the issue I was seeing was a `<button>` inside the `<pfe-cta>` (or some other issue) and I falsely attributed it to the `ch`. Anyway it seems to work just fine. Gonna push up that change.

starryeyez024: For posterity, I re-tested the `ch` in Firefox and IE11 and iphone 10

castastrophe:
The resize updates look great but please add back the demo page layout styles as they are not related to the PR and require discussion and additional consideration.

elements/pfe-cta/demo/demo.css
```diff
@@ -18,33 +18,23 @@ pfe-band[pfe-color="lightest"] pfe-card[pfe-color="lightest"] {
...
   grid-gap: 10px;
-  grid-template-columns: repeat(auto-fit, minmax(100px, max-content));
-  margin-top: -10px;
-  margin-left: -20px;
+  grid-template-columns: repeat(auto-fit, minmax(280px, max-content));
 }
 
 .card-layout > pfe-cta,
 .card-layout .cta-align > pfe-cta {
   margin-top: 10px;
```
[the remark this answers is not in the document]
starryeyez024: changes to all demo files reverted

castastrophe:
I opened an issue for current updates needed on the demo page for pfe-cta, I'd love to collaborate on it with you: https://github.com/patternfly/patternfly-elements/issues/775

Issue #736, opened 2020-02-12
pfe-switch: Need to create a switch component


kylebuch8:
Create a switch component just like: https://www.patternfly.org/v4/documentation/core/components/switch

We need to support
- With a label
- Without a label
- Disabled state

wesruv:
If anyone's bored, I think I have a good version in that commit ^

wesruv:
![image](https://user-images.githubusercontent.com/5607236/75593736-518a6b00-5a54-11ea-990a-bb3048f2848d.png)

coreyvickery:
@starryeyez024 Design ported over from PatternFly. I'm fine with preserving all of the styles 1:1.

https://xd.adobe.com/view/d20a1c17-375c-4c7f-6827-5fe2f8c2d92f-a43b/screen/4074c4b7-04ae-441e-80fd-21210d7680d2

castastrophe:
WIP PR: https://github.com/patternfly/patternfly-elements/pull/767

PR #767, opened 2020-03-02
WIP? Add switch component from from Patternfly 4

wesruv:
#736 
---

### Testing instructions

1. 

#### Browser requirements

- [ ] Internet Explorer 11 (should be useable, not pixel perfect)
- [ ] Latest 2 versions of Firefox (one on Mac OS, one of Windows OS)
- [ ] Firefox 60.7.2 (or latest version for Red Hat Enterprise Linux distribution)
- [ ] Latest 2 versions of Chrome (one on Mac OS, one of Windows OS)
- [ ] Latest 2 versions of Safari
- [ ] Galaxy S9 Firefox
- [ ] iPhone X Safari
- [ ] iPad Pro Safari
- [ ] Pixel 3 Chrome

### Ready-for-merge Checklist

- [ ] Expected files: all files in this pull request are related to one feature request or issue (no stragglers)?
- [ ] Did browser testing pass?
- [ ] Did you update or add any necessary documentation (README.md, WHY.md, etc.)?
- [ ] Was this feature demo'd and the design review approved?
- [ ] Did you update the CHANGELOG.md file with a summary of this update?

### DIFF
--- a/elements/pfe-switch/src/pfe-switch.html
+++ b/elements/pfe-switch/src/pfe-switch.html
@@ -1,0 +1,4 @@
+<div class="pfe-switch">
+  <slot>
+  </slot>
+</div>
--- a/elements/pfe-switch/src/pfe-switch.html
+++ b/elements/pfe-switch/src/pfe-switch.html
@@ -1,2 +1,3 @@
-<slot>
-</slot>
+<div id="switch-container"></div>
+<div id="knob"></div>
+<slot></slot>
--- a/elements/pfe-switch/src/pfe-switch.html
+++ b/elements/pfe-switch/src/pfe-switch.html
@@ -1,4 +1,2 @@
-<div class="pfe-switch">
-  <slot>
-  </slot>
-</div>
+<slot>
+</slot>
--- a/elements/pfe-switch/src/pfe-switch.js
+++ b/elements/pfe-switch/src/pfe-switch.js
@@ -1,0 +1,175 @@
+import PFElement from "../../pfelement/dist/pfelement.js";
+
+class PfeSwitch extends PFElement {
+  static get tag() {
+    return "pfe-switch";
+  }
+
+  get schemaUrl() {
+    return "pfe-switch.json";
+  }
+
+  get templateUrl() {
+    return "pfe-switch.html";
+  }
+
+  get styleUrl() {
+    return "pfe-switch.scss";
+  }
+
+  static get observedAttributes() {
+    return ["pfe-form"];
+  }
+
+  get pfeChecked() {
+    return this.getAttribute("pfe-checked");
+  }
+
+  // Declare the type of this component
+  static get PfeType() {
+    return PFElement.PfeTypes.Content;
+  }
+
+  constructor() {
+    super(PfeSwitch, { type: PfeSwitch.PfeType });
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    const $pfeSwitch = this.shadowRoot.querySelector(".pfe-switch");
+
+    // Get settings from attributes
+    const settingAttributes = [
+      "pfe-message-on",
+      "pfe-message-off",
+      "pfe-hide-label"
+    ];
+    const settings = {};
+    for (let index = 0; index < settingAttributes.length; index++) {
+      const settingAttribute = settingAttributes[index];
+      // Adds settings to an object and lops of `data-` from the index
+      settings[settingAttribute.substr(4)] = this.getAttribute(
+        settingAttribute
+      );
+    }
+
+    // Get a reference to the checkbox and label that should be slotted
+    this.checkbox = null;
+    this.label = null;
+    const $assignedElements = this.shadowRoot
+      .querySelector("slot")
+      .assignedElements();
+    for (let index = 0; index < $assignedElements.length; index++) {
+      const $assignedElement = $assignedElements[index];
+      switch ($assignedElement.tagName) {
+        case "LABEL":
+          this.label = $assignedElement;
+          break;
+        case "INPUT":
+          this.checkbox = $assignedElement;
+          break;
+      }
+    }
+
+    if (this.label) {
+      // Get the fallback text and store it as a var
+      settings["fallback-text"] = this.label.textContent;
+
+      // Hide the label if necessary
+      if (settings["hide-label"]) {
+        this.label.classList.add("sr-only");
+        const switchIcon = document.createElement("span");
+        switchIcon.classList.add("pfe-switch__checkmark");
+        $pfeSwitch.appendChild(switchIcon);
+      }
+
+      // Check for label text
+      if (!this.label.textContent.trim().length) {
+        console.error(
+          "pfe-switch: Label text not found. Form elements require meaningful label text for accessibility. The label can be visually hidden, but must exist."
+        );
+        return;
+      }
+
+      // Make sure we have a valid label with for
+      const labelFor = this.label.getAttribute("for");
+      if (
+        labelFor !== null &&
+        labelFor.length &&
+        !document.getElementById(labelFor)
+      ) {
+        // There is a for attribute, but it doesn't point to anything
+        console.error(
+          `pfe-switch: Label element does not have a corresponding checkbox, label's for attribute is ${labelFor}`
+        );
+        this.setAttribute("disabled", "");
+        return;
+      } else if (labelFor === null || !labelFor.length) {
+        // There isn't a for attribute
+        console.error(
+          "pfe-switch: Label does not have a for attribute, which is required for it to function. For attribute should have the id of the checkbox it is labelling."
+        );
+        return;
+      }
+    } else {
+      // There isn't a label
+      console.error(
+        "pfe-switch: Label element with text and for element required for pfe-switch to function"
+      );
+      return;
+    }
+
+    // Check the state of the checkbox and update everything accordingly
+    const updateSwitchState = () => {
+      if (this.checkbox !== null) {
+        // Update disabled status
+        if (this.checkbox.getAttribute("disabled") !== null) {
+          this.setAttribute("disabled", "");
+        } else {
+          this.removeAttribute("disabled");
+        }
+
+        // Update checked status
+        if (this.checkbox.checked) {
+          this.checked = true;
+          this.setAttribute("checked", "");
+          if (settings["message-on"] || settings["message-off"]) {
+            if (settings["message-on"]) {
+              this.label.textContent = settings["message-on"];
+            } else {
+              this.label.textContent = settings["fallback-text"];
+            }
+          }
+        } else {
+          this.checked = false;
+          this.removeAttribute("checked");
+          if (settings["message-on"] || settings["message-off"]) {
+            if (settings["message-off"]) {
+              this.label.textContent = settings["message-off"];
+            } else {
+              this.label.textContent = settings["fallback-text"];
+            }
+          }
+        }
+      }
+    };
+
+    // Resolve switch state when connected
+    updateSwitchState();
+
+    // Resolve switch state if the checkboxes changes
+    this.checkbox.addEventListener("change", () => updateSwitchState());
+  }
+
+  // disconnectedCallback() {}
+
+  // Process the attribute change
+  attributeChangedCallback(attr, oldValue, newValue) {
+    super.attributeChangedCallback(attr, oldValue, newValue);
+  }
+}
+
+PFElement.create(PfeSwitch);
+
+export default PfeSwitch;
--- a/elements/pfe-switch/src/pfe-switch.js
+++ b/elements/pfe-switch/src/pfe-switch.js
@@ -32,16 +32,22 @@ class PfeSwitch extends PFElement {
     return PFElement.PfeTypes.Content;
   }
 
+  static get observedAttributes() {
+    return ["pfe-message-on", "pfe-message-off", "pfe-hide-label"];
+  }
+
   constructor() {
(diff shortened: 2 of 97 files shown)


elements/pfe-switch/src/pfe-switch.scss
```diff
@@ -0,0 +1,219 @@
...
+/// @param $value {length}  Value to convert with px suffix
+/// @param $base-size {length} Optional, font size to use to calculate em
+/// @return {length} Em version of input
+@function px-to-em($value) {
+  @return $value / $font-size * 1em;
+}
+
+//sass-lint:disable indentation
+// Variables based on PF Next Switch
+$variables: (
```
castastrophe: This is looking really great!  We don't have any linting on this project so the sass-lint comment is not necessary.

elements/pfe-switch/src/pfe-switch.scss
```diff
@@ -0,0 +1,219 @@
...
+  toggle-knob--default__BackgroundColor: #fff,
```
castastrophe: To indicate a region, you would need to create a sub-map like in `pfe-card`

elements/pfe-switch/src/pfe-switch.scss
```diff
@@ -0,0 +1,219 @@
...
+.pfe-switch {
+  position: relative;
+}
+
+::slotted(input) {
+  @include pfe-screen-reader-only;
+}
+
+::slotted(.sr-only),
+.sr-only {
```
castastrophe: There is an `%sr-only` extend defined that you could use.

elements/pfe-switch/src/pfe-switch.scss
```diff
@@ -0,0 +1,219 @@
...
+$toggle-knob-diameter:   14px;
+$toggle-knob-offset:  5px;
+
+///
+/// Convert px to em
+///
+/// @param $value {length}  Value to convert with px suffix
+/// @param $base-size {length} Optional, font size to use to calculate em
+/// @return {length} Em version of input
+@function px-to-em($value) {
```
castastrophe: This is a great addition to pfe-sass functions in the `tools` file

castastrophe: Would this function do what you need?

```
// Return a pixel size in ems
@function pfe-size-pem($pxval, $base: $pfe-global--font-size-root) {
    @return pfe-strip-unit($pxval) / $base * 1em;
}
```

castastrophe: It's in the `elements/pfe-sass/functions/_1-tools.scss`

elements/pfe-switch/src/pfe-switch.scss
```diff
@@ -0,0 +1,219 @@
...
+.sr-only {
+  text-indent: -9999em;
+}
+
+::slotted(label) {
+  position: relative;
+  display: block;
+  // Width of the switch + some white space
+  padding-left: #{px-to-em($toggle-width) + 0.5em};
+  font-size: px-to-em($font-size);
```
castastrophe: These should ideally reference the repository defined values instead of hardcoding the value in the component: `font-size: #{pfe-var(font-size)};`

elements/pfe-switch/src/pfe-switch.scss
```diff
@@ -0,0 +1,219 @@
...
+  line-height: $line-height / $font-size; // Unitless font-size
```
castastrophe:
```suggestion
line-height: #{pfe-var(line-height)};
```

elements/pfe-switch/src/pfe-switch.scss
```diff
@@ -0,0 +1,219 @@
...
+    // First transform gets the right bound,
+    // the second moves the knob back behind the edge
+    transform:
+      translate(#{$toggle-width - ($toggle-knob-offset * 2)})
+      translate(-100%);
+  }
+}
+
+::slotted([disabled]) {
+  color: #{pfe-local(label--disabled__Color)};
```
castastrophe: Instead of redefining properties when state changes, we've found it more efficient to redefine the custom property value.

```
::slotted(label) {
  color: #{pfe-local($cssvar: Color, $region: label)};
}

::slotted([disabled]) {
  --pfe-switch__label--Color: #{pfe-local(label--disabled__Color)};
}
 ```

Comment (castastrophe) [reply elements/pfe-switch/src/pfe-switch.scss]: General template for properties: `--<component name>__<region>--<Property>--<state>`

File: elements/pfe-switch/src/pfe-switch.scss (hunk 2 above)
Comment (castastrophe) [reply elements/pfe-switch/src/pfe-switch.scss]: Noting here too that state appears after the property in BEM

File: elements/pfe-switch/src/pfe-switch.scss (hunk 1 above)
Comment (wesruv) [reply elements/pfe-switch/src/pfe-switch.scss]: Sorry, I have it locally and I didn't like seeing the 'error' underline 😛 will be removed!

File: elements/pfe-switch/src/pfe-switch.scss (hunk 3 above)
Comment (wesruv) [reply elements/pfe-switch/src/pfe-switch.scss]: In this case the usual clip method would make my component unusable, using negative text-indent achieves my purpose with less styles and wouldn't require an extra tag.

File: elements/pfe-switch/src/pfe-switch.scss (hunk 4 above)
Comment (wesruv) [reply elements/pfe-switch/src/pfe-switch.scss]: It is! thanks!

File: elements/pfe-switch/src/pfe-switch.scss (hunk 5 above)
Comment (wesruv) [reply elements/pfe-switch/src/pfe-switch.scss]: Good catch, wrote that before I figured out pfe-sass 😬

File: elements/pfe-switch/src/pfe-switch.scss (hunk 7 above)
Comment (wesruv) [reply elements/pfe-switch/src/pfe-switch.scss]: nice!

File: elements/pfe-switch/src/pfe-switch.scss (hunk 3 above)
Comment (wesruv) [reply elements/pfe-switch/src/pfe-switch.scss]: oh wait, sorry you're looking at older code. In one case I def need to use that extend 😬

**Comment by wesruv** [comment]
@castastrophe I think I got everything you noted, working on writing tests now.

File: elements/pfe-switch/src/pfe-switch.scss [hunk 8]
```diff
@@ -0,0 +1,164 @@
...
+
+// Variables based on PF Next Switch
+$variables: (
+  toggle-knob: (
+    BackgroundColor: pfe-color(surface--lightest),
+    BackgroundColor--disabled: pfe-color(ui-disabled),
+    BoxShadow: '0 0.0625rem 0.0625rem 0rem rgba(3, 3, 3, 0.05), 0 0.25rem 0.5rem 0.25rem rgba(3, 3, 3, 0.06)',
+  ),
+  toggle: (
+    BackgroundColor: #d2d2d2, // @todo - didn't find a great color to match PF4
```
wesruv: @castastrophe couldn't find a color that made sense for this, any thoughts?

elements/pfe-switch/src/pfe-switch.html
```diff
@@ -0,0 +1,4 @@
+<div class="pfe-switch">
```
castastrophe: I'm not sure this div is necessary.  You can style the `:host`, `<pfe-switch>`, directly in the stylesheet by using `:host` without needing to assign a class.  If you really want the class, in the JS you could add the class to the host element too.

elements/pfe-switch/src/pfe-switch.js
```diff
@@ -0,0 +1,178 @@
...
+  constructor() {
+    super(PfeSwitch, { type: PfeSwitch.PfeType });
+    this._updateSwitchState = this._updateSwitchState.bind(this);
+    this.settings = {};
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    const $pfeSwitch = this.shadowRoot.querySelector(".pfe-switch");
```
castastrophe: The `$name` syntax is a jQuery convention to indicate a jQuery element variable but would not make sense in this context since we do not use jQuery in the project.  The convention we've been using for shadow elements is `this._pfeSwitch = this.shadowRoot.querySelector(".pfe-switch");`.  If you remove the div from the template, you can reference the host simply with `this` and there's no need to get the shadow element at all.

elements/pfe-switch/src/pfe-switch.json
```diff
@@ -0,0 +1,46 @@
...
+          }
+        }
+      },
+      "required": ["default"]
+    },
+    "attributes": {
+      "title": "Attributes",
+      "type": "object",
+      "properties": {
+        "form": {
```
castastrophe: The attributes section should list out all the data attributes supported by your component: `message-on`, `message-off`, `hide-label`.  The form indication could go in your slot array above instead of referencing a raw field, you could indicate form to tell the component that it only accepts a form element in the slot.

elements/pfe-switch/src/pfe-switch.js
```diff
@@ -0,0 +1,178 @@
...
+
+    // Get settings from attributes
+    const settingAttributes = [
```
castastrophe: This is done for you by importing the schema.  Any attributes defined in your schema will be attached to your component for you by logic built into `pfelement`.  You can access that information using `this.props["message-on"]` which will list out the object for that attribute.  Each attribute has title, type, enum (if defined), default (if defined), a boolean prefixed value (does it have `pfe` at the beginning of it), and the actual value assigned (if the user has added that attribute to the component).

elements/pfe-switch/src/pfe-switch.js
```diff
@@ -0,0 +1,178 @@
...
+    for (let index = 0; index < settingAttributes.length; index++) {
+      const settingAttribute = settingAttributes[index];
+      // Adds settings to an object and lops of `data-` from the index
+      this.settings[settingAttribute.substr(4)] = this.getAttribute(
+        settingAttribute
+      );
+    }
+
+    // Get a reference to the checkbox and label that should be slotted
+    this.checkbox = null;
```
castastrophe: I love this set-up.  I would just update the name of your variables to remove the jQuery syntax and use `this._foo` convention instead and it's +1.

elements/pfe-switch/src/pfe-switch.js
```diff
@@ -0,0 +1,178 @@
...
+        this.label.classList.add("pfe-sr-only");
+        const switchIcon = document.createElement("span");
+        switchIcon.classList.add("pfe-switch__checkmark");
+        $pfeSwitch.appendChild(switchIcon);
+      }
+
+      // Check for label text
+      if (!this.label.textContent.trim().length) {
+        console.error(
+          "pfe-switch: Label text not found. Form elements require meaningful label text for accessibility. The label can be visually hidden, but must exist."
```
castastrophe:
```suggestion
          `${this.tag}: Label text not found. Form elements require meaningful label text for accessibility. The label can be visually hidden, but must exist.`
```

Using the variable helps in case the component is renamed.

castastrophe:
Link to the preview: https://deploy-preview-767--happy-galileo-ea79c4.netlify.com/elements/pfe-switch/demo/

Man this is a gorgeous component!  

Screen Shot 2020-03-04 at 10 54 35 AM

elements/pfe-switch/src/pfe-switch.scss
castastrophe: I like your todo, we can look into if we're missing a color variable from the map or if we need to update the colors in the map currently at a later date. 👍

elements/pfe-switch/src/pfe-switch.scss
```diff
@@ -0,0 +1,173 @@
...
+}
+
+:host([pfe-disabled]) {
+  --pfe-switch__label--Color: #{pfe-local($cssvar: Color--disabled, $region: label)};
+  --pfe-switch__toggle--BackgroundColor: #{pfe-local($cssvar: BackgroundColor--disabled, $region: toggle)};
+  --pfe-switch__toggle-knob--BackgroundColor: #{pfe-local($cssvar: BackgroundColor--disabled, $region: toggle-knob)};
+}
+
+.pfe-switch__checkmark {
+  // Checkmark is built with CSS instead of using an image
```
castastrophe: Cool!

elements/pfe-switch/src/pfe-switch.html
wesruv: Styling tags makes me itchy... but you're right.. old habits, something something

elements/pfe-switch/src/pfe-switch.json
wesruv: Is there an example or existing component I can code-monkey off of?

elements/pfe-switch/src/pfe-switch.scss
```diff
@@ -0,0 +1,189 @@
...
+  min-height: calc(#{pfe-var(line-height)} * 1em);
+  // Width of the switch + some white space
+  padding-left: #{pfe-size-pem($toggle-width) + 0.5em};
+  font-size: $font-size;
+  line-height: pfe-var(line-height);
+  color: #{pfe-local(Color--default, $region: label)};
+  cursor: pointer;
+
+  // Toggle container
+  &:before {
```
kylebuch8: Unfortunately, this type of selector isn't supported in Safari.
::before and ::after selectors inside ::slotted don't work.
https://github.com/w3c/webcomponents/issues/655

image

kylebuch8: We'll need to come up with an alternative approach. I'm going to investigate moving light DOM content to the shadow DOM.

castastrophe:
I'm going to close this out until we have time to revisit this PR

PR #791, opened 2020-03-18
Us250350 pfe-cta wrapping issues

starryeyez024:
1. Resolves issue where default pfe-ctas are wrapping when there is still enough room.  
image

2. Resolves issue where priority CTAs wrap to two lines and text is left aligned. It should be centered.

image

Preview page with fixes:
https://deploy-preview-791--happy-galileo-ea79c4.netlify.com/elements/pfe-cta/demo/

### DIFF
--- a/elements/pfe-cta/demo/demo.css
+++ b/elements/pfe-cta/demo/demo.css
@@ -18,12 +18,10 @@ pfe-band[pfe-color="lightest"] pfe-card[pfe-color="lightest"] {
 }
 
 .resize {
-  display: flex;
-  flex-flow: row nowrap;
   resize: horizontal;
   overflow: auto;
   border: 1px solid #ddd;
-  width: 500px;
+  width: 700px;
   padding: 10px;
 }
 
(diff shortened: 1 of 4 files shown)


elements/pfe-cta/src/pfe-cta.scss
```diff
@@ -133,6 +136,7 @@ $variables: (
   &--wrapper {
     display: block; 
     white-space: nowrap;
+    min-width: calc(100%);
```
chrisdo1: @starryeyez024 Why do we need the calc function?

starryeyez024: ooh we definitely dont, i'll remove that

elements/pfe-cta/demo/demo.css
```diff
@@ -18,12 +18,10 @@ pfe-band[pfe-color="lightest"] pfe-card[pfe-color="lightest"] {
 }
 
 .resize {
-  display: flex;
-  flex-flow: row nowrap;
   resize: horizontal;
```
starryeyez024: This was causing the focus state to look wonky 
Screen Shot 2020-03-18 at 4 20 40 PM

PR #796, opened 2020-03-22
fix: pfe-tabs vertical height

starryeyez024:
- Update pfe-tabs demo page so that it contains HTML header tags in the tab headers. 
- Add CSS so that tab-panels which are shorter than the tabs are still at least as tall as the tab headers.

## To test

Check the [demo page](https://deploy-preview-796--happy-galileo-ea79c4.netlify.com/elements/pfe-tabs/demo/), under the heading `Style: Vertical earth with custom tab typography & header tags`. Note that the first tab-panel has content which is very short, and yet the panel is at least as tall as the tabs on the left. 

The heading tags are just to show that pfe-tabs support HTML header markup.

### DIFF
--- a/elements/pfe-tabs/src/pfe-tab-panel.scss
+++ b/elements/pfe-tabs/src/pfe-tab-panel.scss
@@ -21,6 +21,7 @@ $variables: (
 
   [tabindex] {
   	display: flex;
+    height: 100%;
   }
 
   .container {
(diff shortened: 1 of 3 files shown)


elements/pfe-tabs/src/pfe-tab-panel.scss
```diff
@@ -21,6 +21,7 @@ $variables: (
 
   [tabindex] {
   	display: flex;
+    height: 100%;
```
castastrophe: @starryeyez024 Maybe this is a trick of the preview tool but it looks like the indention in this file is a bit off.

starryeyez024: adjusted

castastrophe:
@starryeyez024 Looking good!  Needs a changelog.

starryeyez024:
@castastrophe  added a changelog, thx

PR #800, opened 2020-03-23
fix: cascade "on" attribute from pfe-tabs to pfe-tab and pfe-tab-panel

### LINKED ISSUE #799: pfe-tabs: "on" attribute isn't copied to tabs and panels if value is "saturated"
It works if the value is "dark" but for some reason doesn't work with "saturated".

pfe-content-set-issue

The `on` attribute is observed, it's just not doing what we want it to do.
https://github.com/patternfly/patternfly-elements/blob/master/elements/pfe-tabs/src/pfe-tabs.js#L147

We should probably use `cascadingAttributes` just like `pfe-accordion`

```
static get cascadingAttributes() {
  return {
    on: "pfe-accordion-header, pfe-accordion-panel"
  };
}
```

kylebuch8:
Fixes #799

**Testing Instructions**
- Add or modify any of the existing `on` attributes on a `pfe-tabs` component on the pfe-tabs demo page
- Changing the `on` attribute should cascade the value down to the child `pfe-tab` and `pfe-tab-panel` components
- Removing the `on` attribute from `pfe-tab` should remove the `on` attribute from the child `pfe-tab` and `pfe-tab-panel` components

### DIFF
--- a/elements/pfe-content-set/demo/index.html
+++ b/elements/pfe-content-set/demo/index.html
@@ -129,7 +129,7 @@
   <section class="faux-container" style="max-width: 800px">
     <h4 class="label">Default pfe-content-set in a 800px container</h4>
 
-    <pfe-content-set>
+    <pfe-content-set on="dark">
       <h4 pfe-content-set--header>Heading 1</h4>
       <div pfe-content-set--panel>
         <p>Content for heading 1. Maecenas faucibus mollis interdum. Maecenas sed diam eget risus varius blandit sit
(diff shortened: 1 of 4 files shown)


kylebuch8:
A comment from @castastrophe about why we shouldn't merge this:

A tabset exists on a dark band but we're using the manilla folder style where the tabset sets the background color of itself to white and ignores it's context.  So inside the component, maybe we're ignoring the cascading value and resetting theme to light so that any nested components get the appropriate context.

A tab header and a tab body could be styled to use different theme contexts and their children need that information too.

So we really don't want to move the on attribute down because we're moving down the potentially wrong contextual information.

It will work in some use-cases but in others it will suddenly get the wrong theme and this is why.

PR #801, opened 2020-03-24
fix: set pfe-cta with the priority attribute relative to default pfe-cta font-size

chrisdo1:
**What has changed and why**
Set the default pfe-cta font-size to be larger (18px) than the font-size for pfe-priority ctas (16px).

### DIFF
--- a/CHANGELOG-prerelease.md
+++ b/CHANGELOG-prerelease.md
@@ -3,6 +3,7 @@
 Tag: [v1.0.0-prerelease.42](https://github.com/patternfly/patternfly-elements/releases/tag/v1.0.0-prerelease.42)
 
 - [](https://github.com/patternfly/patternfly-elements/commit/) fix: set consistent line-height for pfe-nav triggers (#773)
+- [](https://github.com/patternfly/patternfly-elements/commit/) fix: set pfe-cta default font-size to 18px
 
 
 ## Prerelease 41 ( 2020-03-19 )
--- a/CHANGELOG-prerelease.md
+++ b/CHANGELOG-prerelease.md
@@ -3,7 +3,7 @@
 Tag: [v1.0.0-prerelease.42](https://github.com/patternfly/patternfly-elements/releases/tag/v1.0.0-prerelease.42)
 
 - [](https://github.com/patternfly/patternfly-elements/commit/) fix: set consistent line-height for pfe-nav triggers (#773)
-- [](https://github.com/patternfly/patternfly-elements/commit/) fix: set pfe-cta's with the priority attribute to font-size: 16px
+- [](https://github.com/patternfly/patternfly-elements/commit/) fix: set pfe-cta's with the priority attribute relative to theme font-size
 
 
 ## Prerelease 41 ( 2020-03-19 )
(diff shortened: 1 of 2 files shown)


starryeyez024:
Code looks good, tested by adding

```css
:root {
    --pfe-cta--FontSize--default: 18px;
}
```

@chrisdo1 just needs a changelog entry

elements/pfe-cta/src/pfe-cta.scss
```diff
@@ -5,6 +5,7 @@ $LOCAL: cta;
  
 $pfe-cta--BackgroundColor--focus: rgba(40, 151, 240, 0.2);
 $pfe-cta--Color--fallback: #003366;
+$pfe-cta--FontSize--default: calc(#{pfe-local($cssvar: FontSize)} * 1.125);
```
castastrophe: Is this still being used? I don't see it in the diff.

chrisdo1: Thanks for looking @castastrophe 
you're correct. I removed this sass var.

CHANGELOG-prerelease.md
```diff
@@ -3,6 +3,7 @@
 Tag: [v1.0.0-prerelease.42](https://github.com/patternfly/patternfly-elements/releases/tag/v1.0.0-prerelease.42)
 
 - [](https://github.com/patternfly/patternfly-elements/commit/) fix: set consistent line-height for pfe-nav triggers (#773)
+- [](https://github.com/patternfly/patternfly-elements/commit/) fix: set pfe-cta default font-size to 18px
```
starryeyez024: @chrisdo1 should we modify this message, since the functionality is that we added a hook to override the fontsize of pfe-cta's with the priority attribute?

chrisdo1: oh yes. Good catch. Fixing this right now

elements/pfe-cta/src/pfe-cta.scss
```diff
@@ -120,6 +122,7 @@ $variables: (
 
   :host([pfe-priority]) & {
     text-align: center;
+    font-size: #{pfe-local($cssvar: FontSize--priority, $fallback: #{pfe-var(font-size)})};
   }
 }
```
starryeyez024: @chrisdo1 sorry i just noticed, i think we also need this on the disabled ctas too
```
:host([aria-disabled="true"]) {
    font-size: #{pfe-local($cssvar: FontSize--priority, $fallback: #{pfe-var(font-size)})};
}
```

chrisdo1:
> @chrisdo1 sorry i just noticed, i think we also need this on the disabled ctas too
> 
> ```
> :host([aria-disabled="true"]) {
>     font-size: #{pfe-local($cssvar: FontSize--priority, $fallback: #{pfe-var(font-size)})};
> }
> ```



> @chrisdo1 sorry i just noticed, i think we also need this on the disabled ctas too
> 
> ```
> :host([aria-disabled="true"]) {
>     font-size: #{pfe-local($cssvar: FontSize--priority, $fallback: #{pfe-var(font-size)})};
> }
> ```

@starryeyez024 Good catch. I made this fix and synced to my sandbox (you can see it on the WC demo page).

Screen Shot 2020-03-26 at 10 34 08 PM

PR #806, opened 2020-03-30
fix: make it so pfe-tabs inside pfe-tabs works

### LINKED ISSUE #805: pfe-tabs: tabs in tabs don't work
If you have a `pfe-tabs` component inside another `pfe-tabs` component, the tabs do not function correctly.

tabs-in-tabs

I think it has something to do with the way we're selecting the tabs and the panels in `_allPanels()` and `_allTabs()`. Both are running a query selector that finds _all_ tabs and panels inside the component instead of just trying to find the tabs and panels that are immediate children.

I also think we have an event propagation issue where a click or keydown event bubbles all the way up from a child `pfe-tabs` component and is picked up by the parent `pfe-tabs` component. We should stop event propagation so it remains contained in the `pfe-tabs` component where the event happened.

kylebuch8:
Fixes #805

**Testing Instructions**
- On your local machine, go to http://localhost:8000/elements/pfe-tabs/demo/ and scroll down to the "Tabs In Tabs" heading
- Click on Sub Tab 2 and ensure the content shows
- Click on Tab 2 and ensure the Tab 2 content shows
- Click on Tab 1 and ensure that Sub Tab 2 is still active and the Sub Tab 2 content is showing

This has been tested on:
- Chrome
- Firefox
- Safari
- Edge 80
- IE 11

### DIFF
--- a/CHANGELOG-prerelease.md
+++ b/CHANGELOG-prerelease.md
@@ -3,6 +3,7 @@
 Tag: [v1.0.0-prerelease.42](https://github.com/patternfly/patternfly-elements/releases/tag/v1.0.0-prerelease.42)
 
 - [fbc251b](https://github.com/patternfly/patternfly-elements/commit/fbc251b5180684da26a57c0941235d57f961990e) fix: set consistent line-height for pfe-nav triggers (#773)
+- []() fix: make it so pfe-tabs inside pfe-tabs works
 - [29780ab](https://github.com/patternfly/patternfly-elements/commit/29780abfab1074aa05a021a43d79bee00afed31b) feat: adding event to pfe-cta component
 - [](https://github.com/patternfly/patternfly-elements/commit/) fix: set pfe-cta's with the priority attribute relative to theme font-size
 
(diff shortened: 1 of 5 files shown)


castastrophe:
https://deploy-preview-806--happy-galileo-ea79c4.netlify.com/elements/pfe-tabs/demo

eyevana:
Yay! Thank you so much, Kyle! 💃

karangarg45:
@kylebuch8 This change is stopping all click events for all the elements inside the pfe-tab component. Our code was updated to use latest version of pfe-tabs and this broke most of the functionality since the clicks stopped working. I can see you added a stopPropogation event. Could you please take a look?
https://github.com/patternfly/patternfly-elements/pull/806/files#diff-1916f65565b72aa7b4f75584a362087cR431
image
As you can see inside the screenshot a button inside the tab gets the eventlistener registerd on itself which further prevents the flow of click event.

kylebuch8:
@karangarg45 I've created the issue here: https://github.com/patternfly/patternfly-elements/issues/816

PR #809, opened 2020-03-30
fix: Add a warning about updating the on attribute before upgrade

castastrophe:
### What has changed and why

* Added a warning to be thrown before upgrade if the tag contains the `on` attribute
* Updated documentation and tests, removed current examples of manually pushing the `on` attribute inside the pfe-content-set

### Testing instructions

1. Check out the branch. 
2. Edit the `pfe-content-set` demo page to add the `on="light"` attribute on the `pfe-content-set` and `pfe-cta` tags.  Add an ID to one of the tags.
3. Run `npm install && npm run start pfe-accordion`.
4. Preview the accordion's demo page.
3. Check that the console is registering 2 warnings and that the one for the tag with an ID lists the ID in the warning in square brackets:

```
pfe-content-set[#foo]: The "on" attribute is protected and should not be manually added to a component. The base class will manage this value for you on upgrade.
```
```
pfe-cta: The "on" attribute is protected and should not be manually added to a component. The base class will manage this value for you on upgrade.
```

### Ready-for-merge Checklist

- [ ] Was this feature demo'd and the design review approved?

### DIFF
--- a/.storybook/utils.js
+++ b/.storybook/utils.js
@@ -24,7 +24,7 @@ String.prototype.sentenceCase = function() {
 };
 
 // Print attributes based on an object
-export const listProperties = (obj) =>
+export const listProperties = obj =>
   Object.entries(obj)
     .map(set => {
       let string = " ";
@@ -43,11 +43,7 @@ export const listProperties = (obj) =>
       }
 
       // If printing is allowed, the value exists and is not null
-      if (
-        print &&
-        typeof v !== "undefined" &&
-        (v !== null && v !== "null")
-      ) {
+      if (print && typeof v !== "undefined" && v !== null && v !== "null") {
         string += p;
         // If the value is a string and the value is not equal to the string "true"
         if (typeof v === "string" && v !== "true") {
@@ -81,7 +77,7 @@ export function customTag(obj) {
     // If a tag is defined, use that, else use a div
     if (obj.tag) {
       start += `<${obj.tag}`;
-      end   += !selfClosing.includes(obj.tag) ? `</${obj.tag}>` : "";
+      end += !selfClosing.includes(obj.tag) ? `</${obj.tag}>` : "";
     } else {
       start += "<div";
       end += "</div>";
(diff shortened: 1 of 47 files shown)


starryeyez024:
👍  for testing and documentation, still needs JS review

elements/pfelement/src/pfelement.js
```diff
@@ -139,6 +139,15 @@ class PFElement extends HTMLElement {
       });
     }
 
+    // Throw a warning if the on attribute was manually added
+    if (this.hasAttribute("on")) {
```
kylebuch8, line 143: Can we move this out of the constructor and into the connected callback? The spec has this line about things you shouldn't do in the constructor.

"The element's attributes and children must not be inspected"
https://html.spec.whatwg.org/multipage/custom-elements.html#custom-element-conformance

kylebuch8: Also, when we move this to the connectedCallback(), if the element is ever removed and added somewhere else on the page, I think the connectedCallback() fires again. So if a component gets the `on` attributed after upgrade, and is moved and added somewhere else on the page, this warning will fire. We'd need to test that. A simple check for `this.hasAttribute("pfelement")` after the ShadyCSS check and before the `pfelement` attribute is added could be the right place for the warning.

```
this.connected = true;
this.log(`Connecting...`);

if (window.ShadyCSS) {
  this.log(`Styling...`);
  window.ShadyCSS.styleElement(this);
  this.log(`Styled.`);
}

// PUT THE CHECK HERE
if (!this.hasAttribute("pfelement")) {
  if (this.hasAttribute("on")) {
    ...
  }
}

// @TODO maybe we should use just the attribute instead of the class?
// https://github.com/angular/angular/issues/15399#issuecomment-318785677
this.classList.add("PFElement");
this.setAttribute("pfelement", "");
```

castastrophe: I will try it with the hasAttribute, I put it in the connectedCallback originally but it threw the error on every component because it would run it again after instantiation.

castastrophe: Perfect!  That did the trick!  Thank you!

castastrophe:
@kylebuch8 Update applied!  Working great!

castastrophe:
@kylebuch8 I updated the storybook instances to show theme support.

.storybook/utils.js
```diff
@@ -258,9 +257,30 @@ export function autoPropKnobs(properties, bridge) {
       }
     }
   });
+
   return binding;
 }
 
+export function theming(bridge) {
```
castastrophe, line 264: @kylebuch8 This is the only bit I updated in this file, the rest must be prettier.

castastrophe:
@kylebuch8 Reverted the storybook updates and moved them into a separate branch.  Prettier is still showing updates to the .storybook files but I can't undo that at this point.

PR #817, opened 2020-04-02
fix: removing stopPropagation from pfe-tabs click and keydown events

### LINKED ISSUE #816: pfe-tabs: stopPropagation has caused all events to stop bubbling up
This change is stopping all click events for all the elements inside the pfe-tab component. Our code was updated to use latest version of pfe-tabs and this broke most of the functionality since the clicks stopped working. I can see you added a stopPropogation event. Could you please take a look?
https://github.com/patternfly/patternfly-elements/pull/806/files#diff-1916f65565b72aa7b4f75584a362087cR431

78243000-9ea0a780-7500-11ea-97bd-9caf7a0fb88e

As you can see inside the screenshot a button inside the tab gets the eventlistener registered on itself which further prevents the flow of click event.

kylebuch8:
Fixes #816

### DIFF
--- a/CHANGELOG-prerelease.md
+++ b/CHANGELOG-prerelease.md
@@ -1,3 +1,9 @@
+## Prerelease 43 ( TBD )
+
+Tag: [v1.0.0-prerelease.43](https://github.com/patternfly/patternfly-elements/releases/tag/v1.0.0-prerelease.43)
+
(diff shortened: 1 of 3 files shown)


eyevana:
Tested locally and everything is working as expected. LGTM! :D

Issue #57, opened 2018-07-18
[tools] investigate migrating to LitElement as a base for PFElement


mwcz:
Would have to fork it, have it output UMD instead of the default ESM, and inject a global into the UMD stanza.  Sucky, but our extremely wide module support means that globals are the only one-size-fits-all way to import things.

It would be an excellent, lightweight, and FAST solution to dynamic rendering, without having to include a complex vdom lib.

btopro:
is there any interest in lit-element as an optional base element to build from? Only vaguely understand the UMD vs ESM issue though I assume you pull in css / other module packages, but as web components / custom element spec seems to bank on ESM would forking be the only way to handle this or could they be appended to a RHEL name space that esm built elements could be pulled in?

reviewing using this and we have a lot of elements we build but also even more that we pull from the wider polymer / webcomponents community so trying to figure out if we start building off of this that they'll be compatible (or able to be so)

btopro:
options (and then I'll keep digging through the repo / kicking things around w/ our team); what about loading requirejs stuff in via  shim like https://github.com/systemjs/systemjs which can plug some of these things together on their own?

Another food for thought project that appears to have just been an experiment -- https://github.com/requirejs/element . obviously this is ref'ing html import workflows / polymer / x-tags and is out of date but getting ideas out in this thread.

As for our elements, we build almost entirely data bound / heavy shadowDom using things.

mwcz:
@kylebuch8 shouldn't the inclusion of rollup in #122 resolve the lit-element+requirejs problem?

castastrophe:
https://github.com/patternfly/patternfly-elements/pull/826

castastrophe:
@kylebuch8 Do we want to circle back to this later in the year perhaps and consider it for a 2.0 feature?

castastrophe:
What do we get from using Lit:
- Reactive properties with efficient re-rendering when properties are changed
- "Lit keeps track of your UI’s dynamic parts and updates only those when the underlying state changes–no need to rebuild a whole virtual tree and diff it with the current state of the DOM."
- 5 KB (minified and compressed)
- Shared, importable templates

Useful API provided by LitElement:
- addInitializer
- firstUpdated
- getUpdateComplete
- hasUpdated
- performUpdate, etc.
- [Controllers](https://lit.dev/docs/api/controllers/)

castastrophe:
[...]

PR #826, opened 2020-04-06
feat: First pass at adding lit-html syntax to a template

castastrophe:


castastrophe:
https://github.com/patternfly/patternfly-elements/issues/57

PR #835, opened 2020-04-20
Add clearfix to tabs & accordions, in case of floating content

starryeyez024:
Currently, if there is content that floats inside the accordions, the padding on the container is not clearing the floats, so the content ends up too close to the bottom border:

image

This PR adds clearfixes on the container, so that the padding is always visible around the content

image

### DIFF
--- a/elements/pfe-accordion/src/pfe-accordion-panel.html
+++ b/elements/pfe-accordion/src/pfe-accordion-panel.html
@@ -1,5 +1,5 @@
 <div tabindex="-1" role="tabpanel">
-  <div class="container">
+  <div class="container clearfix">
     <slot></slot>
   </div>
 </div>
--- a/elements/pfe-accordion/src/pfe-accordion-panel.scss
+++ b/elements/pfe-accordion/src/pfe-accordion-panel.scss
@@ -23,6 +23,10 @@ $LOCAL: accordion;
   --pfe-accordion--panel-container--Padding: 0 calc(#{pfe-local(base--Padding)} * 3) #{pfe-local(base--Padding)} calc(#{pfe-local(base--Padding)} * 1.5);
 }
 
+.clearfix {
+  @include pfe-clearfix;
+}
+
 :host(:last-of-type[expanded]) {
   margin-bottom: 0;
 }
--- a/elements/pfe-tabs/src/pfe-tab-panel.html
+++ b/elements/pfe-tabs/src/pfe-tab-panel.html
@@ -1,5 +1,5 @@
 <div tabindex="-1" role="tabpanel">
-  <div class="container">
+  <div class="container clearfix">
     <slot></slot>
   </div>
 </div>
--- a/elements/pfe-content-set/demo/index.html
+++ b/elements/pfe-content-set/demo/index.html
@@ -367,6 +367,26 @@
     </pfe-content-set>
   </section>
 
+    <section class="example">
+      <h2>Content set with floating content</h2>
+      <pfe-content-set pfe-variant="earth" pfe-breakpoint="2000">
+        <h4 pfe-content-set--header role="heading" >Tab 1</h4>
+        <div pfe-content-set--panel role="region" >
+          <div style="float: left;width:48%;margin-right:2%;border: #eee solid 1px; padding: 20px;">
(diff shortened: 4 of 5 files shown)


elements/pfe-accordion/src/pfe-accordion-panel.html
```diff
@@ -1,5 +1,5 @@
 <div tabindex="-1" role="tabpanel">
-  <div class="container">
+  <div class="container clearfix">
```
castastrophe, line 2: I'm not a fan of helper classes in the context of web components because the code is all built inside the shadow DOM and in that way, the styles can't really be shared.  You can add the mixin to the container class directly.  What would you think of the following template updates:

```
<div class="pfe-accordion-panel__container" tabindex="-1" role="tabpanel">
  <slot></slot>
</div>
```

Honestly, I'm not convinced we even need the 2nd div here, does the layout break if we combine lines 1 and 2?

elements/pfe-tabs/src/pfe-tab-panel.html
```diff
@@ -1,5 +1,5 @@
 <div tabindex="-1" role="tabpanel">
```
castastrophe, line 1: Same suggestion here, no need to use the helper class approach in the templates.

elements/pfe-accordion/src/pfe-accordion-panel.html
starryeyez024: Hmm, i don't know if we need both divs, but that would be out of scope for what this PR is trying to accomplish. I removed the clearfix class so now there are no edits to the template in this PR :)

elements/pfe-accordion/src/pfe-accordion-panel.scss
```diff
@@ -23,6 +23,10 @@ $LOCAL: accordion;
   --pfe-accordion--panel-container--Padding: 0 calc(#{pfe-local(base--Padding)} * 3) #{pfe-local(base--Padding)} calc(#{pfe-local(base--Padding)} * 1.5);
 }
 
+.clearfix {
```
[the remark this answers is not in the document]
starryeyez024: True, though I think that could also be in a separate PR since I'm not editing that variable here.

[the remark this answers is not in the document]
castastrophe: Added an issue: https://github.com/patternfly/patternfly-elements/issues/847

CHANGELOG-prerelease.md
```diff
@@ -3,6 +3,7 @@
 - [](https://github.com/patternfly/patternfly-elements/commit/) fix: pfe-tabs adding dynamic tab in IE11 #838
 - [eb74cb8](https://github.com/patternfly/patternfly-elements/commit/eb74cb8f989048164fbb6ed1508c502659a752ed) feat: add event emission to pfe-select
 - [](https://github.com/patternfly/patternfly-elements/commit/) fix: Add a warning about updating the on attribute before upgrade
```
castastrophe, line 5: @starryeyez024 Generally it's helpful to add the hash link for the item(s) in the log before yours (and the person committing next will do that for your addition too).  Less work during the release.

castastrophe:
Test case updates:
1. View the [accordion page](https://deploy-preview-835--happy-galileo-ea79c4.netlify.app/elements/pfe-accordion/demo/).  In the first section titled **Accordion component**, click on the 2nd accordion item titled "Panel with links & floating cards". Inspect the contents (you should see 2 cards). Note that the blocks are floating, and that there is still padding at the bottom.
2. View the [tabs page](https://deploy-preview-835--happy-galileo-ea79c4.netlify.app/elements/pfe-tabs/demo/). In the **Style: Horizontal plain** section, open the third tab and inspect the contents. Note that the blocks are floating, and that there is still padding at the bottom.
2. View the [content-set page](https://deploy-preview-835--happy-galileo-ea79c4.netlify.app/elements/pfe-content-set/demo/). In the first component, open the third accordion labeled "Heading 3" and inspect the contents. Note that the blocks are floating, and that there is still padding at the bottom.


- [x] Latest version of Edge and Edge 18
- [x] Galaxy S9 Firefox
- [x] iPhone X Safari
- [x] iPad Pro Safari
- [x] Pixel 3 Chrome

PR #843, opened 2020-04-28
feat: schema form renderer

castastrophe:


castastrophe:
TODO needs to be updated with the latest properties

PR #592, opened 2019-10-28
WIP: Updating pfe-card with a few new variable hooks

castastrophe:
## Title of this pull request

* Link to discovery documentation (should contain all the planning and requirements for this work):
   - See the `DISCOVERY.md` template at the root of the project
* Link(s) to demo pages where this element can be viewed:

---

### What has changed and why
Summarize files edited as part of this MR along with a brief description of what was changed/why.

*

### Testing instructions
Be sure to include detailed instructions on how your update can be tested by another developer.

1. 

#### Browser requirements
Your component should work in all of the following environments:

- [ ] Latest 2 versions of Edge
- [ ] Internet Explorer 11 (should be useable, not pixel perfect)
- [ ] Latest 2 versions of Firefox (one on Mac OS, one of Windows OS)
- [ ] Firefox 60.7.2 (or latest version for Red Hat Enterprise Linux distribution)
- [ ] Latest 2 versions of Chrome (one on Mac OS, one of Windows OS)
- [ ] Latest 2 versions of Safari
- [ ] Galaxy S9 Firefox
- [ ] iPhone X Safari
- [ ] iPad Pro Safari
- [ ] Pixel 3 Chrome

Your repository infrastructure updates should work for at least:
- [ ] Node v8.x
- [ ] NPM v7.x

### Ready-for-merge Checklist

- [ ] Expected files: all files in this pull request are related to one feature request or issue (no stragglers)?
- [ ] Did browser testing pass?
- [ ] Did you update or add any necessary documentation (README.md, WHY.md, etc.)?
- [ ] Was this feature demo'd and the design review approved?
- [ ] Did you update the CHANGELOG.md file with a summary of this update?

**Be sure to share your updates with the [patternfly-elements-contribute@redhat.com](mailto:patternfly-elements-contribute@redhat.com) mailing list!**

PR #861, opened 2020-05-06
fix: pfe-card header and footer slot issues

### LINKED ISSUE #860: pfe-card: Issues with header and footer slots
I discovered this while using `pfe-card` in a React app. The header and footer slots to render on page load for some reason. 

To recreate the issue, take a look at this codesandbox:
https://codesandbox.io/s/keen-dream-05nx1?file=/src/App.js

Calling the `render` method on `pfe-card` after it has loaded seems to fix the problem but I don't think that's the right solution. Slots are meant to be empty containers that can be used when they're needed. So if a component receives new content that is meant to be in a particular slot, we shouldn't have to re-render the component to make the content visible.

kylebuch8:
Fixes #860

**Testing Instructions**
- Navigate to the [`pfe-card` demo page](https://deploy-preview-861--happy-galileo-ea79c4.netlify.app/elements/pfe-card/demo/)
- Open the developer tools
- Drop in the following code into the console and execute it
```
const card = document.createElement("pfe-card");
card.textContent = "Test Card";
document.body.appendChild(card);
```
- Make sure the card is visible at the bottom of the page
- Add the following code into the console and execute it
```
const header = document.createElement("h2");
header.setAttribute("slot", "pfe-card--header");
header.textContent = "Card Header";
card.prepend(header);
```
- Validate that the new header shows up in the card.

**Browser testing**
- Tested in Firefox, Safari, Chrome, Edge 80+, Edge 18, and IE11

### DIFF
--- a/CHANGELOG-prerelease.md
+++ b/CHANGELOG-prerelease.md
@@ -1,5 +1,6 @@
 ## Prerelese 46 ( TBD )
 
+- [](https://github.com/patternfly/patternfly-elements/commit/) fix: pfe-card header and footer slot issues
 - [27fee5f](https://github.com/patternfly/patternfly-elements/commit/27fee5f5c5eb021ac126f3767dd0299f5cda8231) fix: pfe-tabs check for tagName on addedNode mutation before continuing
 - [2c950b0](https://github.com/patternfly/patternfly-elements/commit/2c950b08f7638787df50aa5ee6738f1205ea3a9d) fix: Add clearfix within tab and accordion panels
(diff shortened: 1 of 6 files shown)


kylebuch8:
I also did an npm link to a React application on my local computer and validated that the header and footer slots are now rendering properly.

Before the fix
Screen Shot 2020-05-06 at 4 41 57 PM

After the fix
Screen Shot 2020-05-06 at 4 45 57 PM

castastrophe:
@kylebuch8 Dang these are such great updates and test cases but I've had a PR opened that refactors the card styles and this would be seriously hard to merge with it: https://github.com/patternfly/patternfly-elements/pull/592

PR #870, opened 2020-05-12
fix: pfe-navigtation-item shouldn't add duplicative event listeners

### LINKED ISSUE #869: pfe-navgiation: event listeners are added twice to the mobile menu navigation item
For some reason, the event listeners in `_init__handlers` (https://github.com/patternfly/patternfly-elements/blob/master/elements/pfe-navigation/src/pfe-navigation-item.js#L318) are being added twice. 

The mobile menu navigation item seems to be the only navigation item that has this issue. It stems from the `this._init__tray()` call in the `connectedCallback` and then `this._init__tray` is called again when the slotchange event added in the `connectedCallback` fires.

Normally, adding multiple event listeners with the same signature doesn't cause the event listener to be added twice (https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Multiple_identical_event_listeners). However, we've noticed that apps or pages that are using Sentry, multiple event listeners is causing the navigation to fire two click events which causes the mobile navigation menu item to open and then immediately close.

See this [codesandbox](https://codesandbox.io/s/dawn-meadow-1orth) for a demo of the issue.

@eyevana has logged an issue with Sentry, https://github.com/getsentry/sentry-javascript/issues/2461, but I don't think we can wait for a fix from them.

kylebuch8:
Fixes #869

### DIFF
--- a/CHANGELOG-prerelease.md
+++ b/CHANGELOG-prerelease.md
@@ -1,3 +1,7 @@
+## Prerelease 47 ( TBD )
+
+- []() fix: pfe-navigtation-item shouldn't add duplicative event listeners #870
+
 ## Prerelease 46 ( 2020-05-12 )
 
 - [27fee5f](https://github.com/patternfly/patternfly-elements/commit/27fee5f5c5eb021ac126f3767dd0299f5cda8231) fix: pfe-tabs check for tagName on addedNode mutation before continuing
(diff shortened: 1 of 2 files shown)


Djfaucette:
Lying graph tells mischief

PR #885, opened 2020-05-22
test: visual regression testing

kylebuch8:


test/vrt/index.js
```diff
@@ -0,0 +1,13 @@
+describe("PatternFly Elements", () => {
```
castastrophe, line 1: @kylebuch8 Can we set up time to share how webrh architects this?  I think it would be good to have test files for each component.

wdio.conf.js
```diff
@@ -0,0 +1,37 @@
+const { join } = require("path");
```
castastrophe, line 1: This is awesome!

PR #906, opened 2020-06-03
fix: Pfe jump links

Djfaucette:
I fixed the `offset` attribute to actually change. It was kind of working before, but not reliably and I hunted down the bug. Now you can adjust the distance from the top of the window where the link "flips on".


starryeyez024:
@Djfaucette any reason this was closed instead of merging?

Issue #923, opened 2020-06-15
pfe-modal | Should remove hidden attribute from host on open


castastrophe:
The pfe-modal needs to remove any hidden attributes from the host when the open function is fired. 
 `this.removeAttribute("hidden");`

https://github.com/patternfly/patternfly-elements/blob/6919a2c9283551d9e25d60ff96d497fab016e270/elements/pfe-modal/src/pfe-modal.js#L161

PR #924, opened 2020-06-15
fix: pfe-modal removing hidden attribute

LyndseyR:
https://github.com/patternfly/patternfly-elements/issues/923

**Testing instructions**
- Test case was added to AT to verify functionality


elements/pfe-modal/src/pfe-modal.js
```diff
@@ -159,6 +159,7 @@ class PfeModal extends PFElement {
 
     this.isOpen = true;
     // Reveal the container and overlay
+    this.removeAttribute("hidden");
```
kylebuch8, line 162: @LyndseyR - Do we want to add a hidden attribute back to the modal light DOM when the modal is closed?

castastrophe: I was pondering this - the challenge is, some modals (maybe most) will have the link that triggers the modal inside a slot so we don't _want_ hidden on the host.  We'd need some flag that says it had hidden on instantiation or something.

elements/pfe-modal/test/pfe-modal_test.html
```diff
@@ -51,6 +59,14 @@ <h2 slot="pfe-modal--header">Modal with a header</h2>
           assert.isTrue(modalWindow.hasAttribute('hidden'));
         
         });
+        
+        test('it should remove the hidden attribute from the host on open', done => {
+          // get the modal and fire the open function
```
castastrophe, line 64: @LyndseyR You'll need to replace the comments I put there with the code to do those things

LyndseyR: @castastrophe I added a test that when the trigger was clicked the modal window appeared and the pfe-modal hidden attribute wasn't appearing. Is this enough?

castastrophe: I love it!  That's perfect.  I also really like that you included the shadow DOM test on the pfe-modal__window. Thank you so much!

elements/pfe-modal/src/pfe-modal.js
castastrophe: How important do you think adding it back is?  The content is hidden in the Shadow DOM, so it's only relevant for before the component upgrades.

elements/pfe-modal/src/pfe-modal.js
```diff
@@ -96,6 +96,7 @@ class PfeModal extends PFElement {
 
     if (this.trigger) {
       this.trigger.addEventListener("click", this.open);
+      this.removeAttribute("hidden");
```
castastrophe, line 99: @kylebuch8 Since this is a progressive enhancement feature, I suggested we remove the hidden attribute in the init function instead of the on click event.  Once the component upgrades, the shadow DOM manages the hidden nature of the modal panel and we don't want to risk hiding the trigger if it's used.  The hidden attribute on host would just be there to prevent a flash of unstyled content or to hide the content should JS fail (guidance would be to leave off the hidden attribute if content is critical to the page).

elements/pfe-modal/test/pfe-modal_test.html
```diff
@@ -51,7 +62,22 @@ <h2 slot="pfe-modal--header">Modal with a header</h2>
           assert.isTrue(modalWindow.hasAttribute('hidden'));
         
         });
-
+        
+        test('it should remove the hidden attribute from the host on open', done => {
```
castastrophe, line 66:
```suggestion
        test('it should remove the hidden attribute from the host on upgrade', done => {
```

elements/pfe-modal/test/pfe-modal_test.html
```diff
@@ -51,7 +62,22 @@ <h2 slot="pfe-modal--header">Modal with a header</h2>
...
         
         });
-
+        
+        test('it should remove the hidden attribute from the host on upgrade', done => {
+          const pfeModal = document.querySelector('pfe-modal');
+          const modalWindow = pfeModal.shadowRoot.querySelector('.pfe-modal__window');
+          const trigger = pfeModal.querySelector('[slot=pfe-modal--trigger');
+          
+          // get the modal and fire the open function
```
castastrophe, line 71: @LyndseyR This comment might need to be updated since we're not firing the o pen function but triggering a click event instead.

elements/pfe-modal/test/pfe-modal_test.html
```diff
@@ -51,7 +62,22 @@ <h2 slot="pfe-modal--header">Modal with a header</h2>
...
+          trigger.click();
+          
+          flush(() => {
+            // test for the hidden attribute on the host
```
castastrophe, line 75: Update the comment to note that we're testing for the hidden attribute in the shadow DOM too

elements/pfe-modal/test/pfe-modal_test.html
```diff
@@ -1,59 +1,91 @@
 <!DOCTYPE html>
 <html>
```
castastrophe, line 2: It looks like this file wasn't running prettier so when I made an update, it is showing a lot more diffs

PR #925, opened 2020-06-15
docs: Fix misspelling of src attribute

wesruv:


castastrophe:
@kylebuch8 It looks like the Travis CI passed but is having a difficult time reporting that back to PRs occasionally.  I've been seeing this from time-to-time in the last couple of weeks.

PR #930, opened 2020-06-16
fix: fixed a edge case where element wasn't being created in mobile accordion, closes #927

### LINKED ISSUE #927: pfe-jump-links | Mobile accordion header missing
## Description of the issue

On the master branch for pfe-jump-links, when viewing the component:
- In Chrome
- Between 0-991px
- On the demo page

The jump link header does not show up in the accordion header.

Screen Shot 2020-06-16 at 11 53 25 AM

### Impacted component(s)

- pfe-jump-links

### Steps to reproduce

1. Go to the jump links demo page on the master branch: http://localhost:8001/elements/pfe-jump-links/demo/
2. Adjust browser size to 991px or below

### Expected behavior

Expect an h3 to be injected into the accordion header with "Jump to section".

### Screenshots

See above

Djfaucette:
Just made a simple mistake when reading the light dom. Should be all fixed now.

### DIFF
--- a/elements/pfe-jump-links/CHANGELOG.md
+++ b/elements/pfe-jump-links/CHANGELOG.md
@@ -1,5 +1,5 @@
-## 1.0.0-prerelease.39 ( TBD )
+## 1.0.0-prerelease.50 ( TBD )
 
-Tag: [1.0.0-prerelease.39](https://github.com/patternfly/patternfly-elements/releases/tag/1.0.0-prerelease.39)
+Tag: [1.0.0-prerelease.50](https://github.com/patternfly/patternfly-elements/releases/tag/1.0.0-prerelease.39)
 
-- [code](url) Init - new component pfe-jump-links
+- [code](url) Fix - fixed demo file to accurately show the working component
--- a/elements/pfe-jump-links/src/pfe-jump-links.js
+++ b/elements/pfe-jump-links/src/pfe-jump-links.js
@@ -97,7 +97,7 @@ class PfeJumpLinksNav extends PFElement {
           return;
         }
         let html = "";
-        if ("[slot='pfe-jump-links-nav--heading']") {
+        if (this.has_slot("pfe-jump-links-nav--heading")) {
           html = this.querySelector("[slot='pfe-jump-links-nav--heading']")
             .innerHTML;
         } else {
@@ -116,7 +116,7 @@ class PfeJumpLinksNav extends PFElement {
             );
           this.shadowRoot.querySelector(
             "#pfe-jump-links-nav--heading"
-          ).innerHTML = "Jump to section hi";
+          ).innerHTML = "Jump to section";
         }
       }
     }
--- a/elements/pfe-jump-links/src/pfe-jump-links.js
+++ b/elements/pfe-jump-links/src/pfe-jump-links.js
@@ -93,18 +93,20 @@ class PfeJumpLinksNav extends PFElement {
         if (this.getAttribute("sr-text")) {
           this.shadowRoot.querySelector("nav").prepend(div);
         }
-
+        if (this.hasAttribute("horizontal")) {
+          return;
+        }
         let html = "";
-        if (this.querySelector(".pfe-jump-links-nav--heading")) {
-          html = this.querySelector(".pfe-jump-links-nav--heading").cloneNode(
-            true
-          );
+        if ("[slot='pfe-jump-links-nav--heading']") {
+          html = this.querySelector("[slot='pfe-jump-links-nav--heading']")
+            .innerHTML;
+        } else {
+          html = "Jump to section";
         }
         if (!this.hasAttribute("horizontal")) {
           if (html) {
-            this.shadowRoot
-              .querySelector("pfe-accordion-header")
-              .appendChild(html);
+            let header = this.shadowRoot.querySelector("pfe-accordion-header");
+            header.shadowRoot.querySelector("button").innerHTML = html;
           }
         } else {
           this.shadowRoot
@@ -114,7 +116,7 @@ class PfeJumpLinksNav extends PFElement {
             );
           this.shadowRoot.querySelector(
             "#pfe-jump-links-nav--heading"
-          ).innerHTML = "Jump to section";
+          ).innerHTML = "Jump to section hi";
         }
       }
     }
@@ -203,8 +205,11 @@ class PfeJumpLinksNav extends PFElement {
         ${buildLinkList()}
     `;
     this.shadowRoot.querySelector("#container").innerHTML = html;
+
     let heading = document.createElement("h3");
-    heading.innerHTML = "Jump to section";
+    heading.innerHTML =
+      this.querySelector("[slot='pfe-jump-links-nav--heading']").innerHTML ||
+      "Jump to section";
     this.shadowRoot.querySelector("pfe-accordion-header").appendChild(heading);
   }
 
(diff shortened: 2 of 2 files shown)


elements/pfe-jump-links/CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
-## 1.0.0-prerelease.39 ( TBD )
+## 1.0.0-prerelease.50 ( TBD )
 
-Tag: [1.0.0-prerelease.39](https://github.com/patternfly/patternfly-elements/releases/tag/1.0.0-prerelease.39)
+Tag: [1.0.0-prerelease.50](https://github.com/patternfly/patternfly-elements/releases/tag/1.0.0-prerelease.39)
 
-- [code](url) Init - new component pfe-jump-links
\ No newline at end of file
+- [code](url) Fix - fixed demo file to accurately show the working component
```
castastrophe, line 5: I love that you updated the component changelog!!  So cool!  We also need to add this to the monorepo changelog for now until I can automate it.

elements/pfe-jump-links/src/pfe-jump-links.js
```diff
@@ -93,18 +93,20 @@ class PfeJumpLinksNav extends PFElement {
...
-
+        if (this.hasAttribute("horizontal")) {
+          return;
+        }
         let html = "";
-        if (this.querySelector(".pfe-jump-links-nav--heading")) {
-          html = this.querySelector(".pfe-jump-links-nav--heading").cloneNode(
-            true
-          );
+        if ("[slot='pfe-jump-links-nav--heading']") {
```
castastrophe, line 100:
```suggestion
        if (this.has_slot("pfe-jump-links-nav--heading")) {
```

There's a built-in function in pfelement that can verify if a slot exists.  Alternatively, I think this needs a query selector instead of just the quotes.

elements/pfe-jump-links/src/pfe-jump-links.js
```diff
@@ -93,18 +93,20 @@ class PfeJumpLinksNav extends PFElement {
...
+          html = this.querySelector("[slot='pfe-jump-links-nav--heading']")
+            .innerHTML;
+        } else {
+          html = "Jump to section";
         }
         if (!this.hasAttribute("horizontal")) {
```
castastrophe, line 106: Do you still need this since you added the escape on line 96?

elements/pfe-jump-links/src/pfe-jump-links.js
```diff
@@ -93,18 +93,20 @@ class PfeJumpLinksNav extends PFElement {
         if (this.getAttribute("sr-text")) {
           this.shadowRoot.querySelector("nav").prepend(div);
         }
-
+        if (this.hasAttribute("horizontal")) {
+          return;
+        }
         let html = "";
```
castastrophe, line 99: If you set this up as:

```suggestion
        let html = "Jump to section";
```

You don't need the else statement and you can drop the `if(html)` query below. on line 107

elements/pfe-jump-links/src/pfe-jump-links.js
```diff
@@ -114,7 +116,7 @@ class PfeJumpLinksNav extends PFElement {
             );
           this.shadowRoot.querySelector(
             "#pfe-jump-links-nav--heading"
-          ).innerHTML = "Jump to section";
+          ).innerHTML = "Jump to section hi";
```
castastrophe, line 119: Do you still need this else statement here since we know HTML will exist above?

elements/pfe-jump-links/src/pfe-jump-links.js
```diff
@@ -203,8 +205,11 @@ class PfeJumpLinksNav extends PFElement {
         ${buildLinkList()}
     `;
     this.shadowRoot.querySelector("#container").innerHTML = html;
+
     let heading = document.createElement("h3");
-    heading.innerHTML = "Jump to section";
+    heading.innerHTML =
```
castastrophe, line 210: Is this separate from the code on line 100?  It looks similar.

elements/pfe-jump-links/src/pfe-jump-links.js
Djfaucette: You're right, good catch. I wonder why this worked though? Weird.

castastrophe: I suspect it's always true.  Maybe we need to add an AT that validates that the default text is used if the heading slot is empty?

elements/pfe-jump-links/src/pfe-jump-links.js
Djfaucette: Yes, I think so. No harm in leaving it because it would handle any edge case like no slot or an undefined value.

elements/pfe-jump-links/src/pfe-jump-links.js
Djfaucette: Yep. One adresses the heading (shown in tablet+) and the other makes sure there is innerHTML in the button inside accordion.

elements/pfe-jump-links/src/pfe-jump-links.js
Djfaucette: Yeah, this is where we set the html in that heading so I think it needs to stay.

elements/pfe-jump-links/src/pfe-jump-links.js
castastrophe: I don't think it will.  It escapes completely on line 96 so you would never hit this point if the horizontal attribute exists.

elements/pfe-jump-links/src/pfe-jump-links.js
castastrophe: If you make the update I suggest on line 96 then you don't need to duplicate this.  It would be good to have a single point in the file where the fallback text is defined so that if it's updated, it will be done in one spot instead of having to find all the references.  Also, I suspect you don't want this postfixed with `hi` at the end.

elements/pfe-jump-links/src/pfe-jump-links.js
castastrophe: See my note above about having a "source of truth" for this `Jump to section` content.

elements/pfe-jump-links/src/pfe-jump-links.js
Djfaucette: I think we do need in case there is no heading slot and it has the horizontal attribute. Both of the previous ifs would be false, and I think that text should still be there for assistive technology. Wdyt?

castastrophe: It can't have the horizontal attribute because you exited the function on line 97 if that attribute existed.  If you initialize the html variable to the text and then reassign it if the slot is used, then you get it without needing the else on line 103.

castastrophe: Let me know if you want to sync up about this - we can jump on a call.

elements/pfe-jump-links/src/pfe-jump-links.js
```diff
@@ -93,18 +93,20 @@ class PfeJumpLinksNav extends PFElement {
...
-            true
-          );
+        if (this.has_slot("pfe-jump-links-nav--heading")) {
+          html = this.querySelector("[slot='pfe-jump-links-nav--heading']")
+            .innerHTML;
+        } else {
+          html = "Jump to section";
         }
         if (!this.hasAttribute("horizontal")) {
           if (html) {
```
castastrophe, line 107: I don't see how this can ever be false since html is initialized on line 99-104 above

elements/pfe-jump-links/src/pfe-jump-links.js
Djfaucette: Ah, I get it. I was writing code thinking html would coerce to false, but setting it to a string initially removes that need.

PR #998, opened 2020-07-21
Cpfed 3743 main nav styles

kelsS:
---
name: CPFED-3743 Main Nav Styles
labels: ready for review
---

## pfe-navigation

Updated focus styles for main menu and dropdowns, cleaned up main nav styles and double checked colors and focus/hover states with design and redhat.com example.

### Testing instructions

1. Using a keyboard, tab through the menu to see the new focus states.

### Merging

### DIFF
--- a/elements/pfe-navigation/src/pfe-navigation.scss
+++ b/elements/pfe-navigation/src/pfe-navigation.scss
@@ -47,6 +47,11 @@
 
   a {
     text-decoration: none;
+    // @todo: move these styles to be more specific with dropdown class so that logo link can be styled without conflicting
+    // &:focus,
+    // &:hover {
+    //   @include focus-styles(#000, 1px);
+    // }
   }
 }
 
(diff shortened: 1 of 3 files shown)


wesruv:
Looking at the code, looks great! I'll test it after I get all these conflicts handled :sweat:

kelsS:
@wesruv sounds good, let me know if there is anything I can help with!

elements/pfe-navigation/src/pfe-navigation.scss
```diff
@@ -329,8 +356,8 @@
       }
     }
     
-    &:hover,
-    &:focus {
+    &:focus,
```
wesruv, line 359: @kelsS so I'm getting that this is an intentional order, you know something I don't?

kelsS: @wesruv Oh yeah, read something about the focus/hover CSS rule order on MDN a while back and just started doing it the way the docs said it should be done and now it is a habit. I think it is discussed in this [MDN article](https://developer.mozilla.org/en-US/docs/Learn/CSS/Styling_text/Styling_links), not sure how much issue it will actually cause though in a different order in practice. The article is vague about the issues it will cause, it just says "If you put these in the wrong order, things won't work properly." lol Let me know your thoughts on that article.

wesruv: Muchos dankes

PR #1043, opened 2020-08-13
Cpfed 3789 update aria util btns

kelsS:
---
name: Cpfed 3789 update aria features for shadow DOM buttons and dropdown content
labels: "ready for review"
---

## pfe-navigation

added aria-hidden/aria-expanded to shadow DOM for the other dropdowns, added some notes on things to regroup on, did some more keyboard testing and fixed aria-hidden issue for search and all red hat buttons but not dynamically, updated alt text of images, removed aria-expanded and has-dropdown from no js menu example.

### Merging

### DIFF
--- a/elements/pfe-navigation/src/pfe-navigation.js
+++ b/elements/pfe-navigation/src/pfe-navigation.js
@@ -354,12 +354,13 @@ class PfeNavigation extends PFElement {
         dropdownElement.style.removeProperty("height");
       }
       // Delay aria-hidden so animations can finish
-      // @todo/note: I do not think we need to delay the change of the aria-hidden state since aria-hidden hides the content of the page from screen readers, as soon as the dropdown is closed by the user the dropdown should get aria-hidden="true" immediately
+      // @todo/note: I do not think we need to delay the change of the aria-hidden state since aria-hidden hides the content of the page from screen readers, as soon as the dropdown is closed by the screen reader user the dropdown should get aria-hidden="true" immediately
       // @note: tested with a screen reader and things seem to be functioning well without the time delay.
-      window.setTimeout(() => {
-        dropdownElement.setAttribute("aria-hidden", "true");
-      }, 500);
-      // dropdownElement.setAttribute("aria-hidden", "true");
+      // @note: when the delay is on it seems to be causing issues with the toggling of the aria-hidden states sometimes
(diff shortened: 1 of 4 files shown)


elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -354,9 +354,13 @@ class PfeNavigation extends PFElement {
         dropdownElement.style.removeProperty("height");
       }
       // Delay aria-hidden so animations can finish
-      window.setTimeout(() => {
-        dropdownElement.setAttribute("aria-hidden", "true");
-      }, 500);
+      // @todo/note: I do not think we need to delay the change of the aria-hidden state since aria-hidden hides the content of the page from screen readers, as soon as the dropdown is closed by the screen reader user the dropdown should get aria-hidden="true" immediately
```
wesruv, line 357: I did this because I'm tying the CSS style `visibility:hidden` to it with an attribute selector, but that's probably not a great way to do that.

I'll use a different method to set visibility that we can delay.

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -410,6 +414,7 @@ class PfeNavigation extends PFElement {
    * Clone them into the shadowRoot
    * @param {array} mutationList Provided by mutation observer
    */
+  // @todo: processLightDom seems to be firing twice, need to look into this and see whether that is okay or if it needs to be fixed, seems like it is not a good thing but not sure if it is avoidable or not
```
wesruv, line 417: Is it once per pfe-navigation on the page? Or 2x per pfe-navigation on the page?

elements/pfe-navigation/src/pfe-navigation.js
kelsS: @wesruv oh okay gotcha, let me know how you end up doing it differently.

elements/pfe-navigation/src/pfe-navigation.js
kelsS: @wesruv I checked it with a `console.count()` and it seemed like it might be 2x per pfe-nav, however I think we need to double check that with only one pfe-nav on the page because I might have just misread the console output. Sorry! didn't think about checking the count with only one pfe-nav on the page when I was testing.

Issue #933, opened 2020-06-17
Update attribute prefixing


kylebuch8:
# All components

### Global
pfe-g-\<name\>

#### Globally-scoped attributes
These would be protected and components would not be allowed to use the same attribute name. i.e., `pfe-g-context="dark"` (system adds on="dark" < styles queue off of this), `pfe-g-type="container"`

Should still avoid reusing these concepts on the component-level to mean something else; could result in conflicting styles or confusing author experience

#### component name (minus the pfe prefix)
Locally-scoped attributes; these are for use inside the component to communicate settings or content. i.e., card-color="accent" or icon-name="rh-atom"
BEM naming? c-color, c-name="rh-atom"
pfe-c-<name>: component-specific

castastrophe:
Goal: Set up a structured method for defining attributes on a component.

How LitElement is doing this: https://lit-element.polymer-project.org/guide/properties

```
static get properties() {
  return { 
    propertyName: options
  };
}
```

What would go in our options?
- CMS-readable name or "title"
- Attribute name (should prefixing be built-in automatically so this value would be `disabled` and upgrade to `pfe-c-disabled`)?
- Observed boolean?

kylebuch8:
In progress Codesandbox: https://codesandbox.io/s/pfe-constructor-properties-dwiwe

markcaron:
Not sure if this is part of this specific issue, but the documentation site (https://patternfly.github.io/patternfly-elements/theme/) needs to be updated to match too. Still says `pfe-color`, etc.

PR #1091, opened 2020-09-10
feat: Add standard attribute/property definitions

mwcz:
### Components updated

- [x] pfe-accordion
- [x] pfe-autocomplete
- [x] pfe-avatar
- [x] pfe-badge
- [x] pfe-band
- [x] pfe-button
- [x] pfe-card
- [x] pfe-collapse
- [x] pfe-content-set
- [x] pfe-cta
- [x] pfe-datetime
- [x] pfe-dropdown
- [x] pfe-health-index
- [x] pfe-icon
- [x] pfe-icon-panel
- [x] pfe-jump-links
- [x] pfe-markdown
- [x] pfe-modal
- [ ] pfe-navigation
- [x] pfe-number
- [x] pfe-page-status
- [x] pfe-progress-indicator
- [x] pfe-select
- [x] pfe-tabs
- [x] pfe-toast

### Related issue

 - #933 Update attribute prefixing

### TODO

- [x] add support for `static get properties()` on the base class (note to self: use `Object.getPrototypeOf(this.constructor).properties`)

### Preview

- [Preview](https://deploy-preview-1091--happy-galileo-ea79c4.netlify.app/examples/) 

### What has changed and why

Property definitions provide a standard, streamlined way to declare the properties and attributes your component needs.  Property definitions provide the following features:

 - Two-way binding between property and attribute values.
 - Default value initialization.
 - Data types (for properties).
 - Observer function to be run when a value changes.
 - Optional namespacing prefix for attribute names.
 - Aliasing properties to one another, for phased deprecation.
 - And more!

[More information](https://github.com/patternfly/patternfly-elements/pull/1091/files#diff-07efd2bdc7cd3054bb50a5dcb1402457)

### Testing instructions

1. Regression testing, no visible changes to components

#### Browser requirements

- [ ] Firefox 68 (or latest version for Red Hat Enterprise Linux distribution)
- [ ] Android mobile device (such as the Galaxy S9)
- [ ] Apple mobile device (such as the iPhone X)
- [ ] Apple tablet device (such as the iPhone Pro)

### Ready-for-merge Checklist

Check off items as they are completed.  Feel free to delete items if they are not applicable.

- [ ] Browser testing passed.
- [x] Changelog updated (not needed for documentation updates).
- [x] Documentation (README.md, WHY.md, etc.) updated or added.

### Merging


docs/content/develop/properties.md
```diff
@@ -0,0 +1,107 @@
...
+
+Below, all the options for property definitions are explained.
+
+## Property / attribute binding
+
+As mentioned in the introduction, defining a property also creates an attribute, and the two values will be kept in sync.  For properties with the String type, this is straightforward.  Attribute values are always strings, and so logic is needed to 
+
+### Handling `null`
+
+If a property is assigned to null, the associated attribute will be removed.
```
castastrophe, line 42: If you leave off the `default` setting, is it setting the property to `null`?

docs/content/develop/properties.md
```diff
@@ -0,0 +1,107 @@
...
+
+Example: `this.count = null` will result in `this.removeAttribute("pfe-c-count")`.
+
+## Type
+
+Three property types are supported: String, Number, and Boolean.
```
castastrophe, line 48:
```suggestion
Three property types are currently supported: String, Number, and Boolean.
```

^ Since this may evolve or expand in the future

docs/content/develop/properties.md
```diff
@@ -0,0 +1,107 @@
...
+### String
+
+```
+name: {
+  type: String
+}
+```
+
+### Boolean
+
```
castastrophe, line 59:
```suggestion
Though all attribute values are technically strings, if you are assigning a `true/ false` paradigm or checking that a variable exists (implicit `true/false`, i.e., `hidden` or `required`), please assign that property a type of `Boolean`.  This will ensure the update and injection of that value is handled correctly by the base class.
```

docs/content/develop/properties.md
```diff
@@ -0,0 +1,107 @@
...
+### Number
+
```
castastrophe, line 61:
```suggestion
Similar to boolean, though attribute values are strings, if you are assigning a numerical value to your attribute, please use this type for your property.  Consider the use-case of `tabindex` for example.  This will output a quoted string when assigned to the attribute but use a numerical type when you query for the property.
```

Is this an int or does this support floats?

docs/content/develop/properties.md
```diff
@@ -0,0 +1,107 @@
...
+Example: `this.count = NaN` (or `Number.NaN`) will result in `pfe-c-count="NaN"`.
+
+## Default
+
+During the component's `connectedCallback`, values specified as defaults will be initialized in both properties and attributes.
```
castastrophe, line 66:
```suggestion
During the component's `connectedCallback`, values with defined defaults will be initialized in both properties and attributes.
```

docs/content/develop/properties.md
```diff
@@ -0,0 +1,107 @@
...
+
+Example: since `count` has a default of `0`, when the component is connected, it will have `this.count === 0` and `this.getAttribute("pfe-c-count") === "0"`.
```
castastrophe, line 68: Unless the user has `pfe-c-count=3`?

docs/content/develop/properties.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Observer
+
+You may choose to provide an `observer` for any property, which is the string name of a (non-static) function which will be called whenever the property changes.  Here's an example wherein changes to the `count` will trigger the `handleCount` function.
```
castastrophe, line 72:
```suggestion
You may provide an `observer` for any property, which is the string name of a (non-static) function inside the web component which will be called whenever the property changes.  Here's an example wherein changes to `count` will trigger the `handleCount` function.
```

docs/content/develop/properties.md
```diff
@@ -0,0 +1,107 @@
...
+  handleCount(oldVal, newVal) {
+    console.log(`count changed from ${oldVal} to ${newVal}`);
+  }
+
+}
+```
+
+Observer functions are called with arguments `(oldVal, newVal)`.  Like property values, both arguments will be cast to the appropriate type, ie String, Number, or Boolean.
+
+Note, `observer` is driven by the `attributeChangedCallback` on the `PFElement` class, so if you choose to provide your own `attributeChangedCallback`, you _must_ call `super.attributeChangedCallback(attr, oldVal, newVal)` within it.  The best practice is to not provide an `attributeChangedCallback`.
```
castastrophe, line 97: Are there any other vanilla web component functions that are overridden when using this property definition approach?  `attributeChangedCallback`...`observedAttributes`?  And should we mention the deprecation of `cascadingAttributes`?

castastrophe: What about getters and setters?  Should we mention that too?

docs/content/develop/properties.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Introduction
+
+To introduce the topic, consider a "Counter" component that counts how many times something has happened.  Such a component could define a counter like this.
+
+```javascript
+static get properties() {
+  return {
+    count: {
+      type: Number,
```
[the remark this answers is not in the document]
castastrophe: Can you add cascade definition too?

elements/pfe-accordion/src/pfe-accordion.js
```diff
@@ -26,9 +26,39 @@ class PfeAccordion extends PFElement {
...
 
-  static get cascadingAttributes() {
+  static get properties() {
     return {
-      on: "pfe-accordion-header, pfe-accordion-panel"
+      color: {
+        title: "Color",
+        type: String
+      },
+      pfeColor: {
```
castastrophe, line 35: Love this syntax!

elements/pfe-accordion/src/pfe-accordion.js
```diff
@@ -26,9 +26,39 @@ class PfeAccordion extends PFElement {
...
+        type: String,
+        enum: ["true", "false"],
+        cascade: ["pfe-accordion-header", "pfe-accordion-panel"],
+        observer: "_disclosureChanged"
+      },
+      pfeDisclosure: {
+        // @TODO: Deprecate property
+        type: String,
+        prefix: false,
+        alias: "disclosure"
```
castastrophe, line 52: We should document this too - does it get the name of the property or the name of the attribute?

elements/pfe-accordion/src/pfe-accordion.js
```diff
@@ -26,9 +26,39 @@ class PfeAccordion extends PFElement {
...
+      },
+      on: {
```
castastrophe, line 54: Yes, please don't include this in the accordion definitions since it will be included in the base class' properties.

elements/pfe-accordion/src/pfe-accordion.js
```diff
@@ -26,9 +26,39 @@ class PfeAccordion extends PFElement {
...
+        title: "Color",
+        type: String
+      },
+      pfeColor: {
+        // @TODO: Deprecate property
+        type: String,
+        prefix: false,
+        alias: "color"
+      },
+      disclosure: {
```
castastrophe, line 41: I don't see `aria-expanded` in the property definition list

elements/pfe-accordion/src/pfe-accordion.js
```diff
@@ -26,9 +26,39 @@ class PfeAccordion extends PFElement {
...
+        title: "Disclosure",
+        type: String,
```
castastrophe, line 43: Wouldn't this be boolean?

elements/pfe-autocomplete/src/pfe-autocomplete.js
```diff
@@ -27,6 +27,74 @@ class PfeAutocomplete extends PFElement {
...
   }
 
+  static get properties() {
+    return {
+      initValue: {
+        title: "Initial Value",
+        type: String,
+        observer: "_initValueChanged"
+      },
+      _initValue: {
```
castastrophe, line 37: I like this underscore approach a lot for highlighting attributes to be deprecated.  Can we do that in accordion too and explicitly define the attr instead?

elements/pfe-autocomplete/src/pfe-autocomplete.js
```diff
@@ -27,6 +27,74 @@ class PfeAutocomplete extends PFElement {
...
+        // @TODO Deprecate property
+        type: String,
+        prefix: false,
+        attr: "init-value",
```
castastrophe, line 41: Is this in the readme about what options are available in the property defintions?

elements/pfe-autocomplete/src/pfe-autocomplete.js
```diff
@@ -411,6 +429,44 @@ class PfeSearchDroplist extends PFElement {
     return "pfe-search-droplist.scss";
   }
 
+  static get properties() {
```
castastrophe, line 432: Can we go ahead and pull out PfeSearchDroplist into its own file?  The compiler is updated to handle this now.

elements/pfe-autocomplete/src/pfe-search-droplist.scss
```diff
@@ -8,7 +8,8 @@
   line-height: pfe-var(line-height);
 }
 
-:host([open]) {
+:host([open]),
+:host([pfe-c-open]) {
```
castastrophe, line 12: Since open is alias'ed, I'm not sure we need both in the CSS.  Probably can just use `pfe-c-open`.

elements/pfe-avatar/src/pfe-avatar.js
```diff
@@ -7,6 +7,48 @@ class PfeAvatar extends PFElement {
     return "pfe-avatar";
   }
 
+  static get properties() {
+    return {
+      name: {
+        title: "Username",
+        type: String,
+        observer: "_updateWhenConnected",
+        default: "",
```
castastrophe, line 16: If the default is empty, does the property need to be set?

elements/pfe-avatar/src/pfe-avatar.js
```diff
@@ -7,6 +7,48 @@ class PfeAvatar extends PFElement {
...
+        alias: "pfeName"
+      },
+      src: {
+        title: "Avatar image URL",
+        type: String,
+        observer: "_updateWhenConnected",
+        alias: "pfeSrc"
+      },
+      pattern: {
+        title: "Shape Pattern",
```
castastrophe, line 26:
```suggestion
        title: "Shape pattern",
```

I mean this is super nit-picky but titles are sentence case in the writing for Red Hat guidelines 😉

elements/pfe-avatar/src/pfe-avatar.js
```diff
@@ -15,9 +57,9 @@ class PfeAvatar extends PFElement {
     return "pfe-avatar.scss";
   }
 
-  static get observedAttributes() {
-    return ["pfe-name", "pfe-pattern", "pfe-src", "pfe-shape"];
-  }
+  // static get observedAttributes() {
```
castastrophe, line 60: Can delete once it's working and you don't need it anymore 🎉

elements/pfe-avatar/src/pfe-avatar.js
```diff
@@ -155,9 +204,17 @@ class PfeAvatar extends PFElement {
...
+      this.update();
+    } else {
+      this.addEventListener(PfeAvatar.events.connected, () => this.update());
+    }
+  }
+
   update() {
     // if we have a src element, update the img, otherwise update the random pattern
-    if (this.hasAttribute("pfe-src")) {
+    if (this.src) {
```
castastrophe, line 217: Such a nice syntax!

elements/pfe-badge/README.md
```diff
@@ -4,33 +4,33 @@ The badge component provides a way to have small numerical descriptors for UI el
 
 ```html
 <section>
-    <pfe-badge aria-label="2 unread messages" number="2" pfe-threshold="10">2</pfe-badge>
+    <pfe-badge aria-label="2 unread messages" pfe-c-pfe-c-number="2" pfe-c-threshold="10">2</pfe-badge>
```
castastrophe, line 7:
```suggestion
    <pfe-badge aria-label="2 unread messages" pfe-c-number="2" pfe-c-threshold="10">2</pfe-badge>
```

elements/pfe-badge/src/pfe-badge.js
```diff
@@ -14,15 +14,48 @@ class PfeBadge extends PFElement {
   }
 
   get schemaUrl() {
-    return "pfe-badge.json";
+    // return "pfe-badge.json";
```
castastrophe, line 17: Want to just delete the `schemaUrl` entirely?

elements/pfe-card/demo/index.html
```diff
@@ -162,7 +162,7 @@ <h2>Imaged card with no overflow</h2>
       </pfe-card>
 
       <pfe-card pfe-color="lightest">
```
castastrophe, line 164: Should it be `pfe-c-color`?

elements/pfe-card/src/pfe-card.js
```diff
@@ -28,8 +28,44 @@ class PfeCard extends PFElement {
...
+  static get properties() {
+    return {
+      color: {
+        title: "Background Color",
+        type: String,
+        enum: ["lightest", "base", "darker", "darkest", "complement", "accent"],
+        // default: "base",
+        observer: "_colorChanged"
+      },
+      pfeColor: {
```
castastrophe, line 40: Can we add the todo here too?

elements/pfe-card/src/pfe-card.js
```diff
@@ -59,32 +95,15 @@ class PfeCard extends PFElement {
...
-  _basicAttributeChanged(attr, oldValue, newValue) {
-    this[attr].value = newValue;
-  }
-
   // Update the color attribute and contexts
-  _colorChanged(attr, oldValue, newValue) {
-    this[attr].value = newValue;
+  _colorChanged(oldValue, newValue) {
     // Trigger an update in nested components
+    console.log("color changed");
```
castastrophe, line 101: Can remove console when debugging is done

elements/pfelement/src/pfelement.js
```diff
@@ -1,11 +1,31 @@
 import { autoReveal } from "./reveal.js";
-const prefix = "pfe-";
+import { isAllowedType, isValidDefaultType } from "./attrDefValidators.js";
```
castastrophe, line 2: Love this approach, very nice.

elements/pfelement/src/pfelement.js
```diff
@@ -1,11 +1,31 @@
...
+
+const tagPrefix = "pfe-";
```
castastrophe, line 4:
```suggestion
const prefix = "pfe-";
```

I think prefix is a good name here because it's still the prefix for tags and attributes; we're just additionally adding scope information to make it `${prefix}-${scope}-${name}` 🤷

elements/pfelement/src/pfelement.js
```diff
@@ -1,11 +1,31 @@
...
 
+  /**
+   * This prefix is prepended to the name of any attributes defined on the base
+   * class.  In other words, attributes that can be used on any element will be
+   * prefixed with this string.
+   *
+   * @example A "theme" prop, when reflected as an attribute, prefixed with "pfe-g-": <pfe-foo pfe-g-theme="bar">
+   */
+  static get globalAttrPrefix() {
+    return "pfe-g-";
```
castastrophe, line 20:
```suggestion
    return `${prefix}-g-`;
```

elements/pfelement/src/pfelement.js
```diff
@@ -1,11 +1,31 @@
...
+  }
+
+  static get attrPrefix() {
+    return "pfe-c-";
```
castastrophe, line 24:
```suggestion
    return `${prefix}-c-`;
```

elements/pfelement/src/pfelement.js
```diff
@@ -1,11 +1,31 @@
...
+  }
+
+  static get properties() {}
```
castastrophe, line 27: Is this empty placeholder needed?  #justcurious

elements/pfelement/src/pfelement.js
```diff
@@ -32,7 +52,14 @@ class PFElement extends HTMLElement {
...
 
   static get observedAttributes() {
-    return ["pfe-theme"];
+    const properties = this.properties;
+    if (properties) {
+      const oa = Object.keys(properties)
+        .filter(prop => properties[prop].observer || properties[prop].cascade || properties[prop].alias)
+        .map(p => this._prop2attr(p));
+      // console.log("observed attributes are", oa);
+      return [`${this.globalAttrPrefix}theme`, ...oa];
```
castastrophe, line 61: Can we have a separate place to store an array of global attributes? Ideally we could define those in a properties function just like components do and have them join?

elements/pfelement/src/pfelement.js
```diff
@@ -49,11 +76,11 @@ class PFElement extends HTMLElement {
   }
 
   get pfeType() {
-    return this.getAttribute(`${prefix}type`);
+    return this.getAttribute(`${tagPrefix}type`);
```
castastrophe, line 79: This should probably be `pfe-g-type`

elements/pfelement/src/pfelement.js
```diff
@@ -176,8 +207,8 @@ class PFElement extends HTMLElement {
     this.classList.add("PFElement");
     this.setAttribute("pfelement", "");
 
-    if (typeof this.props === "object") {
-      this._mapSchemaToProperties(this.tag, this.props);
+    if (typeof this.schemaProps === "object") {
```
castastrophe, line 210: Should we add a TODO to deprecate this for 1.0?

elements/pfelement/src/pfelement.js
```diff
@@ -207,18 +238,200 @@ class PFElement extends HTMLElement {
   }
 
   attributeChangedCallback(attr, oldVal, newVal) {
-    if (!this._pfeClass.cascadingAttributes) {
+    if (attr === "pfe-theme") {
```
castastrophe, line 241: I think we'll need a more dynamic way to set this up because we'll have more than `theme` in the future for global attributes.

elements/pfelement/src/pfelement.js
```diff
@@ -207,18 +238,200 @@ class PFElement extends HTMLElement {
...
+
+      // verify that properties conform to the allowed data types
+      if (!isAllowedType(propDef)) {
+        throw new Error(
+          `Property "${propName}" on ${this.constructor.name} must have type String, Number, or Boolean.`
+        );
+      }
+
+      // verify the property name conforms to our naming rules
+      if (!/^[a-z_]/.test(propName)) {
```
castastrophe, line 286: Is there a way for our compiler to flag this too so it isn't just a client-side notice but an actual compile-time warning/error?

elements/pfelement/src/pfelement.js
```diff
@@ -254,7 +467,7 @@ class PFElement extends HTMLElement {
         }
 
         if (hasPrefix) {
-          attrName = `${prefix}${attr}`;
+          attrName = `${tagPrefix}${attr}`;
```
castastrophe, line 470:
```suggestion
          attrName = `${prefix}-c-${attr}`;
```
?

mwcz: Good catch, that definitely shouldn't be `tagPrefix`, but it's not clear to me what it should be.  I renamed `prefix` to `tagPrefix` to differentiate the `pfe-` tag prefix from `pfe-c-` and `pfe-g-` attribute prefixes.  We could pull the attribute name from the properties system, but it seems wrong for `_mapSchemaToProperties`, whose job is to create properties based on a schema, to look at an alternative property system for attribute names. :thinking: 

Should we dodge this problem by removing `_mapSchemaToProperties`, since all elements will be migrated to the new properties system for 1.0?

elements/pfelement/src/pfelement.js
mwcz: I like it!  This isn't the first time the need for static analysis has come up.  Early on I wanted to be able to `import` or `require` an element in node (in the gulpfile), in order to inject the HTML and CSS strings, rather than doing naive string replacement.  It wasn't possible because HTMLElement is undefined in node, so it threw an error.  I tried a couple of DOM libraries, but ran into issues with both and put it off.

I see two options:

# Stubbed import

Stub out HTMLElement and then import the class.

```js
global.HTMLElement = {};
PfeCard = require("elements/pfe-card/src/pfe-card");
validateProperties(PfeCard);
```

That might work. 

# Full DOM library

```js
require("some-dom-library-that-works");
PfeCard = require("elements/pfe-card/src/pfe-card");
validateProperties(PfeCard);
```

elements/pfelement/src/pfelement.js
mwcz: Yeah, for sure.  I had the idea of giving the base class `static get globalProperties()` which would function exactly like `static get properties()`.  I didn't get around to implementing it yet though.

elements/pfelement/src/pfelement.js
mwcz: As in, remove `_mapSchemaToProperties`?

elements/pfelement/src/pfelement.js
mwcz: Actually, I found a way for the parent and child classes to both use `static get properties()`. :+1:

elements/pfelement/src/pfelement.js
mwcz: I renamed `prefix` to `tagPrefix` to differentiate it from the attribute prefixes, but you're right that if we ever were to change the `pfe-` prefix on tags, it would also make sense to change the prefixes on attributes.  That seems like a good reason to change the name back to `prefix`.  Or, more accurately, prefixPrefix, since it's prefixing the attribute prefix. :laughing:

elements/pfe-badge/src/pfe-badge.js
mwcz: SGTM

elements/pfe-badge/README.md
mwcz: :+1:

elements/pfe-avatar/src/pfe-avatar.js
mwcz: :clinking_glasses:

elements/pfe-avatar/src/pfe-avatar.js
mwcz:
```css
*:first-letter {
    text-transform: uppercase !important;
}
```

:shrug:

elements/pfe-avatar/src/pfe-avatar.js
mwcz: In general, no, but avatar wants there to always be a name, so it defaults to "".

elements/pfe-autocomplete/src/pfe-search-droplist.scss
mwcz: Agree :+1:

elements/pfe-autocomplete/src/pfe-autocomplete.js
mwcz: Yeah, in a different PR though.

elements/pfe-autocomplete/src/pfe-autocomplete.js
mwcz: `docs/content/develop/properties.md`

elements/pfe-autocomplete/src/pfe-autocomplete.js
mwcz: I've been using the pattern `oldInitValue` instead.  Since `_` is a convention for "private", I wanted to use something else.  Whatever the pattern, I agree it should be used in accordion and everywhere.  Kyle and I have also been adding `// @TODO Deprecate property` alongside every old property def.

elements/pfe-accordion/src/pfe-accordion.js
mwcz: I updated the docs to clarify that it expects a property name. :+1:

docs/content/develop/properties.md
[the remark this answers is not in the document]
mwcz: The doc now explains how camelCase gets converted to snake-case, and cascade.

docs/content/develop/properties.md
mwcz: To clarify, `attributeChangedCallback` and `observedAttributes` don't get overriden by using this approach.  They are defined on the base class, and an element author can override them if they choose to.  An element which defines its own `observedAttributes` is overriding the `observedAttributes` that implements the properties system.

I personally don't think it's worth mentioning `cascadingAttributes`.  It's only ever been used in two elements, so I don't think many contributors know or think of it much.  My preference would be to mention it in the changelog or release notes, but not the docs about how things work now.

docs/content/develop/properties.md
mwcz: It uses JS's built-in Number type, which is a floating point type, but floats can represent integers.  That's mentioned, but kind of buried, in the note about precision.  I feel like the name "Number" implies that any kind of number is supported (within the precision limits), but I'm fine with making it explicit if you think it would help.

mwcz: It uses JS's built-in Number type, which is a floating point type, but floats can represent integers.  That's mentioned, but kind of buried, in the note about precision.  I feel like the name "Number" implies that any kind of number is supported (within the precision limits), but I'm fine with making it explicit if you think it would help.

About the paragraph encouraging contributors to use `type: Number` for storing numbers, I feel like that would fit perfectly in a guide.  I see this doc as something like API docs, describing all the things you can do and how they work, whereas describing what you _should_ do would fit better in a guide, such as "Creating your first component" or "Adding typed properties to your component".  Does that make sense?  What do you think?

docs/content/develop/properties.md
mwcz: Good catch!

docs/content/develop/properties.md
mwcz: :+1:

docs/content/develop/properties.md
mwcz: I left a note in your similar suggestion for Number. Looking forward to your thoughts.

elements/pfe-card/src/pfe-card.js
mwcz: Done :+1:

elements/pfe-accordion/src/pfe-accordion.js
kylebuch8: We're going to leave this for now. This is an opt out feature.

elements/pfe-autocomplete/src/pfe-autocomplete.js
kylebuch8: old it is. I'll update any properties prefixed with an underscore.

elements/pfe-autocomplete/src/pfe-autocomplete.js
kylebuch8: "It is now, yeah." said Michael Clayton

elements/pfe-card/demo/index.html
kylebuch8: Yes, it should. I'll make the update to card and band.

elements/pfe-card/src/pfe-card.js
kylebuch8: Done.

elements/pfelement/src/pfelement.js
kylebuch8: No reason for it to be there yet. So we're going to leave it.

elements/pfelement/src/pfelement.js
castastrophe: Cassondra to handle this - need to update the storybook utils

docs/content/develop/properties.md
castastrophe: Since we don't really have a guide yet outside of these docs, maybe we can insert it here for now and pull it out into the guide when we're building it?

elements/pfe-accordion/src/pfe-accordion.js
castastrophe: I mention this because it was listed as an observed attribute which has been deleted.  If the component still works without it, I'm cool leaving it off.

elements/pfe-autocomplete/src/pfe-autocomplete.js
castastrophe: New issue created

elements/pfe-avatar/demo/index.html
```diff
@@ -22,15 +22,15 @@
...
 
     <!-- uncomment the es5-adapter if you're using the umd version -->
-    <script src="/examples/node_modules/@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js"></script>
-    <script src="/examples/node_modules/@webcomponents/webcomponentsjs/webcomponents-bundle.js"></script>
-    <script src="/examples/node_modules/requirejs/require.js"></script>
-    
-    <script>require([
-      '../dist/pfe-avatar.umd.js'
-    ])</script>
+    <!-- <script src="/examples/node_modules/@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js"></script> -->
```
castastrophe, line 25: Why change the demo page rendering in this PR?  I think it's nice to keep it the other way too for the IE11 validation.

elements/pfe-avatar/src/pfe-avatar.js
castastrophe, line 217: Is this documented?  That instead of `this.hasAttribute` we can now use the property instead?

elements/pfe-badge/src/pfe-badge.scss
```diff
@@ -35,7 +35,8 @@ span {
 }
 
 @each $state in (moderate, important, critical, success, info) {
-  :host([pfe-state="#{$state}"]) {
+  :host([pfe-state="#{$state}"]),
```
castastrophe, line 38: Don't need both anymore, just the new prefix

elements/pfe-content-set/src/pfe-content-set.js
```diff
@@ -80,6 +80,12 @@ class PfeContentSet extends PFElement {
       oldId: {
         alias: "id",
         attr: "pfe-id"
+      },
+      upgraded: {
```
mwcz, line 84: What's this used for?  Asking because `this.isConnected` is part of the web component spec and could probably be used instead.

castastrophe: This was going to be how we tag light DOM elements as already having been migrated.  The mutation observer would remove it until the element is moved to the shadow DOM

docs/content/develop/properties.md
castastrophe: Added my thoughts to that thread, let me know.  I think the additional information is extremely helpful.

docs/content/develop/properties.md
```diff
@@ -0,0 +1,348 @@
...
+```
+
+Observer functions are called with arguments `(oldVal, newVal)`.  Like property values, both arguments will be cast to the appropriate type, ie String, Number, or Boolean.
+
+
+## cascade
+
+Default: `undefined`
+
+`cascade` allows an attribute value to be automatically copied to one or more child elements.  The value of `cascade` is a CSS-style selector which is used to match the children that should receive the values.  The selector is applied to the element's children in both the light DOM and the shadow DOM.
```
castastrophe, line 174: Does this accept any selector?  Could I do: `"cascade": "*"` for example?  Or `"cascade": ":not(.foo)"`

docs/content/develop/properties.md
```diff
@@ -0,0 +1,348 @@
...
+      default: "true",
+      prefix: false
+    }
+  };
+}
+```
+
+The property name `ariaHidden` results in an attribute named `aria-hidden`.  If prefix were not set to false, the generated attribute name would be `pfe-c-aria-hidden` instead.
+
+Note: [ARIA attributes](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) must be defined as Strings, for even though attributes like `aria-hidden` seem like booleans, they are not true boolean attributes, and they only accept values of `"true"` and `"false"`.
```
castastrophe, line 228:
```suggestion
Note: [ARIA attributes](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def) must be defined as Strings, for even though attributes like `aria-hidden` seem like Booleans, they are not true boolean attributes (`aria-hidden` vs. `aria-hidden="true"`) and need to receive string values of `"true"` and `"false"`.
```

docs/content/develop/properties.md
```diff
@@ -0,0 +1,348 @@
...
+
+## alias
+
+Default: `undefined`
+
+The `alias` field can be used to link two properties to each other.  `alias` expects to be given the name of another property, to which it will copy its value.  When a property value is changed, if it has an alias, the new value will be sent to the alias.
+
+Aliasing is intended to help when migrating to a new property/attribute name.  It allows for two different properties to be simultaneously "active".
```
castastrophe, line 236:
```suggestion
Aliasing is intended to help when migrating to a new property/attribute name.  It allows for two different properties to be simultaneously "active".

Aliased values do not need to have a new set of definitions for cascade or observer as those states will be managed by the attribute to which it is being aliased (see example below).
```

docs/content/develop/properties.md
castastrophe: Asked directly and the answer was yes!  Anything querySelectorAll accepts.

castastrophe, line 174:
```suggestion
`cascade` allows an attribute value to be automatically copied to one or more child elements in the Light DOM or Shadow DOM<sup>[1]</sup>.  The value of `cascade` is a CSS-style selector which is used to match the children that should receive the values.  The selector is applied to the element's children in both the Light DOM and the Shadow DOM.

<sup>[1]</sup> There is not a way to target _only_ the Light DOM or _only_ the Shadow DOM.  If this is necessary, we recommend adding a unique identifier to the Shadow DOM elements and making use of the `:not` selector.
```

castastrophe (approved):
I've now gone through and familiarized myself with the new code and it looks great! 🎉👍

docs/content/develop/properties.md
```diff
@@ -0,0 +1,352 @@
...
+ - Default value initialization.
+ - Data types (for properties).
+ - Observer function to be run when a value changes.
+ - Optional namespacing prefix for attribute names.
+ - Aliasing properties to one another, for phased deprecation.
+ - And more!
+
+## Introduction
+
+To introduce the topic, consider a component that counts how many times something has happened.We can give it the tag `<pfe-count>` and a `count` attribute (`<pfe-count count="0">`) to set the initial count.  We might also want a local property `this.count` for convenient referencing and updating of the count (ex: `this.count += 1` vs `this.setAttribute(this.getAttribute("count") + 1)`).
```
starryeyez024, line 24:
```suggestion
To introduce the topic, consider a component that counts how many times something has happened. It may have a tag of `<pfe-count>`, and a `count` attribute (`<pfe-count count="0">`) which sets the initial count.  We might also want a local property `this.count` for convenient referencing and updating of the count (ex: `this.count += 1` vs `this.setAttribute(this.getAttribute("count") + 1)`).
```

docs/content/develop/properties.md
```diff
@@ -0,0 +1,352 @@
...
+    }
+  };
+}
+```
+
+With this property definition for `count` we get:
+
+ - **A property**: The property's name is `count`, so `this.count` is created and initialized to the default value, the number `0`.  
+ - **An attribute**: An attribute, `pfe-c-count` is also created, and will be kept in sync with `this.count`.  The `pfe-c-` prefix is explained [below](#prefix).
+
```
starryeyez024, line 43:
```suggestion
```

docs/content/develop/properties.md
```diff
@@ -0,0 +1,352 @@
...
+And many other options for watching the property for changes, deprecating attributes, and more.  All the options for property definitions are explained below, but first, a bit about how property/attribute pairs are synchronized.
```
starryeyez024, line 44:
```suggestion
 - And many other options for watching the property for changes, deprecating attributes, and more.  
 
All the options for property definitions are explained below, but first, a bit about how property/attribute pairs are synchronized.
```

docs/content/develop/properties.md
```diff
@@ -0,0 +1,352 @@
...
+  }
+
+  handleCount(oldVal, newVal) {
+    console.log(`count changed from ${oldVal} to ${newVal}`);
+  }
+
+}
+```
+
+Observer functions are called with arguments `(oldVal, newVal)`.  Like property values, both arguments will be cast to the appropriate type, ie String, Number, or Boolean.
```
starryeyez024, line 167:
```suggestion
Observer functions are called with arguments `(oldVal, newVal)`.  Like property values, both arguments will be cast to the appropriate type, i.e. String, Number, or Boolean.
```

docs/content/develop/properties.md
```diff
@@ -0,0 +1,352 @@
...
+      alias: "new"
+    }
+  };
+}
+```
+
+With these property definitions, changes to `old` will be "forwarded" along to `new`.
+
+```javascript
+el.source = 4;
```
starryeyez024, line 262: I don't understand this example, should these say: 
```
el.old = 4;
el.new === 4; // true
```

docs/content/develop/properties.md
```diff
@@ -0,0 +1,352 @@
...
+el.destination === 4; // true
+```
+
+Two-way aliases are allowed.
```
starryeyez024, line 266: I'm unsure of why you would use a two-way alias?

docs/content/develop/properties.md
```diff
@@ -0,0 +1,352 @@
...
+
+Default: `undefined`
+
+The `attr` field overrides the property's default generated attribute name and specifies a custom name to be used as this property's bound attribute.
+
+```javascript
+static get properties() {
+  return {
+    foo: {
+      attr: "the-one-and-only-foo"
```
starryeyez024, line 291: it could totally just be me but this feels unclear as to which thing is the new source of truth, or the new attribute, is it `foo` or `the-one-and-only-foo`

starryeyez024: I also don't know when I would use this approach vs. setting an alias

docs/content/develop/properties.md
```diff
@@ -0,0 +1,352 @@
...
+
+## title
+
+Default: `undefined`
+
+The `title` field defines a human-friendly title for the property, which will be used to label the property's form fields in the Storybook-powered demo site.
+
+## observedAttributes & attributeChangedCallback
+
+Property definitions are driven by PFElement's attributeChangedCallback and observedAttributes.  Due to this, any PatternFly Element using property definitions should not define an attributeChangedCallback (use [observer](#observer) instead) or observedAttributes (attributes created by property definitions are automatically observed).
```
starryeyez024, line 311:
```suggestion
Property definitions are driven by PFElement's `attributeChangedCallback` and `observedAttributes`.  Due to this, any PatternFly Element using property definitions should not define an `attributeChangedCallback` (use [observer](#observer) instead) or `observedAttributes` (attributes created by property definitions are automatically observed).
```

starryeyez024 (changes requested):
This documentation is very thorough and helpful! 🎉  A few questions/suggestions - but because I'm in unfamiliar territory, please take with a grain of salt. :)

docs/content/develop/properties.md
castastrophe: This is using a setter so if you want to update the value of a property you just set it like a variable.

docs/content/develop/properties.md
mwcz: Sounds good!

docs/content/develop/properties.md
mwcz: What's important is that two-way aliasing doesn't fail catastrophically.  It would be very easy for it to become an infinite loop.  "Support" for two-way aliasing means not falling over dead when it's done.

But for the sake of argument, two-way aliasing could be used to keep two attributes in sync with each other, and that could be useful.  For example, let's say we're integrating a component with some third-party library or framework, and that integration required a framework-specific attribute.  We should not lock in to that attribute, so we create our own custom attribute that aliases to the framework-specific attribute, _and back_.  That way, we can keep on supporting our own custom attribute even if we remove that integration or add another integration.

docs/content/develop/properties.md
mwcz: Good catch, I changed the property names to `new` and `old` but forgot to update the code example.

docs/content/develop/properties.md
```diff
@@ -0,0 +1,352 @@
...
+    }
+  };
+}
+```
+
+With these property definitions, changes to `old` will be "forwarded" along to `new`.
+
+```javascript
+el.source = 4;
+el.destination === 4; // true
```
mwcz, line 265:
```suggestion
el.old = 4;
el.new === 4; // true, because setting el.old caused the value to be copied into el.new
```

docs/content/develop/properties.md
mwcz: For the alias question, you set an alias when you want to copy a value from one prop/attr to another.  It has nothing to do with attribute names.

elements/pfe-avatar/src/pfe-avatar.js
mwcz: Sorry, I didn't see this message before.  It's documented that you _can_ reference properties this way, but not why you'd want to (it's a _lot_ less to type, and you get type conversions for free).  It would be a great thing to put in a tutorial.

mwcz:
I did some testing in the latest release of Edge (Edgium 85) on Windows 10.  I checked...

 - pfe-accordion
 - pfe-autocomplete
 - pfe-avatar
 - pfe-badge
 - pfe-band
 - pfe-button
 - pfe-card
 - pfe-collapse

...before running out of time.  I did not find any Edge-specific problems, but I did find a pfe-avatar issue and two potential pfe-band issues (but I could be just misunderstanding pfe-band).

## pfe-avatar

The pfe-avatar demo page isn't working right now, in any browser.

When pfe-avatar--width isn't defined, the following line errors because `.replace` can't be called on `undefined`.

```js
const size = this.cssVariable("pfe-avatar--width").replace(/px$/, "") || PfeAvatar.defaultSize;
```

## pfe-band

On the demo page, the "Complement band, layout classes" section has a card titled "Aside: right body bottom", but it's displayed on the left.

Screenshot from 2020-10-12 13-17-21

In the "Short content band" section, a card titled "Aside: left body top" appears on the right.  The CTA is also large, but that seems to be inentional since its text reads "This CTA has overrides".

Screenshot from 2020-10-12 13-18-04

(Note: it looks like this in

docs/content/develop/properties.md
```diff
@@ -0,0 +1,355 @@
...
+
+```javascript
+name: {
+  type: String
+}
+```
+
+Valid values: any string, `null`, and `undefined`.
+Though all attribute values are technically strings, if you are assigning a `true/ false` paradigm or checking that a variable exists (implicit `true/false`, i.e., `hidden` or `required`), please assign that property a type of `Boolean`.  This will ensure the update and injection of that value is handled correctly by the base class.
+### Number
```
alazzara, line 73:
```suggestion

### Number

```

docs/content/develop/properties.md
starryeyez024: Not sure if this is new or I just missed it the first time but the line below this is great, maybe you can just move it above the example?

"The example below shows a property definition which results in a property named `foo` and an attribute named `the-one-and-only-foo`.

docs/content/develop/properties.md
```diff
@@ -0,0 +1,355 @@
...
+static get properties() {
+  return {
+    foo: {
+      attr: "the-one-and-only-foo"
+    },
+  };
+}
+```
+
+This property definition results in a property named `foo` and an attribute named `the-one-and-only-foo`.
```
castastrophe, line 299:
```suggestion
```

docs/content/develop/properties.md
```diff
@@ -0,0 +1,355 @@
...
+  };
+}
+```
+
+## attr
+
+Default: `undefined`
+
+The `attr` field overrides the property's default generated attribute name and specifies a custom name to be used as this property's bound attribute.
+
```
castastrophe, line 288:
```suggestion

The example below shows a property definition resulting in a property named `foo` (i.e., js: `this.foo`) and an attribute named `the-one-and-only-foo` (i.e., html: `<pfe-foo the-one-and-only-foo>`).
```

.storybook/utils.js
```diff
@@ -1,4 +1,8 @@
+import PFElement from "../elements/pfelement/dist/pfelement";
```
castastrophe, line 1:
```suggestion
```

elements/pfe-tabs/src/pfe-tabs.js
```diff
@@ -269,6 +275,24 @@ class PfeTabs extends PFElement {
 
     this._linked = false;
     this._linkPanels();
+
+    if (mutationsList) {
```
castastrophe: @kylebuch8 This should be handled by cascade now

kylebuch8: @castastrophe - The cascade doesn't work for items that are dynamically added. It only fires on attribute changes so we'll need to keep that mutationList part of the code in there.

starryeyez024 (approved):
Later: Greater Than Moments

PR #1094, opened 2020-09-10
Cpfed 3805 set focus on pfe nav close

kelsS:
---
name: CPFED 3805 Set focus when closing nav with escape key
labels: "feature, ready for review"
---

## pfe-navigation

- code cleanup and comment clean up
- update keydown event listener to use logic to close menu even when mobile menu dropdown is open
- remove unneeded mobile menu keydown event listener
- add removeEventListeners to disconnected callback
- add dynamic settings for aria controls, aria hidden, aria expanded for util buttons and mobile menu toggle
- update general keyboard listener function to use event.key instead of event.which bc that is deprecated
- added note to look into bug on firefox: escape key keydown function only fires if you have tabbed into the menu but not if you have just clicked the menu open with the mouse click

### Testing instructions

1. Pull down this branch
2. Start up pfe-nav demo locally
3. Tab through the menu items on the /single-nav.html demo page and press the escape key, this should set the focus on the trigger button and close the entire nav

### Merging


elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -672,6 +670,7 @@ class PfeNavigation extends PFElement {
     }
 
     // Copy light DOM menu into new wrapper, to be put in shadow DOM after manipulations
+    // @todo: look into replacing the light dom nav menu with the shadow dom copy instead of copying the light dom into the shadow dom and having two navs at the same time with the same content, not great for a11y to have two copies of the same menu, the two copies are causing tab order issues on firefox
```
wesruv, line 673: We could add a style to `display: none` all ancestors of `pfe-navigation` in the light DOM after the component is mounted?

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -682,10 +681,31 @@ class PfeNavigation extends PFElement {
 
       // Convert dropdown links into buttons
       const dropdownButton = document.createElement("button");
+      // Select mobile menu toggle button
+      const mobileMenuToggleButton = this.shadowRoot.getElementById("mobile__button");
```
wesruv, line 685: Some of these elements have pointers set on the object instance, you can use that instead of re-selecting the element:
https://github.com/patternfly/patternfly-elements/blob/CPFED-3689-new-navigation-epic/elements/pfe-navigation/src/pfe-navigation.js#L91-L104

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -682,10 +681,31 @@ class PfeNavigation extends PFElement {
...
+      const siteSwitcherContent = this.shadowRoot.querySelector(".pfe-navigation__all-red-hat-wrapper");
+      // Get Site Switcher content id
+      const siteSwitcherContentId = siteSwitcherContent.getAttribute("id");
+
       // Move over or add important attributes and content
       dropdownButton.setAttribute("class", dropdownLink.getAttribute("class"));
       dropdownButton.classList.add("pfe-navigation__menu-link--has-dropdown");
+      // set aria-expanded to false initially bc they will be closed on page load
       dropdownButton.setAttribute("aria-expanded", "false");
+      searchButton.setAttribute("aria-expanded", "false");
```
wesruv, line 707: I _think_ `this._addCloseDropdownAttributes()` would be a better fit than individually setting these things, but double check me...

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -697,26 +717,29 @@ class PfeNavigation extends PFElement {
...
 
       dropdownWrapper.classList.add("pfe-navigation__dropdown-wrapper");
       if (dropdown.classList.contains("pfe-navigation__dropdown--single-column")) {
         dropdownWrapper.classList.add("pfe-navigation__dropdown-wrapper--single-column");
       }
       dropdownWrapper.setAttribute("id", dropdownId);
+      // set aria-hidden to true initially bc the content is hidden on page load
       dropdownWrapper.setAttribute("aria-hidden", "true");
+      mobileMenuToggleContent.setAttribute("aria-hidden", "true");
+      siteSwitcherContent.setAttribute("aria-hidden", "true");
```
wesruv, line 733: Same comment with `this._addCloseDropdownAttributes()`

wesruv (changes requested):
Minor comments

wesruv:
I resolved a merge conflict, the id on single-nav.html had a -2 at the end? Not sure if you wanted that?

kelsS:
> I resolved a merge conflict, the id on single-nav.html had a -2 at the end? Not sure if you wanted that?

@wesruv  ahh yes thank you, I think that was leftover from copying one of the navs from the demo page with multiple ids to the single-page version.

elements/pfe-navigation/src/pfe-navigation.js
kelsS: @wesruv that might work, seems worth a try. Any ideas on what is the best way to adds some styles to a web component only if it mounts successfully? Not sure what the logic is for determining when a web component mounts. I think we only need to hide the nav in the light dom and not hide the slots.

wesruv: I've been styling `.PFElement`, which is a class that gets added to mounted pfe components, but I never asked @castastrophe (or anyone else) maybe there's something better?

elements/pfe-navigation/demo/single-nav.html
```diff
@@ -43,7 +43,7 @@
     ])</script>
   </head>
   <body unresolved style="background-color: #ddd;">
-     <pfe-navigation id="pfe-navigation-2">
```
castastrophe: Suggestion: You can add `role="navigation"` to pfe-navigation, still get the semantic value of the navigation tag but with less mark-up for the author to have to remember to put on their page.

elements/pfe-navigation/demo/single-nav.html
```diff
@@ -43,7 +43,7 @@
...
+    <pfe-navigation id="pfe-navigation-2">
       <nav class="pfe-navigation" aria-label="Main Navigation">
         <div class="pfe-navigation__logo-wrapper" id="pfe-navigation__logo-wrapper">
```
castastrophe: I strongly advise against requiring classes in light DOM in order to upgrade the component.  This is a poor author experience because if you make a typo or put them in the wrong order, the entire thing breaks and our team will get a lot of calls to "fix" the component when it's a case of poor implementation.  My approach has been build the simplest light DOM mark-up possible, focus on the required tags for graceful degradation, and then build the component up from there.  Slots can help with moving items into their appropriate locations so that if the user puts them in the wrong place, you can grab and move it easily.  Happy to break out on this.

elements/pfe-navigation/src/pfe-navigation.js
castastrophe: In the template, you can do: `<slot hidden></slot>` and add styles for `[hidden]` to your sheet to display none.  This will essentially hide all the light DOM.

kelsS: thanks @castastrophe! hidden attr sounds good to me. really want to get used to using that attr when I need to display: none content. How does one go about adding an attribute to the light dom once the component mounts, what is the logic to know when has actually successfully mounted? Do you have any examples of that I could take a look at?

elements/pfe-navigation/src/pfe-navigation.js
kelsS: @wesruv I updated to using the pointers already created and am setting the initial aria settings using an updated version of this._addCloseDropdownAttributes(). I updated that function to also set the aria-controls to the id of the corresponding content panels.

elements/pfe-navigation/src/pfe-navigation.js
kelsS: @wesruv see above comment, using that updated version of that function seems to be working well.

elements/pfe-navigation/src/pfe-navigation.js
kelsS: @wesruv see above comment

elements/pfe-navigation/demo/single-nav.html
kelsS: @castastrophe good point, @wesruv wanted to make sure you saw the above comment. I do think this is a larger conversation outside of this PR, so maybe we should all regroup on this.

elements/pfe-navigation/src/pfe-navigation.html
```diff
@@ -17,11 +17,9 @@
       <div id="pfe-navigation__menu-wrapper" class="pfe-navigation__menu-wrapper">
```
castastrophe, line 17: @kelsS If you add a `<slot hidden></slot>` here, that will capture all your light DOM and wrap it in the hidden styles.  No altering light DOM mark-up at all needed!  🎊

elements/pfe-navigation/demo/index_react.html
```diff
@@ -11,27 +11,25 @@
...
-                <a href="#" className="pfe-navigation__menu-link" aria-haspopup="true" aria-expanded="false">
+            <ul class="pfe-navigation__menu" id="pfe-navigation__menu">
+              <li class="pfe-navigation__menu-item">
+                <a href="#" class="pfe-navigation__menu-link">
                   Products
                 </a>
 
-                <div className="pfe-navigation__dropdown">
-                  <section>
+                <div class="pfe-navigation__dropdown pfe-navigation__12-column-grid-wrapper">
```
wesruv, line 27: I think that might need to be `className`

wesruv: Probably something all of the `class` attributes need for React in this file.

elements/pfe-navigation/demo/index_react.html
```diff
@@ -175,31 +201,78 @@ <h3>
...
+              Cart
+            </button>
+          </li>
+
+          <div slot="pfe-navigation--search" class="pfe-navigation__search pfe-navigation__search--default-styles">
             <form>
-              <label>Search</label>
-              <input type="text" placeholder="Search" />
-              <button>Search</button>
+              <label for="pfe-navigation__search-label1" class="sr-only">Search the Red Hat Customer Portal</label>
```
wesruv, line 273: I believe this will need to be `htmlFor` instead of `for`

elements/pfe-navigation/src/pfe-navigation.html
wesruv: Which `slot`? 🤔 Is this code comment not on the line it was intended for?

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -182,11 +175,16 @@ class PfeNavigation extends PFElement {
   }
 
   disconnectedCallback() {
-    // @todo Remove all listeners to be thorogh!
+    // @todo Remove all listeners to be thorough!
```
wesruv, line 178: 😂 woops

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -314,20 +312,28 @@ class PfeNavigation extends PFElement {
    * Almost all open/close actions should go through this._changeNavigationState, not this function
    *
    * @param {object} toggleElement Toggle Button DOM Element
+   * @param {object} toggleId Toggle Button DOM Element ID attribute
```
wesruv, line 315: Confused here, `toggleId` isn't a param for the fuction? This could work as a code comment above the var declaration?

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -314,20 +312,28 @@ class PfeNavigation extends PFElement {
...
    * @param {object} dropdownWrapper Dropdown wrapper DOM element
+   * @param {object} dropdownWrapperId Dropdown wrapper DOM element ID attribute
```
wesruv, line 317: Same comment as above ^

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -750,24 +756,17 @@ class PfeNavigation extends PFElement {
     this._allRedHatToggle.addEventListener("click", this._toggleAllRedHat);
 
     // General keyboard listener attached to the entire component
+    // @todo/bug: figure out why this event listener only fires once you have tabbed into the menu but not if you have just clicked open menu items with a mouse click on Firefox - functions properly on Chrome
```
wesruv, line 759: That might be why I had the menuButton event listener?

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -1173,6 +1103,10 @@ class PfeNavigation extends PFElement {
     }
   }
 
+  /**
+   * All Red Hat Site Switcher XMLHttpRequest API Request
```
wesruv, line 1107: ++

wesruv (changes requested):
Sorry, a few issues and comments. Might want to check if this issue was fixed by the menubutton listener:
```
// @todo/bug: figure out why this event listener only fires once you have tabbed into the menu but not if you have just clicked open menu items...
```

elements/pfe-navigation/demo/single-nav.html
wesruv: Yea, we can talk about this in our upcoming meeting.

elements/pfe-navigation/demo/single-nav.html
wesruv: That's a good thought, I'm pretty sure I need the wrapper for non-JS version layout, so might as well make it a `<nav>` tag. I'll double check though.

elements/pfe-navigation/src/pfe-navigation.html
castastrophe: Yeah GitHub sometimes moves things around a bit.  Add an unnamed slot which will capture all other light DOM added to the component.

wesruv: Gotcha, I intentionally avoided a default slot, we can talk about this in our meeting on Monday about the approach and pros/cons.

elements/pfe-navigation/src/pfe-navigation.js
kelsS: @wesruv oh oops meant to move that down as a comment above the var, thanks for catching that. I will get that updated.

elements/pfe-navigation/src/pfe-navigation.js
kelsS: See the comment above.

elements/pfe-navigation/src/pfe-navigation.js
kelsS: @wesruv I don't think so since the issue is happening at the desktop breakpoint on firefox when the menu button is not present. But I will make a ticket to investigate further.

elements/pfe-navigation/src/pfe-navigation.js
kelsS: lol

elements/pfe-navigation/demo/index_react.html
kelsS: Thanks for catching! I will get that updated.

elements/pfe-navigation/demo/index_react.html
kelsS: See comment above ^

PR #1137, opened 2020-10-08
chore: GitHub Access

castastrophe:
## Action updates

Big performance improvements in this PR.  By splitting out the build into it's own job and then chaining that with the test and e2e jobs running concurrently (utilizing a cache created by the first job of the compiled assets). 7-8 minutes for the test task down to 4-5 minutes.

### Global 
- Use `GITHUB_TOKEN` where possible so that merges are done by the GitHub Bot instead of @castastrophe 

### Automerge
- Run when a pull review request is submitted
- Run when a pull request is labeled or unlocked

### Labeling
- Run when pull request review is submitted
- Run when pull requests are converted to or from draft
- Flag PRs as "work in progress" when converted to draft state

### Project labeling
- Simplify the logic

### Build & test
- Test and E2E can now run concurrently
- NPM and compiled assets are now cached; accessed by the downstream jobs
- Support manual runs
- Run on any pushes to the master or main branch
- Make the compiled assets and the baseline assets available as artifacts

## Updated test script
- Updated the test script to use node


zhawkins (approved):

castastrophe:
Todo for me: it looks like the builds are being fired multiple times so it could be something to do with labeling.

PR #1148, opened 2020-10-20
feat: Add script that will auto-update open PRs when their base-branch is updated

castastrophe:
This action will update any open PR if their base branch is updated.  It will _not_ resolve merge conflicts.  If it detects a merge conflict, the auto-update will fail.

### Ready-for-merge Checklist

Check off items as they are completed.  Feel free to delete items if they are not applicable.

- [x] Changelog updated (not needed for documentation updates).

### Merging


.github/workflows/autoupdate.yml
```diff
@@ -0,0 +1,20 @@
+# This workflow updates all open PRs if their base branch is pushed to
+# https://github.com/chinthakagodawita/autoupdate
+name: Auto-update open PRs
+on:
+  # This will trigger on all pushes to all branches
+  push: {}
+  # Alternatively, you can only trigger if commits are pushed to certain branches, e.g.:
+  # push:
+  #   branches:
+  #     - master
```
mwcz, line 10: This is commented out, did it not work?  It seems more efficient to only try to update branches when master is updated, instead of when any branch is updated.  With the number of open PRs we have, updating any of them would cause this job to run on every other PR branch, right?  Trying to understand how this works.

I can imagine a PR that was branched off of another PR (rather than master) needing this, though.

mwcz (commented):
This looks amazing!  I only have one...

...

dramatic pause

...

... _question_.

.github/workflows/autoupdate.yml
castastrophe: > I can imagine a PR that was branched off of another PR (rather than master) needing this, though.

This was why I commented it out but left it in there.  I thought it would be nice if PRs against other branches (the CP-FED nav project for example) would be valuable and helpful for larger initiatives that we don't want to merge directly into master; granted we could just add those big feature branches to the supported branches but 🤷 .

I thought I'd leave it commented out in case we're finding all pushes against open PRs is just too much work on the system.  I can also incorporate some label-based (or other) logic to limit runs in the future (not on WIP for example).

mwcz: Sounds good!

mwcz (approved):
Lift Gumbo To Mouth

PR #1150, opened 2020-10-20
chore: Add pretty labels to clarify actions view in UI

castastrophe:
- Prettify the UI by adding clearer labels and descriptions
- Separate labelling steps into distinct jobs so they run concurrently


mwcz (approved):
"Lush Garden", Types Me

PR #1156, opened 2020-10-26
Cpfed 3852 sticky nav test

kelsS:
---
name: Cpfed 3852 sticky nav test - add sticky nav feature and test top banner
labels: "feature"
---

## pfe-navigtation

- Added sticky nav feature and test RHEL Insights top banner. 
- Added sticky nav demo page. 
- Added content from redhat.com to the page so that it scrolled. 
- Updated overlay to fix issue were you could click on the content below the overlay when menu was open.

### Testing instructions

1. Pull down branch
2. Scroll up and down the page and you will see the sticky nav stick to the top and hide the banner when you scroll down the page then reshow the banner when you scroll back up.
3. Open menu then click on the overlay anywhere above content - should close menu but not activate any links or content on the page
4. Acceptance Criteria for feature was replicated from redhat.com

### Merging


elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -178,6 +179,18 @@ class PfeNavigation extends PFElement {
...
+    // Initial position of this element from the top of the screen
+    this.top = this.getBoundingClientRect().top || 0;
+
+    // If the nav is set to sticky, run the sticky handler and attach scroll event to window
+    if (this.hasAttribute("pfe-sticky") && this.getAttribute("pfe-sticky") != "false") {
+      // Run the sticky check on first page load
+      this._stickyHandler();
+
+      // Attach the scroll event to the window
+      window.addEventListener("scroll", this._stickyHandler);
```
wesruv, line 192: Can you throttle this with requestAnimationFrame, e.g. https://wiki.developer.mozilla.org/en-US/docs/Web/API/Document/scroll_event#Scroll_event_throttling

elements/pfe-navigation/src/pfe-navigation.scss
```diff
@@ -39,6 +39,15 @@ See mixin blocks before selectors that contain the selector name and the three l
   display: none;
 }
 
+:host(.pfe-sticky) {
+  position: sticky;
```
wesruv, line 43: Change this to fixed if we're going with a JS solution for this (we could also try just using `position: sticky` instead with no JS and event listener, but that wouldn't work on IE, and sticky has bugs.)

elements/pfe-navigation/src/pfe-navigation.scss
```diff
@@ -64,7 +73,7 @@ See mixin blocks before selectors that contain the selector name and the three l
   left: 0;
   width: 100%;
   height: 100%;
-  z-index: -1;
+  z-index: 97;
```
wesruv, line 76: Instead of setting this separately, if you add a z-index to the `pfe-navigation` tag and leave this at -1 it'll produce the same effect without making two z-indexes we have to manage.

Parent elements that are 'positioned' and have a z-index will 'scope' children elements. So this will be -1 in pfe-nav, not -1 in the page.

wesruv (changes requested):

elements/pfe-navigation/src/pfe-navigation.scss
kelsS: @wesruv the JS and position sticky is in place to work with banners when they are above the navigation. I didn't think we could replicate that feature using position fixed (at least not very easily) so I went with the solution already used in the original nav.

elements/pfe-navigation/src/pfe-navigation.scss
kelsS: @wesruv Oh cool, thanks for explaining that. I will get that updated.

kelsS: @wesruv added those changes.

elements/pfe-navigation/src/pfe-navigation.js
kelsS: @wesruv added those changes.

elements/pfe-navigation/src/pfe-navigation.scss
wesruv: Position sticky does what your JS scroll & sticky handler does, but with bugs. If position sticky was better supported **wasn't** a buggy mess, I'd say just use `sticky` in the CSS, and no JS.

Have you tested with this line being changed to fixed? It should work exactly the same, but with less browser bugs.

wesruv (changes requested):

elements/pfe-navigation/src/pfe-navigation.scss
kelsS: @wesruv ahh I understand now, thanks for explaining further! When I tested initially I missed some more CSS tweaks I needed to make using position fixed for everything to look good which made me concerned I could not replicate the sticky behavior with fixed.  I'll get those changes made.

kelsS: @wesruv added those changes.

PR #1172, opened 2020-11-04
fix: Fix e2e testing to take full page shots

castastrophe:
## tldr;

Visual regression tests were only snapshotting the visible portion of the page.  This PR adjust them to snapshot the entire page.

## Changes outline

- .gitattributes updates to make images easier to diff; threw in archive settings for good measure
- Replaced all `browser.saveScreen(element);` with `browser.saveFullPageScreen(element);`
- Replaced all `expect(browser.checkScreen(element)).toBeLessThan(1.25);` with `expect(browser.checkFullPageScreen(element)).toBeLessThan(1.25);`
- Added a pfe-styles visual regression test
- Committed updated baselines

### Webdriver config
- Add ability to e2e test single components (needed this for testing)
- Add `disableCSSAnimation: true`
- Add the `spec` reporter to make results easier to grok


wdio.conf.js
```diff
@@ -45,13 +47,15 @@ exports.config = {
...
-        blockOutToolBar: true
+        blockOutToolBar: true,
+        disableCSSAnimation: true
       }
     ]
   ],
   onPrepare: () => {
     proc = exec("http-server");
   },
+  // Located in your wdio.conf.js file
```
castastrophe, line 58:
```suggestion
```

kylebuch8 (approved):
Loathe Government? Try Monarchy!

starryeyez024:
I noticed that both dropdowns are expanded, and thus one is overlapping the other. maybe we could add a few lines of space between them? [test/vrt-baseline/desktop_ie/pfe-dropdown.png](https://github.com/patternfly/patternfly-elements/blob/a5169461f5da0f9df9a00623389027e0210b35e2/test/vrt-baseline/desktop_ie/pfe-dropdown.png)

castastrophe:
> I noticed that both dropdowns are expanded, and thus one is overlapping the other. maybe we could add a few lines of space between them? [test/vrt-baseline/desktop_ie/pfe-dropdown.png](https://github.com/patternfly/patternfly-elements/blob/a5169461f5da0f9df9a00623389027e0210b35e2/test/vrt-baseline/desktop_ie/pfe-dropdown.png)

That's just the first dropdown - notice that the order is `Link 1, Link 2...Link 4`; I think what you're seeing in the line-rule break which looks really fat in IE11

starryeyez024:
> > I noticed that both dropdowns are expanded, and thus one is overlapping the other. maybe we could add a few lines of space between them? [test/vrt-baseline/desktop_ie/pfe-dropdown.png](https://github.com/patternfly/patternfly-elements/blob/a5169461f5da0f9df9a00623389027e0210b35e2/test/vrt-baseline/desktop_ie/pfe-dropdown.png)
> 
> That's just the first dropdown - notice that the order is `Link 1, Link 2...Link 4`; I think what you're seeing in the line-rule break which looks really fat in IE11

Ah! yeah it definitely looked like two. Thx! 👍

PR #1185, opened 2020-11-19
test: updating pfe-accordion vrt baseline images

kylebuch8:
pfe-accordion baseline images need to be updated.


starryeyez024 (approved):
Last Great Thanksgiving Meal

PR #1186, opened 2020-11-19
Cpfed 3850 close menu nav loses focus

kelsS:
---
name: Close any open menus when nav loses focus
labels: "feature, ready for review"
---

## pfe-navigation

Added feature to close any open menu dropdowns when the entire nav loses focus so that keyboard users cannot get to the rest of the page content with the nav still open.

### Testing instructions

1. Pull down changes
2. Tab through nav and onto the page content with any dropdown menu open
3. The dropdown menu should close when the last item in the nav loses focus and the focus is on the page content and not the nav.

### Merging

### DIFF
--- a/elements/pfe-navigation/src/pfe-navigation.js
+++ b/elements/pfe-navigation/src/pfe-navigation.js
@@ -1408,11 +1408,12 @@ class PfeNavigation extends PFElement {
    * Trap keyboard while Menu is open
    */
   _tabKeyEventListener(event) {
-    // when menu is open pfe-navigation-open-toggle !== null
     const currentlyOpenToggleId = this.getAttribute(`${this.tag}-open-toggle`);
     const openToggle = this.shadowRoot.getElementById(currentlyOpenToggleId);
     const openToggleId = this.getAttribute(`${this.tag}-open-toggle`);
-    const mobileMenuToggle = this.shadowRoot.querySelector("#mobile__button");
+    // const mobileMenuToggle = this.shadowRoot.querySelector("#mobile__button");
+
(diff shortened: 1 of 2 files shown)


elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -195,7 +196,190 @@ class PfeNavigation extends PFElement {
         });
       });
     }
-  }
+
+    // Last link visually
```
wesruv, line 200: This is a lot of commented out code?

elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -838,6 +1023,18 @@ class PfeNavigation extends PFElement {
     // @todo/bug: figure out why this event listener only fires once you have tabbed into the menu but not if you have just clicked open menu items with a mouse click on Firefox - functions properly on Chrome
     this.addEventListener("keydown", this._generalKeyboardListener);
 
+    // @todo: check to see if this is the best place for these variables to be defined or should they go in the constructor?
```
wesruv, line 1026: Looks like you're not using these vars anywhere but here, so here seems fine?

wesruv (approved):
Feel like the commented code could go, but otherwise it looks good.

I haven't tested it yet, but we can handle any issues as bugs, looks sound to me!

elements/pfe-navigation/src/pfe-navigation.js
kelsS: @wesruv thanks for catching that, that was from my original testing and working through the feature. Got that stale code removed.

PR #1207, opened 2020-12-02
CPFED-3942: Initial try at adding appropriate color variables

wesruv:


elements/pfe-navigation/src/_variables-mixins.scss
```diff
@@ -6,11 +6,39 @@ $LOCAL: navigation;
 /// PFE-Local setup variables
 /// These variables will be namespaced and become css vars
 $LOCAL-VARIABLES: (
```
wesruv, line 8: @castastrophe 
The main event.
After we meet my next step will be to 'default' certain local vars to theme vars.
Also think there'll end up being more vars, be curious on your impression of what could use vars

elements/pfe-navigation/src/pfe-navigation.scss
```diff
@@ -145,25 +145,25 @@ See mixin blocks before selectors that contain the selector name and the three l
   display: none;
```
wesruv, line 145: I only implemented the new local vars in this file so far.

PR #1218, opened 2020-12-09
Making primary-list shadow DOM and adding basic styling to component

wesruv:
Styling the list as light DOM was causing problems
Added a primary-list footer per designs for this in the Portal nav
Added some @todo stuff to follow up on


wesruv:
Figured out a better way to do this without shadow root...

PR #1235, opened 2020-12-17
feat: pfe-clipboard

heyMP:
## pfe-clipboard

pfe-clipboard allows developers to include button to "copy url to clipboard" that can be placed web pages and handle the entire process of displaying a button that, when clicked, will copy the url to the users system clipboard.

### Related Issues

- #803
- #810 (possibly)

### Preview

Link(s) to demo page(s) where this element can be viewed:
- [Link](https://deploy-preview-1235--happy-galileo-ea79c4.netlify.app/elements/pfe-clipboard/demo/) 

### Testing instructions

1. Open the pfe-clipboard demo page in your browser.
2. In your url bar, append a unique query parameter`?testing123`
3. Click on one of the 'Copy Url' buttons
4. Open a separate browser tab and paste your clipboard into the url
5. You should see the url of the original demo page as well as the unique query parameter
- Example: `https://deploy-preview-1235--happy-galileo-ea79c4.netlify.app/elements/pfe-clipboard/demo/?testing123`

#### Browser requirements

Your component should work in all of the following environments:

- [x] Latest 2 versions of Edge
- [x] Internet Explorer 11 (should be useable, not pixel perfect)
- [x] Latest 2 versions of Firefox (one on Mac OS, one of Windows OS)
- [x] Firefox 68 (or latest version for Red Hat Enterprise Linux distribution)
- [x] Latest 2 versions of Chrome (one on Mac OS, one of Windows OS)
- [x] Latest 2 versions of Safari
- [x] Android mobile device (such as the Galaxy S9)
- [ ] Apple mobile device (such as the iPhone X)
- [x] Apple tablet device (such as the iPhone Pro)

### Ready-for-merge Checklist

Check off items as they are completed.  Feel free to delete items if they are not applicable.

- [x] Expected files: all files in this pull request are related to one request or issue (no stragglers or scope-creep).
- [x] Tests have been updated to cover these changes.
- [x] Browser testing passed.
- [x] Repository compiles and tests pass.
- [x] Changelog updated (not needed for documentation updates).
- [x] Documentation (README.md, WHY.md, etc.) updated or added.
- [ ] Link to the demo recording: []()
- [x] Approved by designer.

### Merging

Please **squash** when merging and ensure your commit message uses [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) formatting.

**Be sure to share your updates with the [patternfly-elements-contribute@redhat.com](mailto:patternfly-elements-contribute@redhat.com) mailing list!**

PR #1289, opened 2021-01-21
Us267056 pfe clipboard  copied duration

heyMP:
This pull request adds the `copied-duration` setting to pfe-clipboard.

## Component name

pfe-clipboard

### Related issue

- https://github.com/patternfly/patternfly-elements/pull/1235#discussion_r556751274

### DIFF
--- a/elements/pfe-clipboard/src/pfe-clipboard.js
+++ b/elements/pfe-clipboard/src/pfe-clipboard.js
@@ -42,7 +42,6 @@ class PfeClipboard extends PFElement {
       copiedDuration: {
         title: "Copied Duration",
         type: Number,
-        attr: "copied-duration",
         default: 3
       },
       role: {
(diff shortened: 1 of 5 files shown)


elements/pfe-clipboard/src/pfe-clipboard.js
```diff
@@ -39,6 +39,12 @@ class PfeClipboard extends PFElement {
         attr: "no-icon",
         observer: "_noIconChanged"
       },
+      copiedDuration: {
+        title: "Copied Duration",
+        type: Number,
+        attr: "copied-duration",
```
castastrophe, line 45: This is the default for the attribute name so you won't need to define it here.  Since you used camel case, the properties parsing will automatically assumed `copied-duration` is the name of the attribute.

heyMP: Noice! Removed.  I'll remove the one on no-icon as well.

PR #1290, opened 2021-01-22
feat: added vars to control padding on host and margin on icon

heyMP:
Fixes: https://github.com/patternfly/patternfly-elements/pull/1235#discussion_r556761775


castastrophe (approved):
This looks great!

PR #1292, opened 2021-01-23
doc: Updating function docblock of this.getSlot

wesruv:
Previous comment seemed off, suggesting an update that I think is clearer


wesruv:
cc @castastrophe || @mwcz

castastrophe:
They seem to be describing the same thing. Can you clarify what you find off?

elements/pfelement/src/pfelement.js
```diff
@@ -230,10 +230,10 @@ class PFElement extends HTMLElement {
   }
 
   /**
-   * Returns an array with all the slot with the provided name defined in the light DOM.
-   * If no value is provided (i.e., `this.getSlot()`), it returns all unassigned slots.
+   * Given a slot name, returns elements assigned to the slot.
```
castastrophe, line 233: Please add back the array reference somewhere in here as that is significant since it is using a spread operator to specifically return an array.

elements/pfelement/src/pfelement.js
```diff
@@ -230,10 +230,10 @@ class PFElement extends HTMLElement {
...
+   * If no value is provided (i.e., `this.getSlot()`), it returns all children without a slot attribute.
```
castastrophe, line 234:
```suggestion
   * If no value is provided (i.e., `this.getSlot()`), it returns all children not assigned to a slot.
```

The assignment is the correct terminology when referring to slots.  Suggest adding that back in.

castastrophe:
@wesruv This has been hanging out about a month.  Do you need any help making the requested changes or have any questions about the feedback?

castastrophe:
@wesruv Following up again.  Should I close this out if you don't think you'll be able to make code review updates?

castastrophe:
@wesruv This PR is getting a bit stale, can you respond to code review or close out the PR if it's no longer relevant?

elements/pfelement/src/pfelement.js
wesruv: I was thinking that may not be easy to understand for folks new to web components, how do you feel about having a parenthetical?
```
* If no value is provided (i.e., `this.getSlot()`), it returns all children not assigned to a slot (without a slot attribute).
```

wesruv:
Holy crap, sorry I missed all of this! Made some updates based on feedback, is there a reason we don't have JSDoc-ish stuff like `@return` or `@param`?
` * @returns [array] An array of DOM Elements`

castastrophe:
@wesruv re: JSDoc -> https://github.com/patternfly/patternfly-elements/pull/1520

castastrophe (approved):

PR #1307, opened 2021-01-26
fork: Attempt first pass at the "one map to rule them all

castastrophe:
First pass at converting to use 1 mixin and 1 map


elements/pfe-primary-detail/demo/pfe-primary-detail.story.js
```diff
@@ -8,7 +8,7 @@
 // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
 import { storiesOf } from "@storybook/polymer";
-import * as bridge from "@storybook/addon-knobs/polymer";
+import * as bridge from "@storybook/addon-knobs";
```
castastrophe, line 11: This just patches a compile bug - Wes is working on a patch for it now

elements/pfe-sass/maps/_typography.scss
```diff
@@ -81,3 +81,148 @@ $pf-type-sizing--titles: (
   title--m-sm--FontSize:    #{pfe-var(FontSize--sm ) },   // 14px
   title--m-xs--FontSize:    #{pfe-var(FontSize--xs ) },   // 12px
 );
+
+$pf-typography-sets: (
```
castastrophe, line 85: This is the money-maker right here.  I'd like to ideally get this information out of a design token someday which would let us sprinkle the goodness in the library. cc @dcaryll @coreyvickery

elements/pfe-sass/maps/_typography.scss
```diff
@@ -81,3 +81,148 @@ $pf-type-sizing--titles: (
...
+  title: (
+    "6xl": (
```
castastrophe, line 87: The quoting is important here since it starts with a number.  If Sass sees the number first, it assumes type = Number and for some reason ignores the "xl" bit.  If it's quoted, all is good.

elements/pfe-sass/maps/_typography.scss
```diff
@@ -81,3 +81,148 @@ $pf-type-sizing--titles: (
...
+      FontFamily: map-get($pfe-typography-base, font-family--heading),
+      FontSize:   map-get($pf-type-sizing--titles, title--m-6xl--FontSize), // alt: could use pfe-var(FontSize--6xl ) directly instead of abstracting it above
```
castastrophe, line 89: We can keep the references in separate maps if you prefer or, if you don't think there's value in it, we can migrate them directly into this map too.

elements/pfe-sass/maps/_typography.scss
```diff
@@ -81,3 +81,148 @@ $pf-type-sizing--titles: (
...
+    ),
+    // @TODO Update this to match above; this is a temp test for the pfe-text mixin
+    md: (
+      FontSize: pfe-var(content--FontSize)
+    ),
+    sm: (
+      FontSize: pfe-var(content--small--FontSize)
+    )
+  ),
+  /// The variable stack includes a hook just for headings when they appear in the
```
castastrophe, line 174: This was such useful information so I wanted to keep that somewhere so I added it here where the hook shows up now

elements/pfe-sass/maps/_typography.scss
```diff
@@ -81,3 +81,148 @@ $pf-type-sizing--titles: (
   title--m-sm--FontSize:    #{pfe-var(FontSize--sm ) },   // 14px
   title--m-xs--FontSize:    #{pfe-var(FontSize--xs ) },   // 12px
 );
+
+$pf-typography-sets: (
+  title: (
+    "6xl": (
+      FontFamily: map-get($pfe-typography-base, font-family--heading),
```
starryeyez024, line 88: I'm putting notes here so I can remember things I want to ask you, but we can chat about it verbally too :) 
I'm wondering why map-get here vs the pfe-var function? Is it because of order of operations, or that its one map referencing another and pfe-var leans on maps, so it wont work?

elements/pfe-sass/maps/_typography.scss
```diff
@@ -81,3 +81,148 @@ $pf-type-sizing--titles: (
...
+  /// Components can't control the font size of block level elements like HTML headings 
+  /// when they are in the light DOM.  A component _could_ redefine the variables for 
+  /// standard headings by using the variable hooks such as: `--pf-c--content--h4--FontSize` 
+  /// but it can't know if the light DOM might contain an H1, H2, H3, etc. it shouldn't
+  /// redefine the variables for all possible headings. Instead, there is one dedicated
+  /// variable for this purpose. 
+  content: (
+    h1: (
+      FontFamily: map-get($pfe-typography-base, font-family),
+      FontSize:   var(--pf-c--FontSize, pfe-var(content--h1--FontSize)),
```
starryeyez024, line 186: thinking aloud: originally I think I wanted the `--pf-c--FontSize` var to only appear sometimes in the stack... but if its always empty by default, maybe this is beneficial to have it *always* in the stack. The only potential downside is developers not knowing when to leverage it...

elements/pfe-sass/maps/_typography.scss
```diff
@@ -81,3 +81,148 @@ $pf-type-sizing--titles: (
...
+      content-spacer: map-get($pfe-vars, content-spacer--heading--sm), // 16px
+    ),
+    h6: (
+      FontFamily: map-get($pfe-typography-base, font-family),
+      FontSize:   var(--pf-c--FontSize, pfe-var(content--h6--FontSize)),
+      LineHeight: map-get($pfe-typography-base, line-height--heading),
+      FontWeight: map-get($pfe-typography-base, font-weight--normal),
+      content-spacer: map-get($pfe-vars, content-spacer--heading--sm), // 16px
+    ),
+  )
```
starryeyez024, line 226: Overall I think this is cool, I'm really wanting to know what the output is though. Like will this print , in this order, these variables in the stack?
 - component `--pf-c--FontSize`
 - content block h6 `--pf-c-content--h6--FontSize `
 - global t-shirt size  `--pf-global--FontSize--sm`
 - fallback value  `0.875rem;`

elements/pfe-sass/maps/_typography.scss
castastrophe: 👍 I didn't want to delete the new maps you built so I referenced them instead but I love the idea of just 1 map that is the source of truth

elements/pfe-sass/maps/_typography.scss
castastrophe: The mixin really handles a lot about the stack, my goal was for the map to contain the data points and the mixin to assemble how it's being output and which properties to use.

elements/pfe-sass/maps/_typography.scss
starryeyez024: If we moved all the variables into one map, I think it makes it harder to use `map-get` or `pfe-var` to use some values to populate others. Which is how I ended up with multiple maps.

My question above though, was more around use of the `map-get` function vs the `pfe-var` function, if there was a benefit of one over the other?

castastrophe: Have you checked out `map-deep-get`?  I don't think map-get is difficult if you have a map-deep-get set up. `map-deep-get($pf-typography-sets, title, 6xl, FontSize)` -> it almost reads like a sentence.

castastrophe: You can use pfe-var or map-get.  Since they're in the same file, map-get just seems more intuitive to me and requires less processing.  pfe-var assumes you don't know what map the data you want is in and checks several for you.

starryeyez024: Cool, that makes sense :)

PR #1318, opened 2021-02-01
feat: set aggressive threshold on IntersectionObserver, sort active i…

heyMP:
## Component name

pfe-jump-links

## Description of changes

The threshold for Intersection Observer can take an array of increments that determine when the Observer should fire a new update.  I set the threshold to fire on increments of 1 percent change.

Then I'm sorting the active items by the intersection observers value of how much of the element is visible.

### DIFF
--- a/elements/pfe-jump-links/src/pfe-jump-links-panel.js
+++ b/elements/pfe-jump-links/src/pfe-jump-links-panel.js
@@ -319,21 +319,13 @@ class PfeJumpLinksPanel extends PFElement {
       // We want only sections that are visible
       .filter(section => section.isVisible)
       // Sort the items by largest intersectionRatio which will be the item
-      // that is the most visible on the sceen.
+      // that is the most visible on the screen.
       // @todo we could take into account other variables like how big the section is on the page
       .sort((a, b) => a.intersectionRatio - b.intersectionRatio)
       .reverse()
       // Now that they are sorted, all we need is the section id
       .map(item => item.id);
 
-    // @todo REMOVE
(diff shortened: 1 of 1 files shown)


elements/pfe-jump-links/src/pfe-jump-links-panel.js
```diff
@@ -307,13 +311,30 @@ class PfeJumpLinksPanel extends PFElement {
...
         if (ref) ref.isVisible = entry.isIntersecting;
+        if (ref) ref.intersectionRatio = entry.intersectionRatio;
       }
     });
 
     const ids = Object.values(this.sectionRefs)
+      // We want only sections that are visible
       .filter(section => section.isVisible)
+      // Sort the items by largest intersectionRatio which will be the item
+      // that is the most visible on the sceen.
```
castastrophe, line 322:
```suggestion
      // that is the most visible on the screen.
```

elements/pfe-jump-links/src/pfe-jump-links-panel.js
```diff
@@ -307,13 +311,30 @@ class PfeJumpLinksPanel extends PFElement {
...
       .map(item => item.id);
 
+    // @todo REMOVE
+    ids.forEach(i => {
+      console.log({
+        id: i,
+        percentVisible: this.sectionRefs[i].intersectionRatio,
+        isVisible: this.sectionRefs[i].isVisible
+      });
+    });
```
castastrophe, line 336:
```suggestion
```

castastrophe (approved):

PR #1319, opened 2021-02-01
Cpfed 4000 arrow key handler all redhat menu

kelsS:
## pfe-nav

removed arrow right key features since they are a no go. Added missing variables from PR request that had to be redone and merged into the epic branch.

### Merging


elements/pfe-navigation/src/pfe-navigation--lightdom.scss
```diff
@@ -33,6 +33,14 @@ pfe-navigation {
     }
   }
 
+  // Hide fallback login link if component has mounted
+  [pfelement],
```
castastrophe, line 37: @kylebuch8 We mentioned deprecating the class `.PFElement` at one point in favor of `pfelement` attribute - is that still desired?  I ask because maybe only 1 of those selectors is needed here and the `.PFElement` option can be removed.  Both are added so you're safe either way.

elements/pfe-navigation/src/pfe-navigation.scss
```diff
@@ -1086,6 +1086,30 @@ See mixin blocks before selectors that contain the selector name and the three l
...
+  // Hide account toggle if NOT logged in
+  // .pfe-navigation__account-toggle {
+  //   display: none;
+  // }
+  // // Show login link if NOT logged in
+  // .pfe-navigation__log-in-link {
+  //   display: flex;
+  // }
+
+  .pfe-navigation__account-wrapper--logged-in {
```
castastrophe, line 1101:
```suggestion
  &--logged-in {
```
^ This is a nice option in Sass and reduces the specificity generated by stacking selectors.

castastrophe: The `[class][class][class]` override for example might not be needed to override specificity if styles are attached to a single class rather than a class stack.

castastrophe, line 1101: I've read some really interesting best practices lately around representing "state" such as "logged-in" via attributes instead of adding additional classes.  This makes it a lot easier to tell what is a state as well vs. an element.  If elements are always represented via class names and states represented as attributes, the entire stylesheet becomes a bit easier to parse.

wesruv: I'd prefer not to do string concatenation in selectors, it makes things harder to find in the codebase, but I agree on the specificty if it isn't necessary.

castastrophe: Can you please share more about what's wrong with string concatenation?  It's how ever other component in the system is written and is extremely common practice in Sass.  It's also, I would suggest, when done correctly, a much cleaner and easier way to read and find selectors.  Especially if you don't nest them more than 1 or 2 levels deep.

castastrophe: I will add, it lets you roll up code based on regions in a more logical way so you can easily get to the sections you are interested in for debugging purposes.

Screen Shot 2021-03-19 at 11 38 53 AM

wesruv: Sounds interesting, but this will have to be done separately from this PR since that's a large shift

castastrophe: I'd also suggest that since the other components in the project use attributes to represent state, this is a big deviation to use classes for state instead.  This component needs to follow the coding conventions of the rest of the project if it wants to be merged into the main repo at some point.

elements/pfe-navigation/src/pfe-navigation--lightdom.scss
kylebuch8: We can drop the .PFElement class here. The `.PFElement` class will be dropped in favor of the `pfelement` attribute that's added after upgrade.

elements/pfe-navigation/src/pfe-navigation.scss
wesruv: My issue is if I want to find `.pfe-card__footer`, searching for it with a find won't work. I also prefer to try and keep Sass declaration blocks on the small side so you can tell what styles roll up to, and this encourages larger declaration blocks.

wesruv: It should be done in a separate ticket is my point, thanks for calling it out.

wesruv (approved):

PR #1335, opened 2021-02-08
Cpfed 3933 i18n

TylerMartin5:
@wesruv first PR for i18n work. Currently supporting all languages listed on rhdc. 

**Couple notes:**

- I was using rhdc's translations but it looks like they aren't translating "menu" so we still need that one
- need to discuss language slot (not a blocker for this PR)

### DIFF
--- a/elements/pfe-navigation/src/pfe-navigation.js
+++ b/elements/pfe-navigation/src/pfe-navigation.js
@@ -82,7 +82,7 @@ class PfeNavigation extends PFElement {
   }
 
   static get observedAttributes() {
-    return [`${this.tag}-open-toggle`];
+    return [`${this.tag}-open-toggle`, "lang"];
   }
 
   constructor() {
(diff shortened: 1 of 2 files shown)


elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -823,6 +898,28 @@ class PfeNavigation extends PFElement {
     }
   }
 
+  /**
+   * If lang is not set to "en", translate strings based on object defined in constructor
+   */
+  _translateStrings(lang) {
```
wesruv, line 904: @TylerMartin5 it looks like it's translating even if it is set to en? There's no if statement wrapping `this._translateStrings()` and I don't see anything in this function either?

wesruv:
Everything else looks great!

elements/pfe-navigation/src/pfe-navigation.js
TylerMartin5: @wesruv I think I forgot to update that comment, I'll do that now. I'm not going to add that logic. If "en" is ever passed to the lang attribute it should just read from the object

wesruv (approved):

PR #1428, opened 2021-03-04
Translate user menu

TylerMartin5:
Adds everything necessary for translations. We need to make a request for most of these strings but the groundworks is in place here. To test, update the lang attribute and the correct language code should appear after "Account details".

### DIFF
--- a/elements/pfe-navigation-account/demo/index.html
+++ b/elements/pfe-navigation-account/demo/index.html
@@ -51,7 +51,7 @@
   <body unresolved>
     <h1><code>pfe-navigation-account</code></h1>
 
-    <pfe-navigation-account logout-link="#log-out">
+    <pfe-navigation-account lang="en" logout-link="#log-out">
     </pfe-navigation-account>
 
     <script>
--- a/elements/pfe-navigation-account/src/pfe-navigation-account.js
+++ b/elements/pfe-navigation-account/src/pfe-navigation-account.js
@@ -17,6 +17,10 @@ class PfeNavigationAccount extends PFElement {
     };
   }
 
+  static get observedAttributes() {
+    return ["lang"];
+  }
+
   get templateUrl() {
     return "pfe-navigation-account.html";
   }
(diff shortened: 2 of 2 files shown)


elements/pfe-navigation-account/demo/index.html
```diff
@@ -51,7 +51,7 @@
   <body unresolved>
     <h1><code>pfe-navigation-account</code></h1>
 
-    <pfe-navigation-account logout-link="#log-out">
+    <pfe-navigation-account lang="en" logout-link="#log-out">
```
castastrophe, line 54: Love this!  Another possibility is to perform a lookup to the html tag for the language code so that Drupal can set it without having to put it directly on the component. https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang

Maybe this is a #feature idea to include support for that but it's a great source of truth that requires less mark-up.

elements/pfe-navigation-account/src/pfe-navigation-account.js
```diff
@@ -17,6 +17,10 @@ class PfeNavigationAccount extends PFElement {
     };
   }
 
+  static get observedAttributes() {
+    return ["lang"];
```
castastrophe, line 21: We no longer need to set observedAttributes on the component.  By adding this attribute to the properties object, you can set the default and point it to the observer you want to use.

elements/pfe-navigation-account/src/pfe-navigation-account.js
```diff
@@ -83,6 +87,184 @@ class PfeNavigationAccount extends PFElement {
     this._userData = null;
     this._avatars = [];
 
+    // set default language, overridden by check in connected callback
+    this._lang = "en";
+
+    // Translations
+    this._navTranslations = {
```
castastrophe, line 94: I'd like to talk more as a team about alternate ways to capture content besides embedding it in components.  This makes the component very difficult to reuse in other situations and scopes it specifically for Red Hat-only.  Perhaps passing the data through an API or adding support for a JSON object is a way to go?  Kendall and I have gone down this path in a few different ways and would be happy to consult.  In the past, we've had a lot of consensus around not putting content into a component. cc @starryeyez024

elements/pfe-navigation-account/src/pfe-navigation-account.js
```diff
@@ -98,6 +280,10 @@ class PfeNavigationAccount extends PFElement {
 
   connectedCallback() {
     super.connectedCallback();
+
+    if (this.hasAttribute("lang")) {
```
castastrophe, line 284: The properties system will manage this for you so it won't be necessary any more.  Example:

```
static get properties() {
    return {
      _lang: {    // underscore is required because `lang` is a global HTML attribute
        title: "Language support",
        type: String,
        default: "en",    // <-- Can I recommend `en-US` to be more accurate?
        observer: "_langHandler" // <-- method for managing the update of the lang property
      },
```

castastrophe:
Oops!  Converted the wrong PR to draft!  Sorry!

elements/pfe-navigation-account/src/pfe-navigation-account.js
wesruv: I'd like to decouple needing to upgrade all of PFE & using the new nav. It sounds like this change might be hard to implement in a way that we could support pre-release.

elements/pfe-navigation-account/demo/index.html
wesruv: 💯

elements/pfe-navigation-account/src/pfe-navigation-account.js
wesruv: Would like to keep this prerelease friendly.

elements/pfe-navigation-account/src/pfe-navigation-account.js
castastrophe: If you're planning to publish this component as a 1.x version, I'd disagree.  It really should use the 1.0 PFElement.  We can discuss as a group if you like.

elements/pfe-navigation-account/src/pfe-navigation-account.js
wesruv: Sounds awesome to me, I think this is good enough for the time being. Definitely not the long term solution though.

wesruv:
Going to merge this so we can get this over to CP Elements

PR #1484, opened 2021-03-22
fix: added build debounce, added observer opt-out attribute for dynam…

heyMP:
I added a small debouncer to the build function to prevent it from re-rendering way too often.  I'm not sure that's going to be enough to prevent some dynamic components from rebuilding out of control so I added an attribute that will prevent the re-renders if it contains dynamic content.  Just add `pfe-content-set--ignore-changes` attribute to an element and it will ignore any changes to that element and it's children.


castastrophe:
Going to close this since the slots approach doesn't require parsing the light DOM anymore.

PR #1487, opened 2021-03-24
feat: added getPreviousSibling utility function for getting closest pa…

heyMP:
I added a small utility function that will recursively search for a closest sibling matching a selector.  I'm thinking this will prevent an edge case where some erroneous element will weasel its way in between the header and a panel and break a content set.

It still uses `previousElementSibling` but will iterate until if finds a match.

### DIFF
--- a/elements/pfe-content-set/src/pfe-content-set.js
+++ b/elements/pfe-content-set/src/pfe-content-set.js
@@ -10,6 +10,25 @@ const CONTENT_MUTATION_CONFIG = {
   childList: true,
   subtree: false
 };
+
+/**
+ * Utility function for getting closest previous sibling.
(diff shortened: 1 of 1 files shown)


castastrophe:
Very fancy but this validation is already being handled by `this._isHeader`.  The approach you proposed leaves out h-tag inclusion for pfe-content-set use-cases where they don't have the attributes present.

heyMP:
Oh gotcha, I didn't catch that.  Let me update it to account for that and see if it looks any better.

castastrophe:
I'm going to close this out - _isHeader is already validating the selectors and it's a source of truth for what a "header" is in the content set

PR #1488, opened 2021-03-24
fork: pfe-tabs baseline images

castastrophe:
Splitting out baselines into 3 separate context-specific snapshots to make quick diff'ing easier.  The taller the pages get, the harder they are to see without downloading assets.


package.json
```diff
@@ -64,7 +64,7 @@
     "@storybook/cli": "^5.3.8",
     "@storybook/polymer": "^5.3.8",
     "@storybook/storybook-deployer": "^2.8.1",
-    "@wdio/browserstack-service": "^7.0.0",
+    "@wdio/browserstack-service": "^6.1.12",
```
castastrophe, line 67: @alazzara I had to revert this version bump that I just did the other day because I couldn't get baselines to run at all locally!

elements/pfe-tabs/demo/demo.css
```diff
@@ -41,118 +31,21 @@ html {
...
-.toc ul {
-    font-size: 11px;
-    list-style: disc;
-}
-@media screen and (max-width: 1240px) {
-  .toc {
-      display: none;
-  }
-}
-.light-background {
```
starryeyez024, line 92: Can we bring back the light background color so we can see that the default tabs variant of wind does not include background colors? @castastrophe

elements/pfe-tabs/test/pfe-tabs_e2e.js
```diff
@@ -1,15 +1,24 @@
...
+      removeElements: [
+        $("#toc"),
+        $$(".skip-e2e")
+      ]
+    };
   });
 
-  it("should compare to the baseline", () => {
-    expect(browser.checkFullPageScreen(element)).toBeLessThan(3.1);
+  ["light", "dark", "saturated"].forEach(context => {
```
starryeyez024, line 17: This is elegant! Do you think we lose any development benefits of seeing all the variants on one page? What if we leave the simplified demo/index.html as you have it, and then we add in a new demo/all-combos.html or something, where you can see everything in a glance? @castastrophe

castastrophe: The challenge is that you can't see it at a glance because the page is so tall. You have to download it and zoom in to try to compare. It gets tricky. I was hoping separating them into individual files makes it easier to diff.

elements/pfe-tabs/demo/demo.css
castastrophe: For sure. How about a system color though like lightest? It's just easier to apply using the dropdown.

starryeyez024: This color was chosen because it's not an PFE system color, so its easy to tell that its not part of the components. Lightest is white, which is what it is now. Lighter could work potentially, but I think the earth variant tabs use lightest, so that could blend in too. I don't super care what the hex value is, just as long as it stands out as not part of the tabs themselves.

elements/pfe-tabs/test/pfe-tabs_e2e.js
starryeyez024: For sure, I'm just highlighting the different between AT and development. I'm in favor of this smaller version for AT and VRT, but as a dev working on tabs which have a lot of variants, its nice to not have to recreate all the combos.

castastrophe: I get that but currently there is no difference between AT and the demo pages so we'll have to find a happy in-between.

starryeyez024: pfe-styles has multiple demo files, couldn't this component have a couple too? and just use the index.html as the one AT uses?

castastrophe: AT fails on those other URLs more often.  No idea why but when we've tried to snapshot other pages, it fails more often than passes due to 404 errors.  It will all get a lot easier after Zack's work but for now, we work with what we've got.

starryeyez024: I'm suggesting 
1. `demo/index.html` - use the exact copy you have in this PR, and run AT on this file only
2. `demo/all-combos.html` - devs can navigate to this page and use it during development. The old version of the demo markup lives here. AT does not run on this file

castastrophe: Yes! Great point. I hadn't considered that. I'll do that first thing tomorrow!

Issue #1490, opened 2021-03-25
PFE-Icon is setting local variables overriding author's intentions


wesruv:
Needs to be updated to style the attributes the variables set, instead of setting the variables directly.

PR #1491, opened 2021-03-25
fix: #1490: Make pfe-icon set CSS properties instead of overriding local var

wesruv:
Overriding local vars is overriding author's intentions in some cases, making pfe-icon set the CSS properties the vars effect instead of overriding the vars

### Related issues

- (#1490) PFE-Icon is setting local variables overriding author's intentions

### Preview

- [Icon demo](https://deploy-preview-1491--happy-galileo-ea79c4.netlify.app/elements/pfe-icon/demo/) 

### Testing instructions

1. Go to pfe-icon demo page and make sure nothing has changed, but setting --pfe-icon-Color should change icon color in all circumstances.

#### Browser requirements

- [x] Firefox 68 (or latest version for Red Hat Enterprise Linux distribution)
- [x] Android mobile device (such as the Galaxy S9)
- [x] Apple mobile device (such as the iPhone X)
- [x] Apple tablet device (such as the iPhone Pro)

### Ready-for-merge Checklist

- [ ] Browser testing passed.
- [x] Changelog updated.

### Merging


elements/pfe-icon/src/pfe-icon.scss
```diff
@@ -113,7 +113,9 @@ $sizing: (
 
 @each $level in (critical, important, moderate, success, info, default) {
   :host([color="#{$level}"]) {
-    --pfe-icon--color: #{pfe-var(feedback--#{$level})};
+    filter feFlood {
+      flood-color: #{pfe-var(feedback--#{$level})};
```
castastrophe, line 117: To prevent regressions, I'd recommend:

```suggestion
      flood-color: pfe-local(color, $fallback: pfe-var(feedback--#{$level}));
```

Also just a note that you don't need interpolation on functions when assigning them to css properties, only when assigning them to custom properties.

elements/pfe-icon/src/pfe-icon.scss
```diff
@@ -130,11 +132,14 @@ $sizing: (
...
-    --pfe-icon--Padding:         .5em;
-    --pfe-icon--BorderWidth:     #{pfe-var(ui--border-width)};
-    --pfe-icon--BorderColor:     #{pfe-local(color, $fallback: pfe-var(surface--border))};
+  padding: pfe-local(Padding, $fallback: .5em);
+  border-width: pfe-local(BorderWidth, $fallback: pfe-var(ui--border-width));
+  border-color: pfe-local(color, $fallback: pfe-var(surface--border));
+  background-color: pfe-local(BackgroundColor, pfe-local(color, $fallback: pfe-var(surface--lightest)));
+
+  filter feFlood {
+    flood-color: pfe-local(Color, $fallback: pfe-broadcasted(text));
```
castastrophe, line 141:
```suggestion
    flood-color: pfe-local(color, $fallback: pfe-broadcasted(text));
```

This component is using a lowercase `color` instead because it's the concept of color rather than the literal `color` property in CSS.

castastrophe (changes requested):
The code looks good, just a couple small updates.  Needs a changelog added as well.  Thanks!

elements/pfe-icon/src/pfe-icon.scss
wesruv: Finding that very confusing, it doesn't seem to be documented in the code or the README, is that something someone who understands it can write down the differences between the vars?

castastrophe: Yeah, now that Kyle is close to merging the documentation update, Kendall & I are working on bringing over the write-up but the gist is BEM.  In BEM, a property that is specific to a CSS property uses this format: BackgroundColor vs. background-color or backgroundColor.  It's how you can separate the idea of a color variable from a variable that literally is assigned to the color property in CSS.  Since this variable is assigned to flood-color, the options were: FloodColor or a more abstract naming.  Since you don't really need to know how the inner workings of the SVG function to style this component, `--pfe-icon--color` was chosen.

It will all be written out but for now, the summary is: `--pfe-icon--color` is correct over `--pfe-icon--Color`.

castastrophe: It's also in the README: 

```
For more fine-grained control, `--pfe-icon--color` is available to override the color of a specific icon or sets of icons and will be applied to either the SVG lines or the background of the circle (if circled).
```

wesruv: Oh, damn. I thought it was, but then I checked and skimmed over it. 😵 Resolving convo

wesruv:
Rut-roh, pretty sure this isn't related to this PR, can anyone else duplicate this:

My PR:
image

PatternflyElements.com
image

wesruv:
Well, this PR hasn't caused any regressions at least 😬 

image

image

castastrophe:
@wesruv What environment?  IE11?  Edge?  I know that patternflyelements.com won't work in IE11 because it uses module imports but the icon demo page should: https://patternflyelements.com/elements/pfe-icon/demo/

This is Chrome on Mac:
Screen Shot 2021-03-25 at 1 53 38 PM

castastrophe:
Something's up with the colors in Chrome though.  This is https://deploy-preview-1491--happy-galileo-ea79c4.netlify.app/elements/pfe-icon/demo/ on Chrome Mac:
Screen Shot 2021-03-25 at 1 55 11 PM

castastrophe:
This section might need to be updated as well to not use the `--pfe-icon--color` override approach: https://github.com/patternfly/patternfly-elements/blob/1987edcd2bfa20cdfd40754b0de2e9e879ae0514/elements/pfe-icon/src/pfe-icon.scss#L127

wesruv:
Getting the same issues in browser-stack :(

image

wesruv:
> This section might need to be updated as well to not use the `--pfe-icon--color` override approach:
> 
> https://github.com/patternfly/patternfly-elements/blob/1987edcd2bfa20cdfd40754b0de2e9e879ae0514/elements/pfe-icon/src/pfe-icon.scss#L127

I see it on Linux too, I'm on it

elements/pfe-icon/src/pfe-icon.scss
```diff
@@ -107,13 +107,23 @@ $sizing: (
 // @TODO: Sort out sizing
 @each $label, $size in $sizing {
   :host([size="#{$label}"]) {
-    --pfe-icon--size: #{$size};
+    // Respect var if user set CSS var, but fallback to sizing property map
+    $desired-size: pfe-local(size, $fallback: $size);
+    max-width:   calc(#{$desired-size} + #{pfe-local(Padding)} * 2);
```
castastrophe, line 112: This looks really great, I love the addition of the $desired-color and $desired-size to create that shared source of truth.  Can you tell me more about the addition of padding into the width calculation?

castastrophe: I'm wondering also if, without breaking it out into `PaddingTop`, `PaddingRight` etc. if this will be an issue.  Imagine I've done: `--pfe-icon--Padding: 10px 5px;` in my stylesheet.  How would that work inside this calc?

elements/pfe-icon/demo/index.html
```diff
@@ -125,8 +125,8 @@
 
   <style type="text/css" media="screen">
     body {
-      --pfe-icon--color: orange;
-      --pfe-icon--context: light;
+      /* --pfe-icon--color: orange;
```
castastrophe, line 128:
```suggestion
      /* Icon overrides for testing */
      /* --pfe-icon--color: orange;
```

^ A comment here might be helpful for future investigators

elements/pfe-icon/src/pfe-icon.scss
wesruv: I'm just maintaining what was happening when the variables were updated (which turned out to be much more complicated than I expected). If you look here: https://github.com/patternfly/patternfly-elements/blob/b8a3c6e8ba0e5e97286c9765fee9159f51bd15f0/elements/pfe-icon/src/pfe-icon.scss#L40

That's what I'm maintaining, I think it's because the width is set to the SVG, and the max-width/height are set on the host, so the host has to take the padding into account. I thought about messing with it to see if I could get `box-sizing` or something a little more straightforward to make the same result, but I'm thinking it wouldn't work and I was having enough issues updating the SCSS and not getting a regression, so I just maintained what was there.

castastrophe: Awesome!  Thanks for clarifying.  I'll add an issue to re-address that later.

castastrophe: https://github.com/patternfly/patternfly-elements/issues/1502

castastrophe (commented):
Looks great!  Just needs a changelog

wesruv:
I updated the Changelog, hopefully did that well?

I re-tested everything in IE, and for some reason now we're getting black icons, where before they were blank. Going to consider that a win?
image

wesruv:
Changelog in, retested everything in the different browsers, and got some IE weirdness.

So I deleted a comment about IE, had a comment about IE that icons weren't showing up. Somehow I think taking a screenshot got me icons? Maybe forced a paint?? I swear they weren't there right before I screencapped.

I'm seeing pfe-icon render much slower in IE on patternflyelements.com consistently, but it seems to be ok on my feature branch? No idea why that would be.

Left is feature branch, right is current state:
image

castastrophe:
This component is definitely a bit odd in IE as-is but I think that's a-okay.  Since it's icons, it's not critical that it upgrade fast or perfectly in IE.

castastrophe:
Re: changelog.  The one we need to keep updated is actually the CHANGELOG-1.x.md on the root of the project.  We talked about using local changelogs at one point and aggregating that up into a shared one but... nothing came from that.  So for now, no component-specific changelogs, just adding our entries to the CHANGELOG-1.x.md.

castastrophe:
Can you take a look at the failing tests when you get a chance? It looks like pfe-icon is getting black instead of the colors but I suspect it's how the styles are being queried in the test.

https://github.com/patternfly/patternfly-elements/pull/1491/checks?check_run_id=2247922228

wesruv:
Everything should be good to go, I updated an existing test that checks color change to check `--pfe-icon--color` and added a copy of that test to check if the legacy var `--pfe-icon--Color` works and updated the verbiage so it's more clear what's being tested.

elements/pfe-icon/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+## 1.3.4 ( TBD )
```
castastrophe, line 1: This can be deleted

elements/pfe-icon/test/index.html
```diff
@@ -12,12 +12,12 @@
     // Load and run all tests (.html, .js):
     WCT.loadSuites([
       'pfe-icon_test.html',
-      'pfe-icon_react_test.html',
-      'pfe-icon_vue_test.html',
+      // 'pfe-icon_react_test.html',
```
castastrophe, line 15: The other tests suites are still commented out

wesruv: OH CRAP, good catch.

Sorry, I was debugging something else and forgot to take this back 😬

wesruv:
Ok, noooooow everything should be good to go?

castastrophe:
In running the visual snapshots, I'm seeing a degredation in the fallback for IE11 in the forced stretch section.

Current branch:
Screen Shot 2021-04-08 at 11 47 18 AM

Main branch:
Screen Shot 2021-04-08 at 11 47 33 AM

wesruv:
Thaaaat's weird, it's not doing that on my Win 10 IE11:
image

castastrophe:
Try opening the `/elements/pfe-icon/demo/index_e2e.html` page.  We snapshot that one for visual regression tests rather than the index.html file.

wesruv:
Got it! Thanks :D

wesruv:
For posterity, it looks like IE11 hates a unitless length in a `calc()` in this case it was `calc(64px + 0 * 2)`, `0` being the calculated CSS property of the padding.

_Fortunately_ I was able to switch the `:host` to `box-sizing: content-box` and that made that calc unecessary, w00t! Seems like using that approach is just better? I added `!important` because grid styles in the demo were making it `border-box` and that's a pretty common light DOM style.

elements/pfe-icon/src/pfe-icon.scss
```diff
@@ -31,15 +38,16 @@ $LOCAL-VARIABLES: map-deep-merge($LOCAL-VARIABLES, (
 }
 
 :host {
-  --context: #{pfe-local(context)};  
+  --context: #{pfe-local(context)};
 
   position: relative;
   display:  inline-block;
 
   // Supports sizing while preserving aspect ratio (despite layout)
```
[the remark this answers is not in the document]
castastrophe: Added the additional bit about the content-box model since that is an important part of making this work and we want to make sure no one accidentally removes it in the future.

castastrophe (approved):

PR #1530, opened 2021-04-13
docs: Update pfe-typography documentation in readme

castastrophe:
### Ready-for-merge Checklist

- [x] Browser testing passed.
- [x] Changelog updated.

### Merging


CHANGELOG-1.x.md
```diff
@@ -2,6 +2,7 @@
 
 - [a1e4b67](https://github.com/patternfly/patternfly-elements/commit/a1e4b67ac012f5987e6cddf2cc7b532a135fa989) feat: Add CSS variable support for typography & background colors on tabs
 - [](https://github.com/patternfly/patternfly-elements/commit/) fix: Updated CSS to allow CSS custom properties to override dynamically calculated theme context
```
castastrophe, line 4:
```suggestion
- [84cb0b1](https://github.com/patternfly/patternfly-elements/commit/84cb0b1ea257a33dc28954367e82771bb3e17a52) fix: Updated CSS to allow CSS custom properties to override dynamically calculated theme context
```

starryeyez024:
I was originally thinking this would allow us to remove duplicate lines from the container components, but then people who are only using pfe-card and not loading pfe-base.css would be left out, so we probably want to leave those as-is.

So I'm wondering what use case is this solving for, assuming that only container components should be able to change the context and therefore broadcast new colors to text & links?

castastrophe:
> So I'm wondering what use case is this solving for, assuming that only container components should be able to change the context and therefore broadcast new colors to text & links?

Let me answer it by asking another question, if we have: `color: pfe-var(text);` right now - does it hurt anything to include the empty hook for the broadcast text variable the way we do for the link variables?

starryeyez024:
>  if we have: color: pfe-var(text); right now - does it hurt anything to include the empty hook for the broadcast text variable the way we do for the link variables?  

In principle, sure, I can get on board!

However when I followed the test instructions on the [tabs demo page](https://deploy-preview-1530--patternfly-elements.netlify.app/elements/pfe-tabs/demo/index_e2e.html), it doesn't seem to work. 
https://www.evernote.com/l/AA7t_Ba0fIlNrLl-yDXapto_m7X6C7mf97c

This is a tangent, but the `--context` attribute on the wrapper is not cascading down to the tabs and resetting the value of the `on` attribute to `dark`. But even If I manually reset the value to dark, it *seems* to work as expected, but it's actually getting its color value from the :host selector on the tabs component. 

image


If we back up and just take a boring text element on the page outside of a web component, I'd figure that it should inherit the color property from the body element, and it does. Until you throw in an empty CSS var... and then it doesn't behave as expected. 
https://www.evernote.com/l/AA4flMUeff5Mxrxa7h0BRtWqwEAcUtAJhxk

Here's a simplified example, which shows that the CSS variable within the color property only seems to work when it's directly on the text element, but not on the body or any other wrapper element?? https://codepen.io/starryeyez024/pen/vYxENeB?editors=1100

I was surprised to see this... I'm hoping I'm just tired and there's an easy explanation here. :)

castastrophe:
> However when I followed the test instructions on the tabs demo page, it doesn't seem to work.

Except that it does.  Notice the elements using the pfe classes for styling but that don't exist inside the tab components.  They do react because they are responding to the broadcasted variable in the color property.

What can't respond on the fly is the components because they either don't use broadcasted variables for their text or they are relying on context which can't know that the variable was updated outside the component (there's no observer for CSS variables yet).

elements/pfe-styles/src/pfe-base.scss
```diff
@@ -84,8 +84,8 @@ th {
 body {
   @extend %pfe-content-base;
   /* This variable is intentionally left empty for light DOM within components */
-  font-size: var( --pf-c--FontSize, pfe-var(content--FontSize));
-  color: pfe-var(text);
+  font-size: var(--pf-c--FontSize, pfe-var(content--FontSize));
+  color: var(--pfe-broadcasted--text, #{pfe-var(text)});
```
castastrophe, line 88:
```suggestion
  color: pfe-var(text);
```

castastrophe:
@starryeyez024 I pulled out the code updates since we don't have consensus and just left the documentation updates.  Thanks for the review if you have time!

starryeyez024 (approved):
Listen; Grok The Music

PR #1531, opened 2021-04-13
chore: Update package with clean tasks

castastrophe:
Adds a few clean tasks and incorporates their use before build and e2e testing.

### Testing instructions

- [x] `npm run e2e` -> should clean out any previous snapshot images in vrt-snapshots before running the tests

New commands:
- [x] `npm run clean` -> should remove the node_modules folder and the node_modules and package-lock.json at the element level but NOT delete the package-lock at the root
- [x] `npm run clean:components` -> should delete the dist and _temp directories in all components
- [x] `npm run clean:e2e` -> should delete the vrt-snapshots content
- [x] `npm run clean:all` -> should run <s>`npm run clean`</s>, `npm run clean:components`, `npm run clean:e2e` in parallel

### Ready-for-merge Checklist

- [x] Changelog updated.

### Merging


package.json
```diff
@@ -27,10 +27,13 @@
...
     "build": "node scripts/build.js",
     "postbuild": "rm -rf examples/node_modules && mkdir examples/node_modules && cp -r node_modules/@webcomponents node_modules/requirejs examples/node_modules/",
-    "clean": "rm -rf node_modules package-lock.json elements/*/{node_modules,package-lock.json,_temp} test/vrt-snapshots",
+    "clean": "rm -rf node_modules elements/*/{node_modules,package-lock.json}",
+    "clean:components": "rm -rf elements/*/{dist,_temp}",
+    "clean:e2e": "rm -rf test/vrt-snapshots",
     "bootstrap": "lerna bootstrap --hoist",
     "deploy-docs": "bash scripts/deploy-docs.sh",
     "release": "bash scripts/release.sh",
+    "preinstall": "npm run clean",
```
zhawkins, line 36: How much does this increase the time it takes to npm install? I guess my only concern would be making every install a clean install when you might only need a dependency update or something.

Another question is if you were to npm install a single module would it clean everything (and not install all the other dependencies)?

I think I'd prefer to keep this a separate script that you could use as needed rather than forcing it every time. 🤔

package.json
```diff
@@ -27,10 +27,13 @@
     "prebuild": "npm run doc-listing-inject",
     "build": "node scripts/build.js",
     "postbuild": "rm -rf examples/node_modules && mkdir examples/node_modules && cp -r node_modules/@webcomponents node_modules/requirejs examples/node_modules/",
-    "clean": "rm -rf node_modules package-lock.json elements/*/{node_modules,package-lock.json,_temp} test/vrt-snapshots",
+    "clean": "rm -rf node_modules elements/*/{node_modules,package-lock.json}",
```
zhawkins, line 30: It'd be cool if you had a "clean:all" command which would run all the separate clean commands at one time.

zhawkins, line 30: It also looks like you're no longer removing the root `package-lock.json`. Is this intentional?

zhawkins (commented):
It looks good, my main question is if we should be forcing a clean preinstall every time.

package.json
castastrophe: Oh I like that - adding that now

castastrophe: Yeah, the package-lock is committed and is our source of truth for shared versioning for packages - I don't think we should delete it on install

package.json
castastrophe: Yeah, I think you're right.  The ci command is actually failing on this and I'm debugging it.  I think I'll just remove it.

package.json
castastrophe: added

package.json
```diff
@@ -27,7 +27,10 @@
...
+    "clean:components": "rm -rf elements/*/{dist,_temp}",
+    "clean:e2e": "rm -rf test/vrt-snapshots",
+	"clean:all": "npm-run-all --parallel clean clean:components clean:e2e",
```
zhawkins, line 33: Indentation looks a bit weird but I like 👍

zhawkins (approved):

package.json
castastrophe: Yeah why does it look weird

castastrophe: I think it's something about github.  It's fine in the file itself

castastrophe:
@zhawkins I'm a dork - the gulp task already cleans up the components before build. 🤦‍♀️

package.json
```diff
@@ -27,7 +27,10 @@
     "prebuild": "npm run doc-listing-inject",
     "build": "node scripts/build.js",
     "postbuild": "rm -rf examples/node_modules && mkdir examples/node_modules && cp -r node_modules/@webcomponents node_modules/requirejs examples/node_modules/",
-    "clean": "rm -rf node_modules package-lock.json elements/*/{node_modules,package-lock.json,_temp} test/vrt-snapshots",
+    "clean": "rm -rf node_modules elements/*/{node_modules,package-lock.json}",
+    "clean:components": "rm -rf elements/*/{dist,_temp}",
+    "clean:e2e": "rm -rf test/vrt-snapshots",
```
castastrophe, line 32:
```suggestion
    "clean:e2e": "rm -rf test/vrt-snapshots/*",
```

castastrophe:
@zhawkins Q: Do you think that checking out a new branch should run the `npm run clean:components`?  So you don't risk running a preview without rebuilding for that branch?

castastrophe:
Ah branch testing.  I ran through the check list and found a couple small issues so I just pushed up those fixes.  I'm done now I swear. 😬

package.json
```diff
@@ -45,12 +48,13 @@
...
     "live-demo": "npm run dev",
     "pr": "node scripts/open-pr.js",
+    "pree2e": "npm run clean:e2e",
     "e2e": "wdio wdio.conf.js"
   },
   "husky": {
     "hooks": {
       "pre-commit": "npm run test-suite-inject > /dev/null && pretty-quick --staged",
-      "post-checkout": "npm run leftover-check"
+      "post-checkout": "npm run clean:components && npm run leftover-check"
```
castastrophe, line 57: Sanity check me here @zhawkins - this makes sense right?  It would be bad to maintain a build from a different branch of code wouldn't it?

zhawkins: I think that makes sense. We can try it out for a while and see if it's super annoying. 😄

PR #1599, opened 2021-05-06
chore: Bump yo version

castastrophe:
Bump yo version to 4.1.0


castastrophe:
Tested this on the dependabot PR which got a bit messed up

PR #1600, opened 2021-05-06
Issue 1565 list styles migrate to pfe-local

heyMP:
Do not merge!


elements/pfe-sass/extends/_extends.scss
```diff
@@ -57,45 +57,45 @@
...
-  // amount of space between list items
-  --margin: var(--pfe-list--margin, 24px);
-  // font color of the numbers in the circle
-  --Color: var(--pfe-list--Color, rgba(238, 0, 0, 1));
-  // background color of circle
-  --BackgroundColor: var(--pfe-list--BackgroundColor, rgba(238, 0, 0, 0.07));
+// Mixins and extends can't use the LOCAL-VARIABLES map because we would be
+// at risk of overriding an elements map if they call our mixin. We need to
+// use just regular sass variables and pfe-local and specify the $fallbacks.
+@mixin ordered-list-primary {
```
castastrophe, line 63:
```suggestion
@mixin ordered-list-primary($size: 56px) {
```

You can move the $size up to an optional input; that seems like the most likely candidate for customization.

elements/pfe-sass/extends/_extends.scss
```diff
@@ -57,45 +57,45 @@
...
-  --PaddingLeft: var(--pfe-list--PaddingLeft, 0);
+  $size: 56px;
+  // we need to wrap this in a calc function so it can be dynamic.
+  // rgb value of the color theme of the circle
+  $color: (238, 0, 0);
+  $font-size: $pf-global--FontSize--2xl;
+
+  // Then we create the same Sass variables as CSS variables so we can use
+  // them in CSS calc functions
+  --pfe-#{$LOCAL}__gap: #{$gap};
```
castastrophe, line 75: You can delete the variables here - the goal is for these to remain empty, that's why we use the fallback input in the local mixin below.

elements/pfe-sass/extends/_extends.scss
```diff
@@ -57,45 +57,45 @@
...
     // increment the local counter variable
     counter-increment: pfe-list;
     position: relative;
-    // we have to make sure that the li is at least the size
-    // of the circle
-    min-height: calc(var(--size));
+    min-height: pfe-local(size, $region: circle, $fallback: $size);
     // the li content will start after the circle and a defined gap
-    padding-left: calc(var(--gap) + var(--size));
+    padding-left: calc(var(--pfe-#{$LOCAL}__gap) + #{var(--pfe-#{$LOCAL}__circle--size)});
```
castastrophe, line 91:
```suggestion
    padding-left: calc(#{pfe-local(gap)} +  #{pfe-local(size, $region: circle)});
```

elements/pfe-sass/extends/_extends.scss
castastrophe, line 75: `--pfe-list__gap` is not a valid variable name because it specifies a region but not property.  I suspect you wanted `--pfe-list--gap` instead to support gap styles on the overall component.

PR #1602, opened 2021-05-07
chore: Add a tool to open an issue if the build on master breaks

castastrophe:
This action validates the build on master and if it failed, it opens a new issue.


castastrophe:
Validated: https://github.com/patternfly/patternfly-elements/issues/1604

PR #1664, opened 2021-06-01
WIP Code Review for Account Dropdown

wesruv:


elements/pfe-navigation-account/src/pfe-navigation-account.js
```diff
@@ -17,6 +17,8 @@ class PfeNavigationAccount extends PFElement {
     };
   }
 
+  // @todo: why is this being observed?
+  // Nothing is being modified in the attributeChangedCallback
```
wesruv, line 21: @TylerMartin5 this one is for you, I don't think I added lang in observed attributes

Issue #1567, opened 2021-04-29
[fix] hasSlot's array case is unreachable


mwcz:
## Description of the issue

pfelement.js includes [a case statement](https://github.com/patternfly/patternfly-elements/blob/38290ea/elements/pfelement/src/pfelement.js#L242) that expects `typeof` to return `"array"`, but typeof can't return array (only [these values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#description)).

I was bringing a PF element into a project that uses [esbuild](https://esbuild.github.io/) and it warned of the problem.

```
[16:13:16] [snowpack] ! optimizing build...
 > error: Could not resolve "/home/mclayton/projects/chrome-next/snowpack-embedded/build/html/index.js"

 > build/_snowpack/pkg/common/pfelement-a68c903f.js:378:11: warning: The "typeof" operator will never evaluate to "array"
    378 │       case "array":
        ╵            ~~~~~~~

 > build/_snowpack/pkg/common/pfelement-a68c903f.js:1044:36: warning: The "typeof" operator will never evaluate to "array"
    1044 │       else if (typeof selectors === "array" || typeof selectors === "object") return selectors;
         ╵                                     ~~~~~~~
```

The full switch statement in question:

```js
    switch (typeof name) {
      case "string":
        return (
          [...this.children].filter(child => child.hasAttribute("slot") && child.getAttribute("slot") === name).length >
          0
        );
      case "array":
        return name.reduce(
          n =>
            [...this.children].filter(child => child.hasAttribute("slot") && child.getAttribute("slot") === n).length >
            0
        );
      default:
        this.warn(
          `Did not recognize the type of the name provided to hasSlot; this funciton can accept a string or an array.`
        );
        return;
    }
```

Currently, the array case is unreachable since typeof never returns "array", so if `name` is truly an array, the default case is taken.

I'm not sure how important the array case is.  The easiest fix would be to remove the array case entirely, but if it _should_ be reachable, the code could be changed to propertly detect arrays such as with an if/elseif/else where the first two check for `name instanceof String` and `name instanceof Array`.

### Impacted component(s)

- maybe all, maybe none?

### Steps to reproduce

None, since I'm not sure what effect this has.

mwcz:
Fixed by #1666

PR #1666, opened 2021-06-02
chore: remove unreachable array case from hasSlot

mwcz:
## remove unreachable array case from hasSlot

The issue is described in #1567.  It's a fairly low priority issue, however it is blocking me from using PFE with [esbuild](https://esbuild.github.io/) in a new project.

When I looked into it further, I found that hasSlot was intended to accept slot name as a string, or an array of slot names (strings).  However, it's only being called with strings throughout the project.  Because of that, it seemed safe to simply remove the array case, rather than finding a way to make it reachable.  I updated the jsdoc and console messages to reflect that hasSlot now only accepts strings.

My plan is to `npm link` this branch locally, so I can get unblocked.  I figured I'd go ahead and submit it as a PR as well, but I'm totally fine with another solution if the array case is really needed.

### Related issues

- (#1567) [fix] hasSlot's array case is unreachable

### What has changed and why

- hasSlot's behavior has not changed
- the unreachable 'array' case of hasSlot is removed
- jsdoc comments and console logs related to hasSlot's arguments are updated to remove references to array arguments

### Testing instructions

Aside from automated tests, no testing is required, as no behavior has changed.

### Ready-for-merge Checklist

Note: because this isn't changing any behavior, I'm choosing to flag this as a chore, not a fix, so there's no changelog entry.

### Merging


castastrophe:
I know we're not using it, but I'd hate to regress the functionality.  Instead of accepting an array, what if it was a comma-separated list?

mwcz:
> I know we're not using it, but I'd hate to regress the functionality. Instead of accepting an array, what if it was a comma-separated list?

We can certainly make proper arrays work without resorting to a comma-separation, the hasSlot function would just need a different approach to test whether the input is an array, since typeof can't return `"array"`.  I just wanted to go for the minimum viable change: removing the code that was both A) unreachable, and B) unused.  There's no regress in functionality because it already wasn't possible to use the array case.

I did try to "activate" the array case before removing it.  `name.constructor === Array` is one way to check for arrayness, but it can't slot into the existing switch statement, because it's switching on `typeof`.  I'd have had to rewrite the switch as a series of if/elses, but I wasn't comfortable reorganizing the bulk of the function since I'm not that familiar with it, and I didn't see any tests for it.

Just as a side note, here's a way to make single-arg functions like hasSlot work with arrays without having to program support directly:  `["slot1", "slot2"].every(s => this.hasSlot(s))` which is a bit more verbose, but pretty handy.

mwcz:
Come to think of it, the switch statement could switch on `name.constructor` instead of `typeof  name`.  The cases could be `Array`, `String`, etc.  That would be a very straightforward change that would enable the intended array support.

mwcz:
:open_mouth: and somehow, now, esbuild doesn't mind `typeof foo === "array"` and bundles just fine. :facepalm:

elements/pfelement/src/pfelement.js
```diff
@@ -240,11 +240,12 @@ class PFElement extends HTMLElement {
   /**
    * Returns a boolean statement of whether or not that slot exists in the light DOM.
    *
+   * @param {String} name The slot name.
```
castastrophe, line 243:
```suggestion
   * @param {String, Array} name The slot name.
```

elements/pfelement/src/pfelement.js
```diff
@@ -240,11 +240,12 @@ class PFElement extends HTMLElement {
...
    * @example this.hasSlot("header");
    */
   hasSlot(name) {
     if (!name) {
-      this.warn(`Please provide at least one slot name for which to search.`);
+      this.warn(`Please provide a slot name to search for.`);
```
castastrophe, line 248:
```suggestion
      this.warn(`Please provide at least one slot name for which to search.`);
```

castastrophe (approved):

PR #1670, opened 2021-06-03
New nav translation object

wesruv:


elements/pfe-navigation/src/pfe-navigation.js
```diff
@@ -1071,18 +1045,45 @@ class PfeNavigation extends PFElement {
...
+   * Translate search button string
+   */
+  _updateSearchButtonText() {
+    this._searchButtonText.textContent = this.searchButtonTranslation;
+  }
+
+  /**
+   * Translate login string
+   */
+  // _updateLoginText() {
```
wesruv, line 1083: What's up with this commented out fella

elements/pfe-navigation/demo/index.html
```diff
@@ -112,7 +112,7 @@
   <a href="#pfe-navigation" class="visually-hidden skip-link">Skip to navigation</a>
   <a href="#main-content" class="visually-hidden skip-link">Skip to content</a>
 
-    <pfe-navigation id="pfe-navigation" role="banner" debug>
+    <pfe-navigation id="pfe-navigation" role="banner" debug lang="de">
```
wesruv, line 115: Mean to commit this?

wesruv (commented):
A few teensy things

elements/pfe-navigation/demo/index.html
TylerMartin5: I did not. Just testing there

elements/pfe-navigation/src/pfe-navigation.js
TylerMartin5: Ah so I held off on including that until I figure out a workaround. This is to update the translation via attribute but the node isn't available when this runs..

PR #1691, opened 2021-06-24
Docs: Fixing a typo in layout docs

wesruv:


docs/layout/index.md
```diff
@@ -106,7 +106,7 @@ This method allows you to have control over the width and offset of your columns
 | `pfe-l-grid-fill-height` | Allows children elements to fill their container's height completely; equivalent of flex-grow |
 | `pfe-m-gutters` | Adds gutters based on `--pfe-theme--container-spacer` |
 | `pfe-m-all-*[1-12]*-col` | Sets width of children in grid to *[1-12]* columns |
-| `pfe-m-all-*[1-12]*-col-on-[xs-xl]` | Sets width of children to *[1-12]* columns on specified breakpoint *[xs-xl]* |
+| `pfe-m-*[1-12]*-col-on-[xs-xl]` | Sets width of children to *[1-12]* columns on specified breakpoint *[xs-xl]* |
```
castastrophe, line 109: This is a great addition except I think rather than delete this line, we need to add the details for the individual classes.  There are a set of `pfe-m-all-*-col-on-*` classes in the suite.

- `pfe-m-all-1-col-on-xs` sets the width of all children to the columns specified at the breakpoint (sits with the `pfe-l-grid` class) vs.
- `pfe-m-1-col-on-xs` which goes on the child of the `pfe-l-grid` class and sets just that one item to the column width at that breakpoint

castastrophe: Actually, I see now that the `pfe-m-*-col-on-*` classes are described in the section just below this.

castastrophe:
@wesruv Let me know if you have additional questions about these classes or if I can help in any way.  I'm going to close this out since the update is not necessary.

wesruv:
Ooooooooooohhhh, I see now.

I think it'd be helpful if the abbreviations were explained, e.g. what's `pfe-l-` vs `pfe-m-`? At first glance I thought it was large and medium, which combined with poor skimming got me confused.

Adding all of the breakpoint names and the default widths would also be a win.

Sorry for the errant PR 😬

PR #1699, opened 2021-07-02
set environment with attr

TylerMartin5:
Setting the "environment" attribute with point nav-account links to the correct place. Leaving the attribute off defaults to prod. Attribute must be set by each site however they choose to do so.


wesruv:
Closing in favor of https://github.com/patternfly/patternfly-elements/pull/1700

PR #1702, opened 2021-07-07
fix: [pfe-tooltip] forced reflow performance isssues

heyMP:
## Description

When a use hovers over a tooltip very quickly and repetitively it was causing warnings for `forced reflow`.  This is caused when javascript is trying to calculate styles at the same time it's applying them.  This would cause sever performance issues.

The cause is that we are trying to update the position of the tooltip when the it was about to activate.  This would allow the target to potentially move around the screen and the tooltip would always follow it.

## Replicate Issue

- Go here: https://deploy-preview-1406--patternfly-elements.netlify.app/elements/pfe-tooltip/demo/
- Open dev console
- Quickly hover over the different tooltips in quick succession
- You should see a `forced reflow` warning in your console

Screen Shot 2021-07-07 at 1 41 45 PM
  

## Tests
- [ ] Verify it prevents force reflow errors
 - Go here: https://deploy-preview-1702--patternfly-elements.netlify.app/elements/pfe-tooltip/demo/
 - Open dev console
 - Quickly hover over the different tooltips in quick succession
 - There should be no slowdown in how the tooltip renders

### DIFF
--- a/elements/pfe-absolute-position/src/pfe-absolute-position.js
+++ b/elements/pfe-absolute-position/src/pfe-absolute-position.js
@@ -217,7 +217,7 @@ class PfeAbsolutePosition extends PFElement {
       this._updatingPosition = true;
       // ask the manager to update our position over the target element.
       this.__manager.positionElement(this);
-      // we are only going to debounce this for one second.
+      // we are only going to debounce this for 100ms.
       // not doing this will result in a reflow request loop and crash the browser :(
       setTimeout(() => {
         this._updatingPosition = false;
--- a/elements/pfe-absolute-position/src/pfe-absolute-position.js
+++ b/elements/pfe-absolute-position/src/pfe-absolute-position.js
@@ -210,10 +210,21 @@ class PfeAbsolutePosition extends PFElement {
    * @returns {void}
    */
   updatePosition() {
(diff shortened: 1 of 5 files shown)


elements/pfe-absolute-position/src/pfe-absolute-position.js
```diff
@@ -210,10 +210,21 @@ class PfeAbsolutePosition extends PFElement {
...
    */
   updatePosition() {
+    // if we are currently updating the position then ignore this request
+    if (this._updatingPosition) return;
     if (this.__observe === true) {
+      // set the internal state to updating.
+      this._updatingPosition = true;
+      // ask the manager to update our position over the target element.
       this.__manager.positionElement(this);
+      // we are only going to debounce this for one second.
```
kylebuch8, line 220: @heyMP - Not quite one second. Can you change this to 100 ms?

kylebuch8:
@heyMP Can you update the changelog too?

elements/pfe-absolute-position/src/pfe-absolute-position.js
heyMP: 👍

PR #1725, opened 2021-07-26
chore: correct name in citations

castastrophe:
We use a lot of Kitty's tooling and this PR is to correct the citations.


zhawkins (approved):
LGTM!

kylebuch8:
@zhawkins meant to say...
image

PR #1729, opened 2021-07-27
chore: Update accordion baseline

castastrophe:
Bring back the accordion IE11 visual regression test in advance of design updates PR


heyMP (approved):
LGTM!

Issue #1794, opened 2021-10-08
PFE-Clipboard: Copy more than the current URL


wesruv:
I'd like to add the ability to copy arbitrary content that's set to a property of the component, and or set a selector for an element, who's content should be copied.

### Impacted component(s)

- pfe-clipboard

bennypowers:
closed in #1795

PR #1795, opened 2021-10-08
feat: Make it so PFE-Clipboard can copy arbitrary content

wesruv:
Need docs update and tests still, also think we can add a disabled state if it can't find what it should be copying.

For this proof o concept, see the comments I made below, and then check out the demo page here:
https://deploy-preview-1795--patternfly-elements.netlify.app/elements/pfe-clipboard/demo/

### Related issues

- #1794


elements/pfe-clipboard/src/pfe-clipboard.js
```diff
@@ -51,6 +51,10 @@ class PfeClipboard extends PFElement {
         type: Number,
         default: 0,
       },
+      target: {
+        type: String,
+        default: 'url',
+      }
```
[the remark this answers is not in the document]
heyMP: I love this approach.  We could probably use this pattern for other elements. 👍

elements/pfe-clipboard/demo/index.html
```diff
@@ -63,11 +63,21 @@ <h2>Clipboard: with custom icon</h2>
...
       <span slot="text--success">Making some copies!</span>
     </pfe-clipboard>
+    <input type="text" id="textToCopy" value="This text will be copied!!11"></input>
+
+    <h2>Clipboard: Copying text from property</h2>
+    <pfe-clipboard role="button" tabindex="0" target="property" id="propertyCopy">
+    </pfe-clipboard>
+    <script>
+      window.addEventListener('load', function() {
+        document.getElementById('propertyCopy').contentToCopy = '    <h2>Clipboard: with custom text & copying text from element</h2>\n    <pfe-clipboard role="button" tabindex="0" target="#textToCopy">\n      <span slot="text">This will copy the text in the text field below!</span>\n      <span slot="text--success">Making some copies!</span>\n    </pfe-clipboard>\n    <input type="text" id="textToCopy" value="This text will be copied!!11"></input>';
```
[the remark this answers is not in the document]
heyMP: ballin!

elements/pfe-clipboard/src/pfe-clipboard.js
```diff
@@ -162,34 +208,36 @@ class PfeClipboard extends PFElement {
...
    * https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard
    *
    * If unavailable, it will use the legacy execCommand("copy")
    * https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand
    * @async
+   * @param {string} text Text to be copied
    * @return {Promise<string>} url
    */
-  copyURLToClipboard() {
+  copyTextToClipboard(text) {
```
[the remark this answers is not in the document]
wesruv: Sure!

wesruv:
The code is good to go as far as I'm concerned, but I it still needs docs. Wanted to bounce this issue off folks before I wrote the docs:

I added a few disabled states for when:
* `target='property'` and `el.contentToCopy` is not set
  * I added a getter and setter so I don't think there's a chance the component will be disabled if that value is set
* `target='selector'` and element in selector can't be found
  * This one *is not* fool proof. The code checks to see if the selector finds an element on `connectedCallback`, on `DOMContentLoaded`, and `load`, but if the content is injected dynamically the site would have to call `el.checkForCopyTarget()`
  * It also runs `checkForCopyTarget()` when the `target` attribute changes, which helps with the issue in the previous bullet
 
The only other ways of checking to see if the element exists would be heavy... like a MutationObserver on the document 🤢 ...

Do you all think this is good enough? Is there a 'cool kid' way to handle that 'not fool proof' concern that I'm missing?

wesruv:
> The code is good to go as far as I'm concerned, but I it still needs docs. Wanted to bounce this issue off folks before I wrote the docs:
>  ...
> Do you all think this is good enough? Is there a 'cool kid' way to handle that 'not fool proof' concern that I'm missing?

Discussed this with Kyle Buchanan, Michael Clayton, and other CPFED folks, and it feels like checking to see if an element exists isn't worth it. We can _know_ if the property is set, but if an element exists is too hairy, and not a necessary thing to guard against.

elements/pfe-clipboard/src/pfe-clipboard.js
```diff
@@ -91,12 +114,17 @@ class PfeClipboard extends PFElement {
...
+    this.checkForCopyTarget();
   }
 
   disconnectedCallback() {
+    // Clean up after ourselves
     this.removeEventListener("click", this._clickHandler.bind(this));
     this.removeEventListener("keydown", this._keydownHandler.bind(this));
-    this.shadowRoot.removeEventListener("slotchange", this._slotchangeHandler.bind(this));
+    window.removeEventListener('DOMContentLoaded', this.checkForCopyTarget);
+    window.removeEventListener('load', this.checkForCopyTarget);
```
heyMP, line 127: @wesruv I think we're removing these right?

heyMP:
@corey Here are the icon changes: 
https://deploy-preview-1795--patternfly-elements.netlify.app/elements/pfe-clipboard/demo/

- [x] Light theme Icon color: #6a6e73; Hover color: #151515
- [x] Dark theme Icon color: #d2d2d2; Hover color: #fff

I also added a saturated theme where I used the following variables.  I tried to semantically match them to their corresponding dark theme variables but they can be changed.
- icon color: `text--muted--on-saturated: #d2d2d2`
- icon color hover: text--on-saturated #fff`

coreyvickery:
@heyMP Looks great.

elements/pfe-clipboard/src/pfe-clipboard.scss
```diff
@@ -47,16 +53,43 @@ $LOCAL-VARIABLES: (
...
+      // Customize icon color for pfe-icons
+      --pfe-icon--Color: #{pfe-local(icon--Color--hover)};
+      svg {
+        // Customize icon color of fallback svg icon
+        fill: pfe-local(icon--Color--hover) !important;
+      }
+    }
+  }
+}
+
```
heyMP, line 92: Simplified our old focus,hover selectors.

elements/pfe-clipboard/src/pfe-clipboard.scss
```diff
@@ -74,16 +107,14 @@ $LOCAL-VARIABLES: (
...
+  --pfe-clipboard--icon--Color--hover: #{pfe-local(icon--Color--hover--dark)};
 }
 
-// Note: Hover states need to be defined after focus states
-:host(:not([aria-disabled="true"]):hover),
-:host(:not([aria-disabled="true"])) ::slotted(:hover) {
-  --pfe-clipboard--Color: #{pfe-local(Color--hover)};
+:host([on="saturated"]) {
+  --pfe-clipboard--icon--Color: #{pfe-local(icon--Color--saturated)};
+  --pfe-clipboard--icon--Color--hover: #{pfe-local(icon--Color--hover--saturated)};
```
heyMP, line 117: Explicitly adding theme context hooks for icon color.  We don't have to include text color variables because that's automatically covered by broadcast variables.

elements/pfe-clipboard/src/pfe-clipboard.js
wesruv: D'OH!

elements/pfe-clipboard/src/pfe-clipboard.scss
wesruv: 👍🏼

wesruv:
Removed those unnecessary removeEventListeners

heyMP (approved):
LGTM!

Issue #1797, opened 2021-10-12
[meta]: Migrate Components to Lit


bennypowers:
## Components

- [ ] `<pfe-accordion>`
    - [ ] Code
    - [ ] Tests
    - [ ] e2e

- [ ] `<pfe-autocomplete>`
    - [ ] Code
    - [ ] Tests
    - [ ] e2e

- [ ] `<pfe-avatar>`
    - [x] Code
    - [x] Tests
    - [ ] e2e

- [ ] `<pfe-badge>`
    - [x] Code
    - [x] Tests
    - [ ] e2e

- [ ] `<pfe-band>`
    - [ ] Code
    - [ ] Tests
    - [ ] e2e

- [ ] `<pfe-button>`
    - [x] Code
    - [x] Tests
    - [ ] e2e

- [ ] `<pfe-card>`
    - [x] Code
    - [x] Tests
    - [ ] e2e

- [ ] `<pfe-clipboard>`
    - [x] Code
    - [x] Tests
    - [ ] e2e

- [ ] `<pfe-codeblock>`
    - [ ] Code
    - [ ] Tests
    - [ ] e2e

- [ ] `<pfe-collapse>`
    - [ ] Code
    - [ ] Tests
    - [ ] e2e

- [ ] `<pfe-content-set>`
    - [ ] Code
    - [ ] Tests
    - [ ] e2e

- [ ] `<pfe-cta>`
    - [x] Code
    - [x] Tests
    - [ ] e2e

- [ ] `<pfe-datetime>`
    - [x] Code
    - [x] Tests
    - [ ] e2e

- [ ] `<pfe-dropdown>`
    - [ ] Code
    - [ ] Tests
    - [ ] e2e

- [ ] `<pfe-health-index>`
    - [ ] Code
    - [ ] Tests
    - [ ] e2e

- [ ] `<pfe-icon-panel>`
    - [ ] Code
    - [ ] Tests
    - [ ] e2e

- [ ] `<pfe-icon>`
    - [ ] Code
    - [ ] Tests
    - [ ] e2e

- [ ] `<pfe-jump-links>`
    - [ ] Code
    - [ ] Tests
    - [ ] e2e

- [ ] `<pfe-markdown>`
    - [x] Code
    - [x] Tests
    - [ ] e2e

- [ ] `<pfe-modal>`
    - [x] Code
    - [x] Tests
    - [ ] e2e

- [ ] `<pfe-navigation>`
    - [ ] Code
[...]

Issue #1780, opened 2021-09-13
[fix] Pfelement dependency should be pinned for all elements.


heyMP:
## Description of the issue

Our components do not currently point to a pinned version of `pfelement` in their dependencies.  They point to the latest minor release.  In other words, they use a carrot(`^`) when defining the dependency.  What you end up with is that there will be a chance that you end up with a version mismatch between the element and pfelement.

For example

```package.json
  "dependencies": {
    "@patternfly/pfe-card": "1.10.0"
  },
```

Will result in the following packages being installed:

```package-lock.json
    "node_modules/@patternfly/pfe-card": {
      "version": "1.10.0",
      "resolved": "https://registry.npmjs.org/@patternfly/pfe-card/-/pfe-card-1.10.0.tgz",
      "integrity": "sha512-/RETcKoE+LGUQRQ80GbJ7IVQxRZ3UU81Apl9NAQNIrrcKEXoQaoQAilv6uQQ2oaXzLvo9HWFvhuTVqDNJcUmYA==",
      "dependencies": {
        "@patternfly/pfelement": "^1.10.0"
      },
      "engines": {
        "node": ">=10 <13"
      }
    },
    "node_modules/@patternfly/pfelement": {
      "version": "1.11.1",
      "resolved": "https://registry.npmjs.org/@patternfly/pfelement/-/pfelement-1.11.1.tgz",
      "integrity": "sha512-qy4saYS+YCecmYzdlp+JJX3PfFg/ksQHblr0MpYKhOueR/KsGm/tjCYoqbyRMqRd4eK+PJf9t1p0506j3ZOAZw==",
      "engines": {
        "node": ">=10 <13"
      }
    },
```

### Steps to reproduce

1. Create a fresh directory
2. cd into the directory and run `npm init -y` to create a package.json file.
3. Run `npm i --save @patternfly/pfe-card@1.10.0`
4. Check `package-lock.json` and verify that `@patternfly/pfelement` is at version `1.11.1`

### Proposed Solution

We should pin the `@patternfly/pfelement` dependency to the save version as the component.  This will prevent version mismatches.

bennypowers:
From pfe-2 this is no longer relevant

Issue #1744, opened 2021-08-05
`engines` blocks yarn install, adds warning to npm install


mwcz:
## Description of the issue

In our elements' package.jsons, the `engines` setting for node is `>=10 <13`, which equates to node 10 and 12.  As far as I remember, the intent of this restriction was to be a guardrail for PFE contributors, because node-sass won't run on node 14 and above.  By restricting node to 10 and 12, contributors get a clear error about their node version when attempting to run the build.

In an ideal world, this would be better accomplished through a `devEngines` flag, because the only thing we wanted to restrict was what node version is used when _building_ PFE.  Unfortunately no `devEngines` flag exists.

Meanwhile, the `engines` flag also limits what node versions _users_ of PFE can be running.  For instance, I'm using node 16, and I can't install PFE components into a project because my node verison is outside the range.  Because our components run in the browser and not in node, the `engines` flag is less useful.

I think nvm and the `.nvmrc` we already have do a good job of being a guardrail for contributors.  As for the `engines` flag, I think we should either:

&nbsp;&nbsp;**a**. Remove `engines` entirely and rely on nvm
&nbsp;&nbsp;**b**. Keep `engines` but expand it to be very generous, such as `>=10`

Thoughts, opinions, ideas?

### Impacted component(s)

- All of them

### Steps to reproduce

With [nvm](https://github.com/nvm-sh/nvm#readme) installed.

1. `nvm install 14 && nvm use 14`
2. `npx yarn add @patternfly/pfelement`
3. `npm i @patternfly/pfelement`

### Actual behavior

npm successfully installs @patternfly/pfelement, but issues a warning:

```
npm WARN notsup Unsupported engine for @patternfly/pfelement@1.10.1: wanted: {"node":">=10 <13"} (current: {"node":"14.17.1","npm":"6.14.13"})
```

yarn refuses to install @patternfly/pfelement

```
$ npx yarn add @patternfly/pfelement
yarn add v1.22.11
info No lockfile found.
[1/4] Resolving packages...
[2/4] Fetching packages...
error @patternfly/pfelement@1.10.1: The engine "node" is incompatible with this module. Expected version ">=10 <13". Got "14.17.1"
error Found incompatible module.
```

Note, yarn does have an `--ignore-engines` flag, but using it is problematic because it has to be included on every single dependency installation, even ones unrelated to PFE.  For instance, if I want to `yarn add lodash`, I must include `--ignore-engines`.

### Expected behavior

[...]

Issue #1480, opened 2021-03-22
[feat] pfelement | updateComplete lifecycle callback needed


castastrophe:
## Description of the requested feature

Adding an `updateComplete` lifecycle callback similar or compatible with lit-element: https://github.com/Polymer/lit-element/blob/c59e2fec1d93dfd3192d5a9338a03f7f8da5129d/src/lib/updating-element.ts

```
 /**
   * Returns a Promise that resolves when the element has completed updating.
   * The Promise value is a boolean that is `true` if the element completed the
   * update without triggering another update. The Promise result is `false` if
   * a property was set inside `updated()`. If the Promise is rejected, an
   * exception was thrown during the update.
   *
   * To await additional asynchronous work, override the `_getUpdateComplete`
   * method. For example, it is sometimes useful to await a rendered element
   * before fulfilling this Promise. To do this, first await
   * `super._getUpdateComplete()`, then any subsequent state.
   *
   * @returns {Promise} The Promise returns a boolean that indicates if the
   * update resolved without triggering another update.
   */
  get updateComplete() {
    return this._getUpdateComplete();
  }
```

Benefit of this is that the open-wc testing tools look for this lifecycle callback in running the tests: https://open-wc.org/docs/testing/helpers/#timings

```
By default fixture awaits the elements "update complete" Promise.

- for lit-element that is el.updateComplete;
- for stencil that is el.componentOnReady();

If none of those specfic Promise hooks are found, it will wait for one frame via await nextFrame().
Note: this does not guarantee that the element is done rendering - it just waits for the next JavaScript tick.
```

### Impacted component(s)

- pfelement

bennypowers:
From pre-2 this is not longer relevant

PR #1804, opened 2021-10-25
⚡ PatternFly Elements 2.0: Lit, TypeScript, pfe-tools, and more

bennypowers:
# 🔥 🐵 Migrate to Lit ⚡ 🚧  

This PR aims to massively overhaul the PatternFly Elements codebase: updating it, reducing its complexity, and improving both the user's and the contributor's experience

## Sticky Notes

- [🌍 Netlify Build](https://github.com/patternfly/patternfly-elements/pull/1804#issuecomment-961901675)
- [🛒 Changesets](https://github.com/patternfly/patternfly-elements/pull/1804#issuecomment-1008307428)
- [🔦 Lighthouse Report](https://github.com/patternfly/patternfly-elements/pull/1804#issuecomment-1023303813)
- [👚 Commitlint Status](https://github.com/patternfly/patternfly-elements/pull/1804#issuecomment-1023966136)

### Adopt TypeScript

> **Gains**: code correctness, API documentation, developer ergonomics

TypeScript allows us to find and eliminate bugs before they hit production. In the process of converting components, several of these types of problems were identified and ameliorated. TypeScript also lets us develop faster and with confidence, as the objects in our project and their behaviour are now are well-defined. As well, adopting TypeScript is an a11y feature for contributors, as the IDE features of ts language server let them access information about the objects 'under their cursor' faster, without having to page back and forth between docs (as much) 

### Adopt Custom Elements Manifest

> **Gains**: standards, complexity, bundle size, tooling

This PR removes the custom manifest schema developed for PFE in favour of the [custom elements manifest](https://github.com/webcomponents/custom-elements-manifest/), a community standard for documenting javascript packages, particularly geared to web components. By documenting our components in the manifest, tooling such as docs-site generators and IDEs will be able to understand the contents of our components and their abilities without requiring us to ship that code over the wire to end users. 

In particular, our 11ty docs site now uses the generated manifests to generate docs pages.

### Removed `PFElement` base class and package in favour of `LitElement`

> **Gains**: maintainability, performance, complexity

The `PFElement` base class implemented a wide variety of features, some related to DOM templating and component lifecycle, others related to analytics, still more related to 'colour context'. The DOM templating and component lifecycle features are the main reason we're switching to Lit - they are well maintained there, so removing that code from PFE frees us up to 'move up the stack' and focus on UI problems. Other features like colour context, analytics events, etc, could therefore be replaced by [reactive controllers](https://lit.dev/docs/composition/controllers/) and [typescript decorators](https://www.typescriptlang.org/docs/handbook/decorators.html), allowing component authors to add them to elements a la carte. This will result in smaller javascript payloads for some pages

### Rename slots and events

> **Gains**: API surface, best practices

Slot names like `pfe-card--header` have been renamed to simply `header`. This is in line with web components community best practices and conventions and simplifies element APIs, amking them easier to use.
The old slots will still work, until the next major version, so upgrading should be smooth.

### Rename events and deprecate `CustomEvent`

> **Gains**: API surface, semantics

`CustomEvent` is a throwback to a time when the spec did not allow us to subclass `Event`. Now that we can, we can prefer to define our own event classes. This lets us validate event's data at run time, use instanceof to validate them in listeners, and generally improve API readability. Events are still composed, so analytics code should be able to pick them up.

As well, events like `pfe-clipboard:copied` are renamed (in their new form) to `copied`, for simplicity, and framework compatibility.

The old CustomEvent-based events are still used, under their old names, until the next version, to aid in migration

### Adopt a 'buildless' development workflow

> **Gains**: Developer's time

Contributors no longer need to keep a file watcher to re-compile components when working on elements. Instead, source files (typescript and sass) are compiled on-the-fly by the dev server. Just install dependencies and run `npm start`

### Update to `dart-sass`

> **Gains**: build performance, language features, security

`node-sass` is deprecated. Updating to dart-sass improved performance and made new language features available.

### Rewrite the element generator

> **Gains**: fewer dependencies

Removed the dependency on yeoman and rewrote the element generator to generate LitElements for PFE. once published, contributors will be able to `npm init @patternfly/element`

### Add `@patternfly/pfe-tools`

> **Gains**: Simplify root repository, unblock template repo

Users wishing to develop their own set of elements based on patternfly will be able to clone a template repo and import tooling helpers from `@patternfly/pfe-tools`. This will include things like
- dev server and testing configuration
- build scripts and config
- custom-elements manifest analyzer config

### Add Linting Rules

> **Gains**: code standards, readability, consistency

This PR adds eslint rules for PFE. It's recommended to install an ESLint plugin for your editor and enable "fix on save"

### Remove `<pfe-navigation>`

The decision was taken to move pfe-navigation downstream to red hat's design system. This frees up PFE to concentrate on lower-level components

### Add Lighthouse Testing in CI

> **Gains**: performance and a11y visibility

PRs will now be tested using lighthouse against each element's demo page and the results formatted and printed to the PR thread

### Commitlint Reporting

> **Gains**: repo consistency

Commitlint errors and warnings will now be reported to the PR thread. 

## Testing instructions

0. Step 0: Edit the checklist in the PR description for codereview
1. Open a tab to patternflyelements.org docs to the element you’re reviewing
2. Open a tab to the deploy preview docs to the element you’re reviewing 
3. Clone the repo and/or fetch/reset the branch
    ```bash
    git fetch && git reset --hard origin/feat/pfe-tools
    ```
4. Install deps and run the dev server
    ```bash
    npm ci && npm start
    ```
5. Do the new docs accurately reflect the new version?
6. Do the changes in the new version make sense?
7. While reviewing, requested changes should go  in github PR review comments

## Review Work Assignments

Check the box and tag your GitHub user to claim an element for review.

### Core

- [x] pfe-sass 
- [x] pfe-styles 
- [x] pfe-core (@kylebuch8, @mwcz, @heyMP)

### Elements

- [x] pfe-accordion(@wesruv)
- [x] pfe-autocomplete(@heymp)
- [x] pfe-avatar(@zeroedin)
- [x] pfe-badge(@zeroedin)
- [x] pfe-band (@heyMP)
- [x] pfe-button(@zeroedin)
- [x] pfe-card (@heymp)
- [x] pfe-clipboard (@heyMP)
- [x] pfe-codeblock(@wesruv)
- [x] pfe-collapse (@heymp)
- [x] pfe-content-set (@wesruv)
- [x] pfe-cool-element (@bennypowers is this real?)
- [x] pfe-cta (@kylebuch8)
- [x] pfe-datetime (@heyMP)
- [x] pfe-dropdown (@heyMP)
- [x] pfe-health-index (@nouveau)
- [x] pfe-icon (@wesruv)
- [x] pfe-icon-panel (@wesruv)
- [x] pfe-jump-links (@heyMP)
- [x] pfe-markdown (@zeroedin)
- [x] pfe-modal (@Nouveau)
- [x] pfe-number (@zeroedin)
- [x] pfe-page-status (@Nouveau)
- [x] pfe-primary-detail (@wesruv, @zeroedin)
- [x] pfe-progress-indicator (@zeroedin)
- [x] pfe-progress-steps (@heyMP)
- [x] pfe-readtime (@heyMP)
- [x] pfe-select (@zeroedin)
- [x] pfe-tabs (@wesruv)
- [x] pfe-toast (@heyMP)

## Ready-for-merge Checklist

- [x] Commit history rebased and cleaned up
- [x] Expected files: all files in this pull request are related to one request or issue (no stragglers ~~or scope-creep~~).
- [x] Browser testing passed.
- [x] Changelog updated.
- [x] Documentation (README.md, WHY.md, etc.) updated or added.
- [ ] Approved by designer.

## Related issues

- (#1797) lit2
- (#1780)
- (#1480)
- (#1744)
- (#886)
- (#1813)

## Preview

- [Deploy Preview Link](https://deploy-preview-1804--patternfly-elements.netlify.app/) 

## Merging

Please **rebase** when merging and ensure your commit messages use [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) formatting.


bennypowers:
Heads up to everyone following this branch, I just *rebased and force pushed* in order to organize the commit history. It should be much easier to follow the progression of commits now, but it does mean you should fetch and reset your local branch

bennypowers:
@alazzara @heyMP @kylebuch8 @mwcz @starryeyez024 @zhawkins (apologies if I missed anyone)

This is now *basically* ready to go

The last things to do (i think) are
- [ ] update the generator to register tsconfig paths for new components
- [ ] update docs to explain how to add ts references when siblings depend on each other
- [ ] Fix the e2e test to run in playwright on browserstack.
    Let's maybe defer that to a subsequent PR?
- [ ] Update CI/CD pipeline
    I have to understand the release branch workflow. Perhaps we can simplify here and use GitHub releases.
    We should also turn on lerna 'prerelease' by setting a `next` dist tag

tools/create-element/templates/element/custom-elements-manifest.config.js
```diff
@@ -0,0 +1,5 @@
+import { pfeCustomElementsManifestConfig } from '@patternfly/pfe-tools/custom-elements-manifest.js';
+
+export default pfeCustomElementsManifestConfig({
+  globs: ['<%= tagPrefix %>-.ts'],
```
zeroedin, line 4: Glob missing `*`

When running `npm run analyze` on a newly created element it will not match the `pfe-*.ts` file

```suggestion
globs: ['<%= tagPrefix %>-*.ts'],
```

bennypowers: Nice catch!

elements/pfe-primary-detail/pfe-primary-detail.ts
```diff
@@ -0,0 +1,667 @@
...
+          <slot name="details-nav"></slot>
+        </div>
+        <div id="details-wrapper">
+          <div id="details-wrapper__header">
+            <button
+                id="details-wrapper__back"
+                aria-controls="${ifDefined(this.controlsId)}"
+                aria-expanded="${ifDefined(this.breakpoint !== 'compact' ? undefined : String(this.active))}"
+                @click="${this.closeAll}">
+              ${this._renderDetailsWrapperHeading()}
```
wesruv, line 165: There seems to be a race condition where this isn't being set before render... not sure how to debug that?

It's being set on this line in `_handleHideShow` on line 457 of this file:
https://github.com/patternfly/patternfly-elements/pull/1804/commits/0f77651dbfaf06298f13461be8d876f2c8f82748#diff-a9cda0cfdb7c551c3c59f90512491b5c978994191e89f33a9ff0ef8528bad640R457

wesruv:
Hope it's cool, I pushed a commit that fixed some borked code:
https://github.com/patternfly/patternfly-elements/pull/1804/commits/3c20a710e7bfaca240cd711066b5a92aefa84311

elements/pfe-clipboard/pfe-clipboard.ts
```diff
@@ -0,0 +1,323 @@
...
+   * If Clipboard is unavailable, it will use the legacy [`execCommand`](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand)
+   *
+   * ```js
+   * document.querySelector("pfe-clipboard").copyTextToClipboard(text)
+   *   .then(res => console.log(`Successfully copied: ${res}`))
+   *   .catch(error => console.error(error));
+   * ```
+   *
+   */
+  async copyTextToClipboard(text: string): Promise<string> {
```
heyMP, line 299: Is there a way to include this in CEM?  I think we would want to advertise this as a publicly available function.

elements/pfe-clipboard/test/pfe-clipboard.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+  it('should render the default slot content.', async function() {
+    const el = await createFixture<PfeClipboard>(element);
+    expect(el.shadowRoot!.querySelector('#text')!.textContent).to.equal('Copy URL');
+    expect(el.shadowRoot!.querySelector('#text--success')!.textContent).to.equal('Copied');
+    expect(el.shadowRoot!.querySelector<HTMLSlotElement>(`#icon`)!.assignedElements().length).to.equal(0);
+  });
+
+  it('it should render slot overrides', async function() {
+    const el = await createFixture<PfeClipboard>(element);
+    // The default slot override will be handled by transposeSlot
```
heyMP, line 34: This comment should be removed.  `transposeSlot` was removed in a previous version.

elements/pfe-clipboard/test/pfe-clipboard.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+    const defaultSlot = `<span slot="text">You can totally click to copy url</span>`;
+    const textSuccessSlot = `<span slot="text--success">Making some copies!</span>`;
+    const iconSlot = `<pfe-icon slot="icon" icon="web-icon-globe" color="darker"></pfe-icon>`;
+    el.innerHTML = `
+        ${defaultSlot}
+        ${textSuccessSlot}
+        ${iconSlot}
+    `;
+    // transposeSlot should have sent it to the text named slot
```
heyMP, line 43: this comment can also be removed.

elements/pfe-clipboard/test/pfe-clipboard.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+    expect(el.shadowRoot!.querySelector<HTMLSlotElement>(`#text`)!.assignedElements({ flatten: true })
+      .map(i => i.outerHTML.trim()).join('')).to.equal(defaultSlot);
+    // The success and icon slots should be working as expected also
+    expect(el.shadowRoot!.querySelector<HTMLSlotElement>(`#text--success`)!.assignedElements({ flatten: true })
+      .map(i => i.outerHTML.trim()).join('')).to.equal(textSuccessSlot);
+    // deprecated slot
+    expect(el.shadowRoot!.querySelector<HTMLSlotElement>(`#text--success`)!.assignedElements({ flatten: true })
```
heyMP, line 50: This test will fail if we stack the the `success-text` and `text--success` slots.  It can probably be removed or a new test will need to be written to test that the `text--success` slot still works.

elements/pfe-clipboard/test/pfe-clipboard.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+      'pfe-clipboard should be disabled when target=property but "contentToCopy" has not been set'
+    ).to.be.true;
+    el.contentToCopy = copyText;
+    await el.updateComplete;
+    expect(
+      el.hasAttribute('disabled'),
+      'pfe-clipboard should not be disabled when target=property and "contentToCopy" has been set'
+    ).to.be.false;
+    el.click();
+    expect(navigator.clipboard.writeText).to.have.been.calledWith(copyText);
```
heyMP, line 141: This is cool!  So we get `calledWith` from Sinon's `stubs` function?

elements/pfe-primary-detail/pfe-primary-detail.ts
```diff
@@ -477,6 +476,11 @@ export class PfePrimaryDetail extends LitElement {
       }));
     }
 
+    if (this.breakpoint === 'compact' && this.active)
+      this._detailsBackButton?.setAttribute('aria-expanded', 'true');
```
bennypowers, line 480: imperative DOM manipulations on elements that lit-html manages tend to break things. Are you able to refactor this to compute the value in the `render` method?

elements/pfe-primary-detail/pfe-primary-detail.ts
```diff
@@ -513,6 +517,7 @@ export class PfePrimaryDetail extends LitElement {
       const details = this.root?.getElementById(this.controlsId ?? '');
       // Set focus back to toggle that was activated
       this.active = null;
+      this._detailsBackButton?.removeAttribute('aria-expanded');
```
bennypowers, line 520: @wesruv same here

elements/pfe-primary-detail/pfe-primary-detail.ts
bennypowers: Should be something like
```ts
render() {
  const detailsBackButtonExpanded = 
      this.breakpoint === 'compact' && this.active ? 'true'
    : undefined
  return html`
    ...
    <button
        id="details-wrapper__back
        aria-expanded="${ifDefined(detailsBackButtonExpanded)}">
  `;
}

**Comment by bennypowers** [comment]
> Hope it's cool, I pushed a commit that fixed some borked code: [3c20a71](https://github.com/patternfly/patternfly-elements/commit/3c20a710e7bfaca240cd711066b5a92aefa84311)

I left some comments there. TL;DR: we should always let lit manage any DOM nodes in it's templates, and never do imperative manipulations on those nodes after-the-fact. Lit provides directives which let us do any operations we'd want in a declarative fashion

File: elements/pfe-clipboard/pfe-clipboard.ts [hunk 10]
```diff
@@ -0,0 +1,323 @@
...
+            </slot>
+          </div>
+      ` : ''}
+      <div class="pfe-clipboard__text">
+          <slot name="text" id="text">${this.textSlotDefault}</slot>
+      </div>
+      <div class="pfe-clipboard__text--success" role="alert">
+          <slot name="success-text" id="success-text">Copied</slot>
+          <slot name="text--success" id="text--success">Copied</slot>
+      </div>
```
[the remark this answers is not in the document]
bennypowers: I went for parallel instead of stacked to allow for piecemeal upgrades - meaning so that users can upgrade the dependency on a given page without having to change all their markup at once. but if you think that's not an issue here (and thinking it over, it probably isn't, since it's unlikely that multiple elements would be slotted into this position) then we can stack this one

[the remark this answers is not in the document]
bennypowers: we might also be able to solve by hiding according to `hasSlot`

.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "@patternfly/eslint-config-elements"
```
wesruv, line 2: Rules are marking 'extra' `{` as an error, would prefer if the project allowed folks that prefer their `{` or prefer to leave out the unnecessary ones.

Chat thread: https://chat.google.com/room/AAAAQFSmB88/CTIQyQNE3W4

elements/pfe-clipboard/pfe-clipboard.ts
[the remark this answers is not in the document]
heyMP: Stacked should still allow for piecemeal upgrades, no?  If there is no element using the `success-text"`slot (new), then it will fallback will expose the `text--succes` slot (old) for elements to opt into.  I tested it locally and it works.

elements/pfe-clipboard/demo/pfe-clipboard.html
```diff
@@ -0,0 +1,113 @@
...
+
+  <h2>Clipboard: with custom icon</h2>
+  <pfe-clipboard role="button" tabindex="0">
+    <pfe-icon slot="icon" icon="web-icon-globe"></pfe-icon>
+  </pfe-clipboard>
+
+  <h2>Clipboard: with custom text & copying text from element</h2>
+  <pfe-clipboard role="button" tabindex="0" copy-from="#textToCopy">
+    <span slot="text">This will copy the text in the text field below!</span>
+    <span slot="text--success">Making some copies!</span>
```
heyMP, line 34: I think we can find and replace these instances of `text--success` with the new `success-text` slots in our demos.

elements/pfe-clipboard/test/pfe-clipboard.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+    expect(el)
+      .to.be.an.instanceOf(customElements.get('pfe-clipboard'))
+      .and
+      .to.be.an.instanceOf(PfeClipboard);
+  });
+
+  it('should render the default slot content.', async function() {
+    const el = await createFixture<PfeClipboard>(element);
+    expect(el.shadowRoot!.querySelector('#text')!.textContent).to.equal('Copy URL');
+    expect(el.shadowRoot!.querySelector('#text--success')!.textContent).to.equal('Copied');
```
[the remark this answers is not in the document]
bennypowers: got a nice fix for this one that also optimizes SlotController

elements/pfe-clipboard/pfe-clipboard.ts
bennypowers: it is included? https://deploy-preview-1804--patternfly-elements.netlify.app/components/clipboard/#copytexttoclipboardtext-string

elements/pfe-clipboard/test/pfe-clipboard.spec.ts
bennypowers: ack

elements/pfe-clipboard/test/pfe-clipboard.spec.ts
bennypowers: ack

elements/pfe-clipboard/test/pfe-clipboard.spec.ts
bennypowers: yeah it's got quite some beef to it

https://www.chaijs.com/plugins/sinon-chai/

sinon-chai lags behind sinon a bit so you can always just get a reference to the stub if you want some of the more advanced apis

elements/pfe-clipboard/demo/pfe-clipboard.html
bennypowers: ack

should we rename `success-text` to just `success`?

elements/pfe-clipboard/test/pfe-clipboard.spec.ts
bennypowers: please check again

elements/pfe-clipboard/test/pfe-clipboard.spec.ts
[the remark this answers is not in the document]
bennypowers: please check again

elements/pfe-clipboard/pfe-clipboard.ts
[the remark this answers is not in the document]
bennypowers: please check again

elements/pfe-clipboard/pfe-clipboard.ts
heyMP: It's included in the docs section but not in the `api-viewer`.  I didn't know if CEM had the connotation of documenting public APIs.

elements/pfe-clipboard/test/pfe-clipboard.spec.ts
heyMP: 🎉

elements/pfe-clipboard/demo/pfe-clipboard.html
heyMP: `success` is good.  Trying to un-BEM my brain. :P

elements/pfe-clipboard/pfe-clipboard.ts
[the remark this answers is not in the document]
heyMP: yep that works.  The only caveat is that if we go with SSR, It won't slot the content into the shadowRoot until the component hydrates.  But in this case, it won't matter because it needs javascript to show the success message. lol

elements/pfe-clipboard/test/pfe-clipboard.spec.ts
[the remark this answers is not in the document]
heyMP: Nice!

elements/pfe-clipboard/test/pfe-clipboard.spec.ts
heyMP: 👍

elements/pfe-clipboard/pfe-clipboard.ts
heyMP: Never mind, didn't realize the docs were automatically generated 🤦   Nice!

elements/pfe-clipboard/pfe-clipboard.ts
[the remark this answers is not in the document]
bennypowers: yeah ?hidden might be better, but we can loop back around to that later if we want, maybe with an ssr-compatible directive?

lmk what you think, and plz resolve if you're satisfied

elements/pfe-clipboard/pfe-clipboard.ts
bennypowers: does api-viewer have methods support?
if we don't already, we should add a parallel analyze watcher to the `start` script

@heyMP what's your opinion on whether-or-not to commit the `custom-elements.json` files?

elements/pfe-clipboard/demo/pfe-clipboard.html
bennypowers: should we rename `text` to `label`? 

I might think `slot="text"` is the text-to-be-copied

elements/pfe-clipboard/pfe-clipboard.ts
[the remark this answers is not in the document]
heyMP: Solid plan 👍

elements/pfe-clipboard/pfe-clipboard.ts
heyMP: I was thinking that they would be generated at build time and published in the npm packages therefore we would keep them in `.gitignore`.

+1 on the parallel analyze watcher.

heyMP: Added it as an issue.

elements/pfe-clipboard/demo/pfe-clipboard.html
heyMP: Makes sense 👍 

note: we'll have to update the slot names in the `docs` also, in addition to the `demo`.

elements/pfe-datetime/pfe-datetime.ts
```diff
@@ -0,0 +1,215 @@
...
+
+
+    if (Date.parse(newVal) && this._datetime === Date.parse(newVal))
+      return;
+
+
+    this.setDate(Date.parse(newVal));
+  }
+
+  protected _timestampChanged(_oldVal?: number, newVal?: number) {
```
heyMP, line 113: Won't these `propertyChanged` callbacks always contain `oldValue, newVal`?  Just wondering if we need the optional typescript `?`.

elements/pfe-datetime/pfe-datetime.ts
```diff
@@ -0,0 +1,215 @@
...
+    if (newVal == null || this._timestamp === newVal) {
+      this.logger.log('early return');
+      return;
+    }
+
+    this._timestamp = newVal;
+    this.setDate(new Date(newVal * 1000));
+  }
+
+  private _getOptions(): Intl.DateTimeFormatOptions {
```
heyMP, line 123: `Intl.DateTimeFormatOptions` that's a really useful feature of TS. 😀

elements/pfe-datetime/test/pfe-datetime.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+    expect(regex.test(textContent!), 'should show a formatted date with time')
+      .to.be.true;
+  });
+
+  it('should show a relative time since the date', async function() {
+    const element = await createFixture<PfeDatetime>(html`
+      <pfe-datetime id="yearsago" type="relative" datetime="Mon Jan 2 15:04:05 EST 2006">
+        Mon Jan 2 15:04:05 EST 2006
+      </pfe-datetime>`);
+    const text = element.shadowRoot!.querySelector('span')!.textContent;
```
heyMP, line 68: I haven't seen the `!` used like this before.  How does it differ from optional chaining with a `?`

heyMP (commented):
Datetime feedback 🎉

elements/pfe-datetime/pfe-datetime.ts
```diff
@@ -0,0 +1,215 @@
...
+  },
+  second: {
+    'numeric': 'numeric',
+    '2-digit': '2-digit',
+  },
+  timeZoneName: {
+    short: 'short',
+    long: 'long',
+  },
+}) as [keyof Required<Intl.DateTimeFormatOptions>, object][];
```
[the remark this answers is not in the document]
bennypowers: `Record` is for object types, this here's a list of [tuple](https://www.typescriptlang.org/play#example/tuples), not to be confused with the [record and tuple ecmascript proposal](https://github.com/tc39/proposal-record-tuple) (stage 2)

elements/pfe-datetime/pfe-datetime.ts
bennypowers: not necessarily, if the property isn't initialized. practically speaking I've observed (pun!) that oldVal is not always defined

elements/pfe-datetime/pfe-datetime.ts
bennypowers: dom.lib.ts top reads of 2021

elements/pfe-datetime/test/pfe-datetime.spec.ts
bennypowers: this is the [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#non-null-assertion-operator-postfix-), a typescript exclusive. It's a postfix operator that modifies a union of anything with a nully type, removing the nully types, i.e. asserting that the expression is not nully.

in this case, `querySelector` returns `HTMLSpanElement|null`, so the `!` here asserts _a priori_ that the result of the expression is an `HTMLSpanElement`.

Unlike the optional chaining operator `?.` which performs a check at run time, the non-null assertion operator `!` has no runtime significance. In other words it's basically the same as `as HTMLSpanElement`, just a little terser. For that reason, our linting rules allow the non-null assertion operator **only** in test files.

bennypowers:
@wesruv @heyMP NB I rebased-and-force-pushed just now. Please fetch and reset

```
git fetch
git reset origin/feat/pfe-tools
```
alternately rebase any pending changes you have.

Also, please review comments

Apologies for the inconvenience.

elements/pfe-primary-detail/pfe-primary-detail.ts
bennypowers: @wesruv I made this change just now (see force-push, below). I also added a test in the spec file. PTAL

elements/pfe-clipboard/pfe-clipboard.ts
bennypowers: ok I implemented this just now

elements/pfe-primary-detail/pfe-primary-detail.ts
bennypowers: @wesruv can you check this again against the latest (rebased) changes, please?

elements/pfe-datetime/pfe-datetime.ts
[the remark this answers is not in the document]
heyMP: 🏀'in

elements/pfe-datetime/pfe-datetime.ts
heyMP: Solid Dad joke. 💯

elements/pfe-datetime/test/pfe-datetime.spec.ts
heyMP: Unbelievable explanation.  Thank you 🙏

elements/pfe-clipboard/demo/pfe-clipboard.html
bennypowers: done

elements/pfe-primary-detail/pfe-primary-detail.ts
bennypowers: Can you also check the errors thrown in the tests?

```
npm run test -- --group default --watch --files elements/pfe-primary-detail/test/primary-detail.spec.ts
```
then hit `d` to launch debug the tests in a browser

.eslintrc.json
bennypowers: I set this rule like you asked :)

Please tell me when is a convenient time for you and i'll rebase/force push again

heyMP (commented):
pfe-toast review:  Everything looks great!  Submitted PR for docs updates.

elements/pfe-readtime/pfe-readtime.ts
```diff
@@ -0,0 +1,171 @@
...
+  @property({ reflect: true })
+  lang: string = document.documentElement.lang || 'en';
+
+  /**
+   * Specify the selector of the content you want to capture the word-count from.  Whatever you enter into this attribute will be found using `document.querySelector(<for attribute value>)`.
+   */
+  @observed
+  @property({ type: String, reflect: true }) for?: string
+
+  @property({ type: Number, reflect: true }) readtime = 0;
```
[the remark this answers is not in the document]
bennypowers: `readonly` flag in CEM is still [just a proposal](https://github.com/webcomponents/custom-elements-manifest/issues/34) but I don't mind using it in a non-standard way.

If the property _is_ readonly, would it make more sense to write it as a getter, and set the attribute imperatively when the dependencies change, rather than decorating a class field with `@property`?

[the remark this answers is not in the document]
heyMP: Nice, yeah I agree.  I'll submit a PR for that. :)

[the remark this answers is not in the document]
bennypowers: you can add [cem-plugin-readonly](https://github.com/bennypowers/cem-plugins/tree/main/plugins/cem-plugin-readonly) as a dependency to pfe-tools package (not repo root!) to get the CEM support

bennypowers:
I am once again asking you to `git fetch && git reset --hard origin/feat/pfe-tools`

I squashed commits into per-package commits, trying to respect `Co-authored-by`

please stash work and 
```bash
git fetch
git reset --hard origin/feat/pfe-tools
```

elements/pfe-button/README.md
```diff
@@ -14,14 +14,6 @@ the only light DOM child of `pfe-button`.
...
-
-```html
-<pfe-button size="large">
-  <button>Large Button</button>
-</pfe-button>
-```
-
 ### Accessibility
 `<pfe-button>` addresses the issues associated with typical implementations of
 web component buttons. When using a `<pfe-button>` in a `<form>` element, the
```
zeroedin, line 19: The light dom element will only pass `id | class | type` attributes to the shadow DOM should the README.md and subsequent docs reflect this restriction?  Would we want to open this up to all valid attributes for button in the future i.e `data` or `value`?

elements/pfe-button/demo/pfe-button.html
```diff
@@ -0,0 +1,21 @@
+<link rel="stylesheet" href="/elements/pfe-button/demo/demo.css">
```
zeroedin, line 1: Expand on demo.html to show to reflect more of the available feature set.  

```Suggestion
<link rel="stylesheet" href="/elements/pfe-button/demo/demo.css">

<section>

  <pfe-band size="small" class="contextual">
    <h2 slot="pfe-band--header">Default - Medium size</h2>
    <pfe-button>
      <button>Primary</button>
    </pfe-button>

    <pfe-button variant="secondary">
      <button>Secondary</button>
    </pfe-button>

    <pfe-button variant="tertiary">
      <button>Tertiary</button>
    </pfe-button>

    <pfe-button variant="danger">
      <button>Danger</button>
    </pfe-button>

    <pfe-button variant="control">
      <button>Control</button>
    </pfe-button>

    <pfe-button disabled>
      <button>Disabled</button>
    </pfe-button>
  </pfe-band>

  <pfe-band size="small" class="contextual">
    <h2 slot="pfe-band--header">Large size</h2>
    <pfe-button size="large">
      <button>Primary</button>
    </pfe-button>

    <pfe-button variant="secondary" size="large">
      <button>Secondary</button>
    </pfe-button>

    <pfe-button variant="tertiary" size="large">
      <button>Tertiary</button>
    </pfe-button>

    <pfe-button variant="danger" size="large">
      <button>Danger</button>
    </pfe-button>

    <pfe-button  variant="control" size="large">
      <button type="submit">Control</button>
    </pfe-button>

      <pfe-button size="large" disabled>
      <button>Disabled</button>
    </pfe-button>
  </pfe-band>

  <pfe-band size="small" class="contextual">
    <h2 slot="pfe-band--header">Button types</h2>
    <pfe-button>
      <button type="submit">Primary [Submit]</button>
    </pfe-button>

    <pfe-button variant="secondary">
      <button type="reset">Secondary [Reset]</button>
    </pfe-button>

    <pfe-button variant="tertiary">
      <button type="button">Tertiary [Button]</button>
    </pfe-button>
  </pfe-band>
</section>

```

elements/pfe-button/pfe-button.ts
```diff
@@ -0,0 +1,180 @@
...
+ * @csspart container
+ * @csspart {HTMLButtonElement} button - The shadow button
+ */
+@customElement('pfe-button') @pfelement()
+export class PfeButton extends LitElement {
+  static readonly styles = styles;
+
+  /** Disables the button */
+  @observed
+  @property({ type: Boolean }) disabled = false;
```
zeroedin: I'm pretty sure this was already fixed.  But the disabled property should be set to `reflect: true`

```Suggestion
```

elements/pfe-button/README.md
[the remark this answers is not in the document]
bennypowers: I'm still not super clear on why we're copying the button into Shadow DOM instead of just using the slotted element. If like to get some regression tests out together so we can evaluate just using slot.

bennypowers:
Rebased just now, please fetch and reset

elements/pfe-button/README.md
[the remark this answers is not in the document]
bennypowers: I added some code to synchronize the `type` attr. I'd prefer to either use the slotted button or the ElementInternals polyfill, but that will have to wait until after the break. I also added your demo cases. Thanks!

bennypowers:
Heads up - I rebased again

heyMP (dismissed):
Looks good to me!  Did browser testing too and everything looks good.

core/pfe-sass/mixins/_custom-properties.scss
```diff
@@ -190,6 +190,12 @@
 ///   // ...as well as all the values for :host([on="saturated"])  and  :host([on="light"])
 @mixin pfe-contexts($contexts: $CONTEXTS) {
   // Helper attributes to aid in theming
+
+  :host([context="dark"]) { --context: dark; }
+  :host([context="light"]) { --context: light; }
+  :host([context="accent"]) { --context: saturated; }
+  :host([context="saturated"]) { --context: saturated; }
+
```
heyMP, line 198: How was this working before without this? 🤔

heyMP (commented):
Changes eleventy config look good 👍

core/pfe-sass/mixins/_custom-properties.scss
bennypowers: 🙈

bennypowers: real answer: [javascript](https://github.com/patternfly/patternfly-elements/blob/0eb2f08d5e78f61104b0503a360e96aaec5a3613/elements/pfelement/src/pfelement.js#L286)

elements/pfe-card/demo/demo.css
```diff
@@ -1,4 +1,7 @@
-/* For demo only */
+:host {
+  padding: 1em;
+}
+
```
heyMP, line 4: What's this selector targeting?

heyMP (commented):
<pfe-card> Submitting some small docs changes too. 👍

elements/pfe-card/demo/demo.css
bennypowers: <html-include>

Nouveau (commented):
The health index element looks good to me. All of the attributes are working as expected, the elements have the correct markup, and all of the tests are present.

elements/pfe-card/demo/demo.css
heyMP: 👍

zeroedin (commented):
While reviewing I ran into a bug of  the css custom properties feature .  I decided to take a look and make an attempt at providing a fix.  I created a branch on my fork https://github.com/zeroedin/patternfly-elements/tree/fix/pfe-avatar-css-custom-prop with some suggested changes for review before providing a PR.  I'd like to have some extra eyes on it to make sure I'm not introducing any breaking changes with some assumptions I made.

I did change the scope of when `customColors` are registered from class level to instance level which could be a misunderstanding of its original intention.  My assumption is you could want to add custom properties to the instance of pfe-avatar by wrapping that component in a class ex:

```
:host {
 --pfe-avatar--colors: #EE0000 #f0ab00 #007a87 #00b9e4 #92d400; 
}

.custom-color-set {
  --pfe-avatar-colors: #ffcc00;
}

.custom-color-set {
  --pfe-avatar-colors: #000000;
}
```

```/
<!-- this avatar will be a selection based on name attr from array of colors on global scope --> 
<pfe-avatar>..</pfe-avatar>

<div class="custom-color-set">
  <!-- this avatar will be #ffcc00; -->
  <pfe-avatar>...</pfe-avatar>
</div>

<div class="custom-color-set-2">
  <!-- this avatar will be #000000; -->
  <pfe-avatar>...</pfe-avatar>
</div>
```

My understanding is that in its previous implementation the custom prop would have always been on the global scope in the cascade?  I could be wrong there but because this feature was not working I haven't gone back to try and do a more detailed test of that theory.

The only other changes made were to align more with my understanding of the Lit Lifecycle events and removal of some decorators I didn't yet fully understand, but if we see the need we could add those back in.  This was around  @observed properties, and the @register decorator on the class level, and subsequent changes to incorporate the Lit lifecycle like `this.requestUpdate()` 

🍻

elements/pfe-card/pfe-card.ts
```diff
@@ -17,6 +17,25 @@ import style from './pfe-card.scss';
  * @slot - Any content that is not designated for the header or footer slot, will go to this slot.
  * @slot footer
  *       Use this slot for anything that you want to be stuck to the base of the card. This region is bottom-aligned.
+ *
+ * @cssproperty --pfe-band--BackgroundColor
```
bennypowers, line 23: awesome

elements/pfe-button/README.md
bennypowers: this is taken care of in the latest iteration - we no longer copy the button into shadow dom

elements/pfe-badge/demo/pfe-badge.js
```diff
@@ -0,0 +1,8 @@
+const { shadowRoot: root } = document.querySelector('[data-demo="pfe-badge"]');
+
+for (const el of root.querySelectorAll('pfe-badge')) {
+  if (el.previousElementSibling.localName === 'pfe-codeblock') {
+    el.previousElementSibling.querySelector('code').textContent =
+      el.outerHTML.replace(' pfelement="" class="PFElement" on="null"', '');
```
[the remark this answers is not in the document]
bennypowers: I believe this was already taken care of, please fetch the latest and try again

docs/demo/index.njk
```diff
@@ -78,6 +78,10 @@ permalink: "demo/{{ element }}/index.html"
       {%- endif -%}
     </template>
     <script type="module" src="{{ demoConfig.module }}"></script>
+    {% if element not 'pfe-band' %}
```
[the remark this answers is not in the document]
bennypowers: Just made a bunch of changes, so stash, then try `git fetch && git reset --hard origin/feat/pfe-tools && run clean:deep && npm ci && run watch:docs`

heyMP (commented):
ballin 👍

heyMP (commented):
pfe-collapse: submitted a PR for doc suggestions also.

https://github.com/patternfly/patternfly-elements/pull/1845

elements/pfe-collapse/pfe-collapse-toggle.ts
```diff
@@ -0,0 +1,152 @@
...
+  @observed
+  @property({ type: Boolean, attribute: 'expanded' }) expanded = false;
+
+  @observed
+  @property({ attribute: 'aria-controls', reflect: true }) ariaControls = '';
+
+  @property({ type: Boolean, reflect: true }) disabled = false;
+
+  @state() private _setTabIndex = false;
+  @state() private _addKeydownHandler = false;
```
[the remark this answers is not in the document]
bennypowers: looking over the code again, these are set once and then never touched again, so I just removed them altogether and assumed they are true in the places which check them.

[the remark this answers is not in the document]
heyMP: 👍

elements/pfe-page-status/demo/pfe-page-status.html
```diff
@@ -0,0 +1,32 @@
+<link rel="stylesheet" href="/core/pfe-core/pfe.min.css">
+<link rel="stylesheet" href="/core/pfe-styles/pfe-base.css">
+<link rel="stylesheet" href="/core/pfe-styles/pfe-context.css">
+<link rel="stylesheet" href="/core/pfe-styles/pfe-layouts.css">
+
+<h3>Set Status</h3>
```
Nouveau: I think we should wrap this h3 and select in a <pfe-band>. That will fix the overlapping elements issue on the status demo and get these elements the correct side padding.

Nouveau:
In looking over the page-status element, I noticed some overlap issues with the api-viewer. We could clear that up plus make the demo more robust if we adjust the z-index and add a background color to the viewer.

core/pfe-sass/mixins/_mixins.scss
```diff
@@ -1,10 +1,5 @@
 @mixin browser-query($browser-list) {
```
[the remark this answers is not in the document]
bennypowers: did that break the build for other packages?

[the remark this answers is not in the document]
bennypowers: the reason i ask it because I've found that changing the sass module graph can have lots of unintended consequences

elements/pfe-markdown/test/pfe-markdown.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+  </pfe-markdown>`,
+  'removed': html`<pfe-markdown>
+    <div pfe-markdown-container># pfe-markdown</div>
+  </pfe-markdown>`,
+};
+
+describe('<pfe-markdown>', function() {
+  it('should upgrade pfe-markdown', async function() {
+    const pfeAccordion = await createFixture<PfeMarkdown>(html`<pfe-markdown></pfe-markdown>`);
+    expect(pfeAccordion, 'pfe-markdown should be an instance of PfeMarkdown')
```
zeroedin, line 27: Rename variable `pfeAccordion` to `element`

```Suggestion
    const element = await createFixture<PfeMarkdown>(html`<pfe-markdown></pfe-markdown>`);
    expect(element, 'pfe-markdown should be an instance of PfeMarkdown')
```

elements/pfe-markdown/demo/pfe-markdown.html
```diff
@@ -0,0 +1,61 @@
+<link rel="stylesheet" href="/core/pfe-styles/red-hat-font.min.css">
+<link rel="stylesheet" href="/core/pfe-styles/pfe-base.css">
+<link rel="stylesheet" href="/core/pfe-styles/pfe-context.css">
+<link rel="stylesheet" href="/core/pfe-styles/pfe-layouts.css">
+<link rel="stylesheet" href="/core/pfe-core/pfe.min.css">
```
zeroedin, line 5: Not sure these files need to be here, I think they can be removed.

```Suggestion
 
```

elements/pfe-markdown/test/pfe-markdown.spec.ts
bennypowers: Yeah I'm all for this

elements/pfe-progress-indicator/demo/pfe-progress-indicator.html
```diff
@@ -0,0 +1,43 @@
+<link rel="stylesheet" href="/core/pfe-core/pfe.min.css">
+<link rel="stylesheet" href="/core/pfe-styles/pfe-base.css">
+<link rel="stylesheet" href="/core/pfe-styles/pfe-context.css">
+<link rel="stylesheet" href="/core/pfe-styles/pfe-layouts.css">
```
zeroedin, line 4: Not sure these stylesheets need to be included in the demo could not find the context in which they are used or needed.

```Suggestion
```

core/pfe-sass/mixins/_mixins.scss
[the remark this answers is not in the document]
zeroedin: I have not seen any adverse effects.  Demo's and Doc's builds seem to work just fine after this change.  Let me know if there is something specific you want me to look at.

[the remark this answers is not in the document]
bennypowers: no that's fine, typically I'd see errors right away if it was a problem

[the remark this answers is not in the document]
bennypowers: I noticed we only use that mixin in pfe-select, maybe we should just copy it over and remove it from sass? WDYT @heyMP @wesruv

elements/pfe-markdown/demo/pfe-markdown.html
bennypowers: did you check `http://localhost:8080/components/markdown/demo/`?

Nouveau:
I've pushed up some small fixes for page status. I think that component is good to go now!

bennypowers:
Jan 30 Changes **PLEASE FETCH AND RESET**

- Removed `<pfe-content-set>` per @wesruv 
- Updated changesets (== changelogs) for all packages
- Added CSS shadow parts to moste elements
- Added `@time()` decorator for class methods (per @heyMP in pfe-2.0-demos)

Nouveau:
While reviewing accessibility for the pfe-modal, I noticed that we need to add alt text for the Patternfly Elements logo. And it looks like we need some focus styles in the header. When tabbing through to pull up the modal, it looks like we need focus styles on the doc pages too. It's easy to get lost.

I think it would be good to add accessibility tests, and extra tests overall to catch things like the trigger function change for modal, but I don't think needing more tests should be release blocking since it's not worse than the current situation.

bennypowers:
Thanks @Nouveau  that's great feedback. Does that relate specifically to the `npm start` development SPA, to the docs site docs pages (e.g. https://deploy-preview-1804--patternfly-elements.netlify.app/components/modal/), or both?

> I think it would be good to add accessibility tests, and extra tests overall to catch things like the trigger function change for modal, but I don't think needing more tests should be release blocking since it's not worse than the current situation.

Yes, definitely more unit tests are needed especially for AT, let's open a new issue to track those

Nouveau:
> Thanks @Nouveau that's great feedback. Does that relate specifically to the `npm start` development SPA, to the docs site docs pages (e.g. https://deploy-preview-1804--patternfly-elements.netlify.app/components/modal/), or both?

That's only for the npm start development SPA, not the docs site docs pages. Those are handling focus and alt tags as expected.

Nouveau:
Functionality for pfe-modal looks good. I noticed several places where the code is cleaner than before.

I think the demo page could use some work. The three column band breaks out of the container at smaller widths, and the examples and text aren't as extensive as other components. But I don't think that should hold up a prerelease. The modal itself looks equivalent, so I think pfe-modal is ready.

core/pfe-sass/maps/_typography_deprecated.scss
```diff
@@ -0,0 +1,29 @@
+@use "@patternfly/patternfly/sass-utilities/all" as *;
+@use "../variables/4-sizes" as sizes;
+
+$pfe-typography-base-deprecated: (
```
heyMP, line 4: What are your thoughts on removing this?  It is marked for deprecation but I worry about debugging instances where we were relying on this output.  For whatever reason that might be.

heyMP (commented):
Looks good @bennypowers!  Just a few questions.

Issue #1817, opened 2021-12-14
bug: 2.0 error running `npm run build`


heyMP:
## Description of the issue

`npm run build` throwing the following error

```
Cannot find module '/Users/michael/Projects/pfe-workspace/patternfly-elements/node_modules/@patternfly/pfe-tools/custom-elements-manifest.js' imported from /Users/michael/Projects/pfe-workspace/patternfly-elements/elements/pfe-collapse/custom-elements-manifest.config.js
```

PR #1818, opened 2021-12-14
pfe-tools: added build ts files to js to build script

heyMP:
The `.js` files werent' being generated during build for pfe-tools causing the rest of the build commands to fail.

fixes: #1817


tools/pfe-tools/package.json
```diff
@@ -41,12 +41,12 @@
...
   "scripts": {
-    "build": "tsc -b .",
+    "build": "node _scripts/build.js",
     "clean": "git clean -fx -e node_modules -e tsconfig.tsbuildinfo .",
-    "dev": "node _scripts/dev.js",
+    "dev": "node _scripts/build.js -- watch",
     "prepublishOnly": "run-s prepublishOnly:*",
     "prepublishOnly:analyze": "npm run analyze",
-    "prepublishOnly:build": "tsc -b ."
+    "prepublishOnly:build": "node _scripts/build.js"
```
bennypowers, line 49: can you run `prepublish` then `npm pack` and confirm the contents of the tarballs?

bennypowers: Also, this won't generate the .d.ts files, which we'd like in the tarball

heyMP: It's complaining about the `npm run analyze` script not being available.
Screen Shot 2021-12-14 at 8 40 00 AM

heyMP:
This issue was solved by running `npm run clean:deep`.

PR #1826, opened 2021-12-20
chore: 2.0 [pfe-toast] update docs

heyMP:
Adding an `.opt-out` class for headers in the docs styles.  I added this so that the headers in `pfe-toast` can opt out.

### DIFF
--- a/docs/main.css
+++ b/docs/main.css
@@ -52,7 +52,7 @@ p.subtitle {
 pfe-band.header h1 {
   margin: 0;
 }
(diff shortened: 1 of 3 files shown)


bennypowers (approved):
Thanks for updating the JSDoc. Feel free to merge. Only nit i can pick is the class name `opt-out` - I have to admit i'm not sure what I'm opting out of. Could there be a more descriptive name? Other than that, fire at will.

PR #1831, opened 2021-12-24
feat(jump-links) 2.0 Add new ComposedEvent to jump-links

heyMP:
Updated jump-links based on your suggestions in the previous PR.  Thanks Benny 👍

### DIFF
--- a/elements/pfe-jump-links/pfe-jump-links-nav.ts
+++ b/elements/pfe-jump-links/pfe-jump-links-nav.ts
@@ -4,6 +4,7 @@ import { LitElement, html } from 'lit';
 import { unsafeHTML } from 'lit/directives/unsafe-html.js';
 import { customElement, property, query, state } from 'lit/decorators.js';
 
+import { ComposedEvent } from '@patternfly/pfe-core';
 import { pfelement, bound, observed } from '@patternfly/pfe-core/decorators.js';
 import { getRandomId } from '@patternfly/pfe-core/functions/random.js';
 import { pfeEvent } from '@patternfly/pfe-core/functions/pfeEvent.js';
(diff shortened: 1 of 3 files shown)


elements/pfe-jump-links/pfe-jump-links-nav.ts
```diff
@@ -33,16 +34,31 @@ function isQueryable(x: Node): x is Document|ShadowRoot {
...
+  ) {
+    super('stuck');
+  }
+}
+
+export class ActiveItemEvent extends ComposedEvent {
+  constructor(
+    index: keyof Array,
+  ) {
+    super('active-item');
```
bennypowers, line 49: WDYT about this name? I was thinking `change` might be better, but we already use `change` in panel...

bennypowers (changes requested):

elements/pfe-jump-links/pfe-jump-links-nav.ts
heyMP: maybe:
`active-item` => `change`
`change` => `content-change`

I think that whatever name we decide for panel contents changing we'll want to standardize for other elements reacting to mutation observer changes.

bennypowers: let's defer that discussion to later, k?

elements/pfe-jump-links/pfe-jump-links-nav.ts
```diff
@@ -33,16 +34,31 @@ function isQueryable(x: Node): x is Document|ShadowRoot {
...
+  constructor(
+    public stuck: Boolean,
+  ) {
+    super('stuck');
+  }
+}
+
+export class ActiveItemEvent extends ComposedEvent {
+  constructor(
+    index: number,
```
bennypowers, line 47:
```suggestion
    public index: number,
```

elements/pfe-jump-links/pfe-jump-links-nav.ts
heyMP: Sounds good.  I made the changes to the event names and updated the tests to cover the new events.

elements/pfe-jump-links/pfe-jump-links-nav.ts
heyMP: Nice. I updated it in the latest commit.

PR #1833, opened 2021-12-28
pfe-primary-detail: Simplifying back button text and removing unused functions

wesruv:
```
Removed aria attributes from element surrounding text, only need htose on the button wrapper
Set back button text to property so we aren't reading the DOM during render (? seems like the that's recommended)
Removed unused functions made during original development, but are now unused
Updated Demo HTML, thought text in 'details' was a bug in functionality, but was content in the slot. Hopefully the slight adjustment makes that more clear.
```

This update isn't as necessary as I thought; the 'bug' I was chasing was content in `slot=details` on the first example, not duplicate content added by the component code 🤦🏼‍♂️ 
Was confused myself about how this thing is supposed to work... so that's not a great sign.

_I think_ that having the back button text be a property is probably simpler? Seems like we're accessing the DOM during render if we don't which was listed as one the 'uncool' things in the Lit render docs.

Whether or not we use this update to the code, I think pfe-primary-details is good to go. So after you decide what you want to do with this we can check it off the list.

### DIFF
--- a/elements/pfe-primary-detail/pfe-primary-detail.ts
+++ b/elements/pfe-primary-detail/pfe-primary-detail.ts
@@ -93,6 +93,9 @@ export class PfePrimaryDetail extends LitElement {
   /** Read only: Displays what breakpoint is currently being used */
   @property({ type: String, reflect: true }) breakpoint?: 'compact'|'desktop';
 
+  /** Used to set text of back button **/
+  @property({ type: String }) expandedSectionTitle?: string;
+
   @state() private pristine = true;
   @state() private controlsId?: string;
   @state() private nextToggle: HTMLElement|null = null;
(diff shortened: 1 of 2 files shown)


elements/pfe-primary-detail/pfe-primary-detail.ts
```diff
@@ -423,24 +426,17 @@ export class PfePrimaryDetail extends LitElement {
...
-    if (!this.nextToggle) {
-      return html`
-        <div id="details-wrapper__heading"
-            aria-selected="${ifDefined(desktop && 'true')}"
-            aria-expanded="${ifDefined(notDesktop && 'true')}"
-        ></div> `;
+    if (!this.expandedSectionTitle) {
+      return html`<strong id="details-wrapper__heading"></strong>`;
     } else {
       const { wasTag } = this.nextToggle.dataset;
```
bennypowers, line 432:
```suggestion
      const { wasTag } = this.nextToggle?.dataset ?? {};
```

bennypowers (changes requested):

bennypowers (approved):

PR #1834, opened 2021-12-28
Feat/pfe tools accordion

wesruv:
Small fixes for demo page that help show that everything is, in fact, working well.
```
Updating floats to grid, because floats aren't for page layout anymore and I don't wanna debug that
Disclosure demo was missing the attribute
```


wesruv:
The dynamic addition of a `header and panel` isn't working, couldn't figure out why:
image

bennypowers (approved):
LGTM

PR #1845, opened 2022-01-07
pfe-tools 2.0: pfe-collapse

heyMP:


elements/pfe-collapse/pfe-collapse-panel.ts
```diff
@@ -45,15 +45,19 @@ export class AnimationEndEvent extends ComposedEvent {
 export class PfeCollapsePanel extends LitElement {
   static readonly styles = [style];
 
+  /**
+   * @readonly
+   * Use the toggle method available on pfe-collapse and pfe-collapse-toggle to manually trigger expand.
```
bennypowers, line 50: Why should this be considered readonly? doesn't setting `el.expanded = true` work?

prior art: 
- https://github.com/material-components/material-web/blob/84857457f0e5daa69744c17675f32ce5403512ec/packages/dialog/mwc-dialog-base.ts#L90
- https://shoelace.style/components/dialog?id=properties
- https://github.com/microsoft/fast/blob/256fe07122c2838e6bf0edd639a539b9ceeaa1bf/packages/web-components/fast-foundation/src/accordion-item/accordion-item.ts#L41-L49

elements/pfe-collapse/pfe-collapse-toggle.ts
```diff
@@ -38,12 +38,23 @@ export class PfeCollapseToggle extends LitElement {
 
   private logger = new Logger(this);
 
+  /**
+   * @readonly
+   * Use the toggle method to manually trigger expand.
```
bennypowers, line 43: same as above. Also the doc block here should preferably describe the property itself and not just defer to another method

elements/pfe-collapse/pfe-collapse-toggle.ts
```diff
@@ -38,12 +38,23 @@ export class PfeCollapseToggle extends LitElement {
...
+   */
   @observed
   @property({ type: Boolean, attribute: 'expanded' }) expanded = false;
 
+  /**
+   * Explicitly specify the id of the corresponding panel to sync aria states.
+   * The toggle needs an aria-controls attribute that links to the id of the panel.
+   * If no aria-controls are specified, a unique ID will be generated and applied
```
bennypowers, line 51:
```suggestion
   * If no aria-controls id is specified, a unique ID will be generated and applied
```

bennypowers (changes requested):
Looks great!

I'm not sure we need the `@readonly` comments, but other than that, good stuff!

elements/pfe-collapse/pfe-collapse-panel.ts
bennypowers: I pushed a commit to feat/pfe-tools which moves the sides effects of `toggle()` into `_expandedChanged` and also adds  public `expand()` and `collapse()`. That allows `expanded = true` to operate like it's cousins linked above without breaking stuff.

PTAL

heyMP: Updated to reflect the new expand functionality.

bennypowers (approved):
🏆

PR #1846, opened 2022-01-10
pfe-tools 2.0: add local fonts to kitchen sink docs page.

heyMP:
This replaces the google fonts usages of "Red Hat Text" and "Red Hat Display" to local woff files.


bennypowers:
considering @zhawkins 's comments in chat do we need to merge this @heyMP ?

heyMP:
No, we can use Google fonts.

PR #1849, opened 2022-01-12
feat: Pfe tools/pfe dropdown

heyMP:
- update docs
- added `expanded` public property
- fixed keyboard navigation issues
- added tests for keyboard navigation

### DIFF
--- a/elements/pfe-dropdown/pfe-dropdown-item.ts
+++ b/elements/pfe-dropdown/pfe-dropdown-item.ts
@@ -10,6 +10,9 @@ import style from './pfe-dropdown-item.scss';
 export class PfeDropdownItem extends LitElement {
   static readonly styles = [style];
 
+  /** Use 'delegatesFocus' to forward focus to the first pfe-dropdown-item when this container is clicked or focused. */
+  static readonly shadowRootOptions = { ...LitElement.shadowRootOptions, delegatesFocus: true };
+
   /**
    * This is an optional attribute string that you should provide to indicate the type of dropdown item. This drives the appropriate assignment of accessibility attributes for each type of item.
    * - `link` : an HTML link
--- a/elements/pfe-dropdown/pfe-dropdown-item.ts
+++ b/elements/pfe-dropdown/pfe-dropdown-item.ts
@@ -72,10 +72,12 @@ export class PfeDropdownItem extends LitElement {
     }
   }
 
-  protected _clickHandler() {
+  protected _clickHandler(event) {
     // Forward all click events to the elements in the lightdom.
     // This fixes <a> tags from not being selected with keyboard events.
-    (this.firstElementChild as HTMLElement)?.click();
+    if (event.target === this) {
+      (this.firstElementChild as HTMLElement)?.click();
+    }
   }
 }
 
--- a/elements/pfe-dropdown/pfe-dropdown-item.ts
+++ b/elements/pfe-dropdown/pfe-dropdown-item.ts
@@ -72,7 +72,7 @@ export class PfeDropdownItem extends LitElement {
     }
   }
 
-  protected _clickHandler(event) {
+  protected _clickHandler(event: MouseEvent) {
     // Forward all click events to the elements in the lightdom.
     // This fixes <a> tags from not being selected with keyboard events.
     if (event.target === this) {
--- a/elements/pfe-dropdown/pfe-dropdown-item.ts
+++ b/elements/pfe-dropdown/pfe-dropdown-item.ts
@@ -75,7 +75,7 @@ export class PfeDropdownItem extends LitElement {
   protected _clickHandler() {
     // Forward all click events to the elements in the lightdom.
     // This fixes <a> tags from not being selected with keyboard events.
-    this.firstElementChild?.click();
+    (this.firstElementChild as HTMLElement)?.click();
   }
 }
 
--- a/elements/pfe-dropdown/pfe-dropdown-item.ts
+++ b/elements/pfe-dropdown/pfe-dropdown-item.ts
@@ -21,12 +21,12 @@ export class PfeDropdownItem extends LitElement {
    */
   @observed
   @property({ reflect: true, attribute: 'item-type' })
-  itemType?: 'link'|'action'|'separator';
+    itemType?: 'link'|'action'|'separator';
 
   /** Disabled item */
   @observed
   @property({ type: Boolean, reflect: true })
-  disabled = false;
+    disabled = false;
 
   @state() containerRole?: 'none'|'menuitem'|'separator';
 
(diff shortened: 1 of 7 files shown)


elements/pfe-dropdown/pfe-dropdown-item.ts
```diff
@@ -63,6 +71,12 @@ export class PfeDropdownItem extends LitElement {
       this.setAttribute('tabindex', '0');
     }
   }
+
+  protected _clickHandler() {
+    // Forward all click events to the elements in the lightdom.
+    // This fixes <a> tags from not being selected with keyboard events.
+    (this.firstElementChild as HTMLElement)?.click();
```
bennypowers, line 78: this will click the link twice, right? should we guard here to make sure the target is `this`?

elements/pfe-dropdown/pfe-dropdown.ts
```diff
@@ -69,14 +70,16 @@ export class PfeDropdown extends LitElement {
   @observed
   @property({ type: Boolean, reflect: true }) disabled = false;
 
-  @state() private isOpen = false;
+  @observed
+  @property({ type: Boolean, reflect: true }) expanded = false;
```
bennypowers, line 74: small nit: would `open` be a more semantic name? 

contrary to my point, and in your favour: https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html

elements/pfe-dropdown/pfe-dropdown.ts
```diff
@@ -249,7 +265,11 @@ export class PfeDropdown extends LitElement {
   }
 
   /** modify DOM if custom options are passed in an array */
-  private async _modifyDOM(options: PfeDropdownOption[]) {
+  private async _modifyDOM(options: PfeDropdownOption[], clean: Boolean = true) {
```
bennypowers, line 268:
```suggestion
  private async _modifyDOM(options: PfeDropdownOption[], clean = true) {
```

`clean` probably isn't the `Boolean` constructor, and anyways for boolean default params, TS can infer the type

elements/pfe-dropdown/pfe-dropdown.ts
```diff
@@ -249,7 +265,11 @@ export class PfeDropdown extends LitElement {
...
+    // remove all dropdown items and separators
+    if (clean) {
+      this.innerHTML = '';
```
bennypowers, line 271: seems expensive. might be better to iterate? we can defer that until we have numbers to back my assertion up

elements/pfe-dropdown/pfe-dropdown.ts
```diff
@@ -328,8 +351,32 @@ export class PfeDropdown extends LitElement {
     return item.shadowRoot?.querySelector(`.pfe-dropdown-item__container`) ?? null;
   }
 
+  /**
+   * Add dropdown items dynamically
```
bennypowers, line 355:
```suggestion
   * Add dropdown items dynamically
   * @example adding a single option
```

elements/pfe-dropdown/pfe-dropdown.ts
```diff
@@ -328,8 +351,32 @@ export class PfeDropdown extends LitElement {
...
   @bound addDropdownOptions(options: PfeDropdownOption[]) {
-    this._modifyDOM(options);
+    this._modifyDOM(options, false);
+  }
+
+  /**
+   * Override the default focus method to give focus to the internal button
+   */
+  @bound focus() {
+    this.shadowRoot?.querySelector('button')?.focus();
```
bennypowers, line 379: isn't this what delegatesFocus does?

elements/pfe-dropdown/test/pfe-dropdown.spec.ts
```diff
@@ -128,43 +140,129 @@ describe('<pfe-dropdown>', function() {
...
-    await element.updateComplete;
+      /** @deprecated */
+      setTimeout(() => element.querySelector('button')?.click());
+      const depEvent = await oneEvent(element, 'pfe-dropdown:change') as unknown as CustomEvent;
+      expect(depEvent.detail.action).to.not.be.empty;
+    });
 
-    expect(element.children.length).to.equal(options.length);
+    /** https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html */
+    describe(`keyboard accessibility`, function() {
```
bennypowers, line 169: _nice_

bennypowers (changes requested):
really nice MP

a few comments and two small changes

elements/pfe-dropdown/pfe-dropdown.ts
heyMP: That's what I thought too but I couldn't get it to work via JS.

elements/pfe-dropdown/test/pfe-dropdown.spec.ts
heyMP: I based this on your accordion work.  Thank you!

elements/pfe-dropdown/pfe-dropdown.ts
heyMP: I agree.  I always cringe with `innerHTML`.  I'll put in a `@todo` to revisit.

elements/pfe-dropdown/pfe-dropdown-item.ts
heyMP: Good call, added guard.

elements/pfe-dropdown/pfe-dropdown.ts
heyMP: I could go either way. If we use `open` then it will create a small deprecation in that `open()` is currently public method.  We also use `expanded` for accordion which is also a collapse type component."

So I suppose I would vote for `expanded`.  What say you?

heyMP:
Thanks @bennypowers!  I addressed your comments.

elements/pfe-dropdown/pfe-dropdown.ts
bennypowers: I agree with you, `expanded` is more conventional

elements/pfe-dropdown/pfe-dropdown.ts
bennypowers: Ok I'm going to give this the once over later today, see if I can get it to work 

Perhaps the issue is that the button isn't the first focusable element in the (fully projected) Shadow DOM?

bennypowers:
@heyMP can you check over the changeset, see if it needs updating? (e.g. expanded attribute)

elements/pfe-dropdown/pfe-dropdown.ts
bennypowers: was the JS problem just in testing?

bennypowers: talking this over on a call:
- the problem is primary related to testing
- our assumptions about how the `HTMLElement#focus` method relates to `:focus` styles don't obtain (https://codepen.io/bennyp/pen/JjrVOWK?editors=1100)
- we'll solve by re-doing the test to use real keyboard events via playwright via web-test-runner a la pfe-accordion tests

bennypowers (approved):
just picked a few nits in the tests for my own preferences. 

Nice work MP

PR #1851, opened 2022-01-18
fix: moved public focus method to private.

heyMP:
Removing the public `focus()` method causes an error in the following test: 
`Enter should exit open dialog and move focus to the dropdown.`.  Meaning that when you're focused on the list item and you click `Enter` on an action item it needs to programatically move the focus to `pfe-dropdown`.  Knowing that, I added a private `_focus()` method to programmatically move the focus to the button.

```
<pfe-dropdown>
  #shadowRoot
    <button>
```

### DIFF
--- a/elements/pfe-dropdown/pfe-dropdown.ts
+++ b/elements/pfe-dropdown/pfe-dropdown.ts
@@ -191,7 +191,7 @@ export class PfeDropdown extends LitElement {
       case 'Escape':
         event.preventDefault();
         event.stopPropagation();
-        this.focus();
+        this._focus();
         await this.updateComplete;
         this.close();
         break;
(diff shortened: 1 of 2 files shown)


elements/pfe-dropdown/pfe-dropdown.ts
```diff
@@ -352,6 +352,11 @@ export class PfeDropdown extends LitElement {
     return item.shadowRoot?.querySelector(`.pfe-dropdown-item__container`) ?? null;
   }
 
+  // move focus to the button
```
bennypowers, line 355:
```suggestion
  /** Move focus back to the button after closing the dropdown */
```

elements/pfe-dropdown/pfe-dropdown.ts
```diff
@@ -352,6 +352,11 @@ export class PfeDropdown extends LitElement {
...
+  private _focus():void {
+    this.shadowRoot?.querySelector<HTMLButtonElement>('#pfe-dropdown-toggle')?.focus();
```
bennypowers, line 357:
```suggestion
    this.button?.focus();
```

PR #1853, opened 2022-01-20
Pfe tools/pfe autocomplete

heyMP:
- Added tests
- Tweaked some of the events documentation

### DIFF
--- a/elements/pfe-autocomplete/pfe-autocomplete.ts
+++ b/elements/pfe-autocomplete/pfe-autocomplete.ts
@@ -26,18 +26,20 @@ export class AutocompleteShowEvent extends ComposedEvent {
   }
 }
 
+/** Fired when the user submits the autocomplete input */
 export class AutocompleteSearchEvent extends ComposedEvent {
   constructor(
-    // The search query
+    /** The search query */
     public value: string
   ) {
     super('search');
   }
 }
 
+/** Fired when the user selects an option from the dropdown */
 export class AutocompleteSelectEvent extends ComposedEvent {
   constructor(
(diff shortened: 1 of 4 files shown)


elements/pfe-autocomplete/pfe-autocomplete.ts
```diff
@@ -28,7 +28,7 @@ export class AutocompleteShowEvent extends ComposedEvent {
 
 export class AutocompleteSearchEvent extends ComposedEvent {
   constructor(
-    /** The search query */
+    // The search query
```
heyMP, line 31: I changed the documentation syntax so it wouldn't autogenerate entries in the "Events" section of the docs.  It was creating duplicate entries.

heyMP:
Looks great @bennypowers.  I added some tests for the events and apis.

elements/pfe-autocomplete/pfe-autocomplete.ts
bennypowers: hmmm 🤔 sounds like that's more a bug in analyzer. It would be a shame to lose the docs on those properties. I'll take a look, maybe we can fix it up

heyMP: oh gotcha!

bennypowers: found the problem there that was generating events with names like 'searchValue' and solved it. is that what you meant? I couldn't replicate how describing the fields on the event classes caused double-documentation, in any event it seems to work for me at the moment

PR #1855, opened 2022-01-26
fix(pfe-core): move styles to pfe-styles

zeroedin:
- updated build tasks removing sass build for pfe-core
- moved CSS/SCSS files from pfe-core to pfe-styles 
- changed all links referencing these styles from pfe-core to pfe-styles
- updated gitignore


core/pfe-core/package.json
```diff
@@ -44,10 +44,6 @@
     "build": "run-s build:*",
     "build:clean": "npm run clean",
     "build:esbuild": "node ../../scripts/build.js --include pfe-core --workspace core",
-    "build:sass": "run-s build:sass:*",
```
bennypowers, line 47: so good

core/pfe-styles/pfe--noscript.css
```diff
@@ -0,0 +1,6 @@
+/* secret weapon in the specificificificificity wars */
```
bennypowers, line 1: we need to ignore/remove this file?

elements/pfe-autocomplete/demo/pfe-autocomplete.html
```diff
@@ -2,7 +2,7 @@
 <link rel="stylesheet" href="/core/pfe-styles/pfe-base.css">
 <link rel="stylesheet" href="/core/pfe-styles/pfe-context.css">
 <link rel="stylesheet" href="/core/pfe-styles/pfe-layouts.css">
-<link rel="stylesheet" href="/core/pfe-core/pfe.css">
+<link rel="stylesheet" href="/core/pfe-styles/pfe.css">
```
bennypowers, line 5: ordinarily we'd want one commit per feature, but that won't work on this branch, unfortunately

this is annoying as heck, but in order to avoid conflicts while rebasing, these all need to be committed with a conventional commit scope like

```
docs(autocomplete): use pfe-styles
```

this is the shell script (fish shell, needs some adjustment for bash, probably) that I use for that

```fish
for p in (ls elements)
    set -lx SCOPE (string replace pfe- '' $p)
    git add "elements/pfe-$SCOPE"
    git cim "fix($SCOPE): use pfe-styles";
end
```

bennypowers (changes requested):
Looks great.

There's one tiny and one unfortunately very annoying change I'd like you to take a crack at.

Great work

elements/pfe-autocomplete/demo/pfe-autocomplete.html
bennypowers: I *think* the bash version looks like this:

```sh
for p in $(ls elements)
do
    set -x SCOPE=`${$p/pfe-/''}`
    git add "elements/pfe-$SCOPE"
    git cim "fix($SCOPE): use pfe-styles";
end
```

bennypowers:
on your branch, can you run an `npm run clean:nuke` then `npm ci` and then after those `npm start` or `npm run site`?

PR #1857, opened 2022-01-27
fix: types for trackPerformance function

heyMP:
I ran into a typescript issue when trying to use `trackPerformance` from `@patternfly/pfe-core`.

Screen Shot 2022-01-26 at 10 31 54 PM

Here is the example project showing how I was using the `trackPrerformance`.
https://github.com/heyMP/pfe-2.0-demos/blob/master/src/track-performance.ts


bennypowers:
Thanks I patched this on branch with you in the `Co-Authored-By`

PR #1896, opened 2022-03-01
chore: remove autoupdate workflow

bennypowers:
Github UI provides a button to update PRs to latest
and the auto-update workflow creates non-conventional commits,
which interferes with commitlint


zeroedin (approved):
LGTM

PR #1903, opened 2022-03-03
fix(create): update docs and tsconfig

bennypowers:
## Update docs and tsconfig

- normalize scopes in tsconfig
- make sure tsconfig references gets updated
- make sure docs bundle gets updated

### Merging

Please **rebase** when merging and ensure your commit message uses [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) formatting.


bennypowers:
we need to add two entries when generating an element
1. paths in /tsconfig.json
2. import map in docs/import-map.json

zeroedin (commented):
LGTM!  Nice work @bennypowers

zeroedin (approved):

PR #1906, opened 2022-03-04
perf: minify html literals in bundles

bennypowers:
### What has changed and why

- use [esbuild-plugin-minify-html-literals](https://npm.im/esbuild-plugin-minify-html-literals) to minify bundles

👉 Reduces all-component bundle size from **>500kb** to **~370kb** 👈 

NOTE: breaks backwards by removing `pfe-typography`, `pfe-c-typography`, and `focus-ring` sass mixins, which were not compiling under dart-sass anyways. for an example, search https://unpkg.com/browse/@patternfly/pfe-accordion@2.0.0-next.0/pfe-accordion-header.js for `pfe-var`

## Testing

Check out the deploy preview, click around and look for problems, inspect the source files and bundle to make sure they're minified

e.g.
- https://deploy-preview-1906--patternfly-elements.netlify.app/pfe.min.js
- https://deploy-preview-1906--patternfly-elements.netlify.app/components/autocomplete/pfe-autocomplete.js
Also make sure that accordion responds to context changes

### Merging

Please **rebase** when merging and ensure your commit message uses [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) formatting.


bennypowers:
Something about the minification is breaking colour context. looks like a JS minification problem and not a SASS issue ... investigating

bennypowers:
So this is fun:
accordion (and, it seems, only accordion) is not picking up color context changes, but *only* when loaded from the bundle, whether minified or not.

bennypowers:
ok i think i got it. if accordion registers before band (as it does in the bundle), then accordion's context event won't get claimed by the parent band. I'm going to tinker a little, but I think what we can do is catch every pfe-color-context request at the window (i.e. unclaimed context consumers) and add them to a Set or some other structure, then as providers upgrade, they can claim any unclaimed elements in the set as their own

bennypowers:
alright context issue seems to have been resolved

docs/demo/bundle.ts
```diff
@@ -3,33 +3,4 @@ export * from '@patternfly/pfe-icon';
...
-import '@patternfly/pfe-icon';
-import '@patternfly/pfe-markdown';
-import '@patternfly/pfe-number';
-import '@patternfly/pfe-page-status';
-import '@patternfly/pfe-progress-indicator';
-import '@patternfly/pfe-progress-steps';
-import '@patternfly/pfe-readtime';
-import '@patternfly/pfe-select';
-import '@patternfly/pfe-toast';
+import './components.js';
```
heyMP, line 6: What generates this file?

bennypowers: https://github.com/patternfly/patternfly-elements/blob/bf3224be19303b2a8d5b21ed66a9f1c3a064b0cf/tools/pfe-tools/esbuild.ts#L107-L126

Previously, we just wrote a new line to this file every time the generator ran,

but now, we just generate that file from the `/elements` dir anytime we need to.

heyMP: noice!

heyMP (approved):
LGTM 👍

PR #1950, opened 2022-03-21
Test/deflake label and autocomplete tests

bennypowers:
- adds `reporter` option to test runner config
- tries to make pfe-label and pfe-autocomplete tests less flaky


heyMP (approved):
LGTM

PR #1952, opened 2022-03-21
fix(tools): try to get repo root automatically

bennypowers:
- Tries to guess the repo root when run either from `/node_modules` or from within the pfe monorepo
- Adds a `Troubleshooting` section to README to offer guidance to users who's case was not covered by the above


heyMP (approved):
LGTM

PR #1958, opened 2022-03-22
hotfix(tools): dev server workspaces

bennypowers:
- removes export conditions from packages, they were for development use only and thus were a runtime error for consumers, and confusing
- removes default import map from dev server, but keeps the option to specify one
- adds 'double checking' step to dev sever module resolution which resolves local modules from local packages *by typescript source* instead of by js source

can be checked against the lastest commit from pfe-monorepo-template, although you'll have to either merge and publish this then update deps there, or copy (not symlink) these changes into that repo's node_modules

### DIFF
--- a/.changeset/big-dingos-taste.md
+++ b/.changeset/big-dingos-taste.md
@@ -1,0 +1,5 @@
+---
+"@patternfly/pfe-tools": patch
+---
+
+fail gracefully when dev server can't find workspace packages
--- a/.changeset/hip-cats-change.md
+++ b/.changeset/hip-cats-change.md
@@ -1,0 +1,24 @@
+---
+"@patternfly/pfe-tools": major
+---
+
+Remove support for custom `esbuild` export condition
+
+Previously, the dev-server config would try to import typescript sources using
+the `esbuild` export condition, but this proved awkward when daughter repos
+would try to import the same packages by the esbuild condition, only to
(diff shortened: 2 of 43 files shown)


zeroedin (commented):
To go along with all the changes to the elements `package.json` example `elements/pfe-band/package.json`

```json
  "exports": {
    ".": "./pfe-band.js",
    "./*": "./*.js"
  }
```
 
We'll want to make the same change to `tools/create-element/templates/element/package.json`

```json
  "exports": {
    ".": "./<%= tagName %>.js",
    "./*": "./*.js"    
  },
```

bennypowers:
> To go along with all the changes to the elements `package.json` example `elements/pfe-band/package.json`
> 
> ```json
>   "exports": {
>     ".": "./pfe-band.js",
>     "./*": "./*.js"
>   }
> ```
> 
> We'll want to make the same change to `tools/create-element/templates/element/package.json`
> 
> ```json
>   "exports": {
>     ".": "./<%= tagName %>.js",
>     "./*": "./*.js"    
>   },
> ```

agreed.

also, we'll need to inject 
```js
import '<%= packageName =>';
```
into the demo file.

if you're able, please commit those, if not don't worry about it I can get to it later tonight or tmrw

zeroedin (approved):
Looks good to me

Issue #1963, opened 2022-03-24
bug: @patternfly/create-element not included in post install scripts


brianferry:
The @patternfly/create-element tool is not included in the post install scripts which means to run `npm run new` you need to first run `npm run build -w @patternfly/create-elements` if you plan to use the new script.  This should be updated so that it is handled by the postinstall scripts.

PR #1964, opened 2022-03-24
fix(tools): adding @patternfly/create-element to postinstall scripts

brianferry:
## <Pull request description>

The @patternfly/create-element tool is not included in the post install scripts which means to run npm run new you need to first run npm run build -w @patternfly/create-elements if you plan to use the new script. This should be updated so that it is handled by the postinstall scripts.

### Related issues

- (#1963)


zeroedin (approved):
LGTM

PR #1965, opened 2022-03-24
docs: patch html-includes if needed

bennypowers:
see https://github.com/justinfagnani/html-include-element/pull/21

`<html-include>` dies when a subresource errors.

This PR patches the `HTMLIncludeElement` class at runtime to prevent that, allowing the rest of the HTML partial to load

> Why not patch-package?

- it's dev-only so i'm less worried about code size
- patch-package would become an error if and when above PR is merged and we update
- i'd rather vend this snippet than vend a patch-package diff 🤷


zeroedin (approved):
Lima Golf Tango Mike

PR #1973, opened 2022-03-29
fix(tools): package demo files

bennypowers:
Makes sure the dev server assets are included in the pfe-tools npm tarball

also updates some dependencies

Instigating problem: updating pfe-tools in pfe-monorepo-template then `npm start`, the dev server complains that it can't find the index.html template. And for good reason, because we weren't including it.


zeroedin (approved):
👍 👍

PR #1975, opened 2022-03-30
feat(tools): 11ty plugins

bennypowers:
Moves the 11ty plugins from `/docs` to pfe-tools, for portability


zeroedin (approved):
Looks great @bennypowers dig the table-of-contents plugin.

brianferry (approved):

Issue #1153, opened 2020-10-22
feat: add ability to give pfe-modal a dark background


mwcz:
I'm attempting to make a barebones dark theme for arcade.redhat.com, and pfe-modal is presenting a small obstacle.  pfe-modal's background is colored with `--pfe-theme--color--surface--lightest` and it feels wrong to set that to a dark color.  Easy to work around, but the name of the background variable could be more abstract.

starryeyez024:
This is related to issue https://github.com/patternfly/patternfly-elements/issues/1090

PR #1976, opened 2022-03-30
fix!: rewrite color context

### LINKED ISSUE #1807: `<pfe-autocomplete on="dark">` renders results in white-on-white
## Description of the issue

if autocomplete is on="dark", the dropdown items will also be on="dark" i.e. have white text, but the bg color is white so you can't read them

### Impacted component(s)

- autocomplete

### Steps to reproduce

1. Go to https://patternflyelements.org/components/autocomplete/
2. execute in console `document.querySelector('pfe-band[color="lightest"]').setAttribute('color', 'darkest')`
3. type `Red` into the component
5. See error

### Expected behavior

Dropdown text should always be 'on' light, alternatively, dropdown should use context styles

### Notes
I may solve this on #1804 , undecided yet.

bennypowers:
---
"@patternfly/pfe-core": major
"@patternfly/pfe-sass": major
"@patternfly/pfe-styles": minor
"@patternfly/pfe-accordion": major
"@patternfly/pfe-band": major
"@patternfly/pfe-button": patch
"@patternfly/pfe-card": major
"@patternfly/pfe-clipboard": patch
"@patternfly/pfe-collapse": minor
"@patternfly/pfe-cta": major
"@patternfly/pfe-datetime": patch
"@patternfly/pfe-dropdown": patch
"@patternfly/pfe-health-index": patch
"@patternfly/pfe-icon-panel": patch
"@patternfly/pfe-icon": patch
"@patternfly/pfe-jump-links": major
"@patternfly/pfe-label": patch
"@patternfly/pfe-markdown": patch
"@patternfly/pfe-modal": major
"@patternfly/pfe-number": patch
"@patternfly/pfe-page-status": patch
"@patternfly/pfe-progress-indicator": patch
"@patternfly/pfe-readtime": patch
"@patternfly/pfe-select": patch
"@patternfly/pfe-tabs": patch
"@patternfly/pfe-toast": patch
---

### Fixes
- Fixes animation timings in pfe-accordion on browsers which don't support `computedStyleMap`
- Fixes context styles for slotted CTAs (provided `pfe.min.css` is loaded)
- Restores reload-on-save function to dev server

### BREAKING CHANGES:
- Renames `ContextTheme` type to `ColorTheme`
- Renames `ColorTheme` type to `ColorPalette`
- Moves both of the above from `core.ts` to `controllers/color-context.ts`
- Moves `controllers/color-context-controller.ts` to `controllers/color-context.ts`
- Removes `ColorContextController` (see below)
- Renames `color` attribute to `color-palette`
  affects `pfe-band`, `pfe-card`, `pfe-jump-links`, and `pfe-modal` explicitly, all others implicitly (see below)
  use `color-palette` instead or set `--context` css custom property
- Removes `pfe-contexts` mixin from `pfe-sass`, use controllers or decorators instead
- Removes `pfe-accordion--expanded` mixin from `pfe-sass` and inlines it
- Deprecates `context` attribute,
    Before;
    ```html
    <pfe-modal context="dark">
      <pfe-card color="lightest">...</pfe-card>
    </pfe-modal>
    ```

    After:
    ```html
    <pfe-modal color-palette="darkest">
      <pfe-card color-palette="lightest">...</pfe-card>
    </pfe-modal>

    <pfe-modal style="--context: dark;">
      <strong>🚨 Warning!</strong> May cause accessibility problems!
    </pfe-modal>
    ```

### New Features:

- Adds `ColorContextProvider` and `ColorContextConsumer` controllers
- Adds `@colorContextProvider` and `@colorContextConsumer` decorators
- Adds `context` option to `@pfelement` decorator
- Adds `className` and `attribute` options to `@pfelement` decorator
- Adds `color-palette` to all PFElements which don't explicitly opt-out
- Adds global (light DOM) link styling to `pfe.min.css`
- Adds `color-palette` support to `<pfe-modal>`

closes #1807

cc @zeroedin @heyMP @Nouveau


.changeset/context-controller-cavalcade.md
```diff
@@ -0,0 +1,26 @@
...
+---
+
+BREAKING CHANGES:
+- Deprecates `context` attribute,
+  use `color` instead or set `--context` css custom property
+- Renames `ContextTheme` type to `ColorTheme`
+- Renames `ColorTheme` type to `ColorPalette`
+- Removes `pfe-contexts` mixin
+
+```html
```
bennypowers, line 12: label these as before and after

elements/pfe-card/pfe-card.ts
```diff
@@ -55,6 +55,10 @@ export class PfeCard extends LitElement {
 
   static readonly styles = [style];
 
+  @colorContextProvider() color?: Color;
```
bennypowers, line 58: discussion: is it worth the boilerplate / bundle size cost to require these to be explicitly defined?

or is it enough to declare them in typescript / jsdoc comments?

.changeset/context-controller-cavalcade.md
zeroedin: Currently `pfe-label` and `pfe-icon` use the `color` attribute explicitly setting a color value which is not part of the context API.  In order to keep consistency across the API, I suggest we use `color-theme` or `color-palette`  instead of `color` or we would have to change the API for those components.

bennypowers: @heyMP i tend to agree, `color` is a bit overloaded here. WDYT?

bennypowers:
@zeroedin not ready to merge necessarily, but I'm eager to hear your thoughts on the changes so far

bennypowers:
Added color-palette support to modal 🎉 
Screen Shot 2022-04-01 at 15 31 30
cc @coreyvickery  does that track?

.changeset/context-controller-cavalcade.md
heyMP: Is it possible mark the `color` property as deprecated but not remove it yet?  I'm concerned that we've used it a bunch throughout our CMS.

bennypowers: yeah i think we could do that, but we should test for it as well

has the cms updated to pfe2 yet? @zhawkins WDYT?

elements/pfe-card/pfe-card.ts
heyMP: I *think* I would prefer them to be explicitly defined.

.changeset/context-controller-cavalcade.md
bennypowers: @heyMP if we take your suggestion from below to require context to be explicitly defined on each element then we can implement a deprecated `color` attr only for card and band, does that sound good?

elements/pfe-card/pfe-card.ts
bennypowers: Yeah, thinking it over this makes sense, @heyMP, so i did it.

One good indication that this is the right choice is that making these attributes explicit raised a dilemma that wasn't obvious previously:

In order to pass the unit tests, I had to set an initializer for both **card** and **band**'s `colorPalette` properties, defaulting to `'base'`. This essentially overrules it's context every time. in other words, since card and band always start with `color-palette="light"`, their `on` attributes are basically irrelevant.

You can see this in action on the [accordion demo](https://patternflyelements.org/components/accordion/demo/?accordion-with-history=2) if you open the "nested 2 card layout" accordion. Run this snippet in the devtools console, and you'll see that the cards retain their initial 'base' color palette despite the surrounding context being 'dark':

```js
document.getElementById('context-band').setAttribute('color', 'darkest')
```

(remember, this is on main branch, so what you're seeing there passed unit tests)

@coreyvickery is this intended? Should card and band always 'reset' their context unless explicitly provided with a palette (i.e. `darker`, `accent`, etc), or would we like card (and band) to respect their ancestors' color context?

marionnegp:
@bennypowers, I noticed some differences between the demos and what we have on patternflyelements.org. Apologies if these differences are just due to the demo environment.


- 600 weight looks bolder in the demo than on patternflyelements.org.

Cards
- CTA in footer looks like a regular link.
- Font weight of header needs to be 500.
- Imaged card with side overflow + header should have a padding-bottom of 16px for header text.

CTA
- default CTA has a gap of 8px between text and icon

Button
- medium secondary button should have blue (#06c) text
- large buttons look like they’re bolder and a slightly larger font size.

Band
- The asides in the default band, darker band, darkest band, and dark band with bg image aren’t lined up with the body copy.
- Can there be space above the CTA in the default band, darker band, and dark band with bg image?
- Heading text should be smaller (25.2px) and lighter (400).

Icon
- circled icons should be white

Jump links
- jump links looks like a regular unordered list instead of nav

bennypowers:
> @bennypowers, I noticed some differences between the demos and what we have on patternflyelements.org. Apologies if these differences are just due to the demo environment.

This is great feedback, thank you for taking the time.
 
>     * 600 weight looks bolder in the demo than on patternflyelements.org.

- [ ] I noticed this as well, I believe this is due to a difference in the versions of red-hat-font that are loaded. I _think_ that also accounts for the slightly different padding values for band.
  
> Cards
> 
>     * CTA in footer looks like a regular link.

- [x] This was caused by neglecting to import the cta definition in the card demo, that's been fixed

>     * Font weight of header needs to be 500.

- [ ] I've added `var(--pfe-card-header-font-weight, 500)`. if that name is unobjectionable, i'll add it to the docs as well. I had this issue on https://patternflyelements.org/components/card/demo/ as well so I think it's an unrelated bug

>     * Imaged card with side overflow + header should have a padding-bottom of 16px for header text.

- [ ] I had this issue on https://patternflyelements.org/components/card/demo/ as well, so it appears unrelated. I'd like to handle it in a separate issue.

> CTA
> 
>     * default CTA has a gap of 8px between text and icon

- [ ] Not sure what to make of this. perhaps this is also related to the different fonts?
Does the spec require the 8px gap? if so perhaps the css here should be revisited
 
> Button
> 
>     * medium secondary button should have blue (#06c) text

- [ ] In which context?

>     * large buttons look like they’re bolder and a slightly larger font size.

- [ ] Could be related to the red-hat-font

> Band
> 
>     * The asides in the default band, darker band, darkest band, and dark band with bg image aren’t lined up with the body copy.

- [ ] That looks like an issue in https://patternflyelements.org/components/band/demo/ as well, so let's deal with it separately.
 
>     * Can there be space above the CTA in the default band, darker band, and dark band with bg image?

- [ ] We can do anything we put our minds to 😉 Can you point me to a spec?

>     * Heading text should be smaller (25.2px) and lighter (400).

In this case, those look like light DOM styles applied by the consuming page.

- [ ] We should rewrite the demo to slot the h2 and p directly into the header slot, where we can specify the weights etc. Like above, this was a preexisting issue and we should deal with it separately.
 
 
> Icon
> 
>     * circled icons should be white

- [ ] I added some rules here based on which `color` attr is assigned. Is there a spec for that? would appreciate some a11y guidance. 
- [ ] 
Screen Shot 2022-04-04 at 16 57 37

 
> Jump links
> 
>     * jump links looks like a regular unordered list instead of nav

- [ ] The component definitions aren't being loaded here. I'll have to investigate why

elements/pfe-card/pfe-card.ts
coreyvickery: @bennypowers I feel like resetting would be best.

coreyvickery:
> @coreyvickery does that track?
@bennypowers Where did this come from? Was there a request to create a dark theme Dialog?

bennypowers:
> > @coreyvickery does that track?
> > @bennypowers Where did this come from? Was there a request to create a dark theme Dialog?

https://github.com/patternfly/patternfly-elements/issues/1153

coreyvickery:
> > > @coreyvickery does that track?
> > > @bennypowers Where did this come from? Was there a request to create a dark theme Dialog?
> 
> #1153

@bennypowers Ah, thanks.

Looks mostly good, I would make the following changes:
- [ ] Change the dialog background color to #212427 (for now, I want to see how this looks)
- [ ] Make sure the background container has 3px rounded corners if not already (kind of hard to tell)
- [ ] Change the headline font family to Red Hat Display
- [ ] Change the color of the close button to #d2d2d2
- [ ] Change the hover color of the close button to #fff
- [ ] Make sure the transparent background behind the dialog is #151515 and has 65% opacity

bennypowers:
>     * Change the dialog background color to #212427 (for now, I want to see how this looks)
That's for modal in dark context, or darker color-palette? a reminder: color-palette can be
- **lighest** - sets **light** context on children
- **light** - sets **light** context on children
- **base** - sets **light** context on children
- **darker** - sets **dark** context on children
- **dark** - sets **dark** context on children
- **accent** - sets **saturated** context on children
- **complement** - sets **saturated** context on children
>     * [ ]  Make sure the background container has 3px rounded corners if not already (kind of hard to tell)

coreyvickery:
@bennypowers Easy way to determine what hex code these values are? (lightest, light, etc.)?

bennypowers:
> @bennypowers Easy way to determine what hex code these values are? (lightest, light, etc.)?

there are defaults in _colors.scss but in many cases they're overridden

in the future i'm hoping design tokens will give us hsla, that'd be way easier to work with

marionnegp:
@bennypowers, I can set up separate issues for anything not mentioned below.

> > ```
> > * Font weight of header needs to be 500.
> > ```
> 
> * [ ]  I've added `var(--pfe-card-header-font-weight, 500)`. if that name is unobjectionable, i'll add it to the docs as well. I had this issue on https://patternflyelements.org/components/card/demo/ as well so I think it's an unrelated bug
>
This sounds good to me


> > CTA
> > ```
> > * default CTA has a gap of 8px between text and icon
> > ```
> 
> * [ ]  Not sure what to make of this. perhaps this is also related to the different fonts?
>   Does the spec require the 8px gap? if so perhaps the css here should be revisited
> 
I'll check with Corey about this. I don't know where the extra space comes from. I can only see it in the XD file.

> > Button
> > ```
> > * medium secondary button should have blue (#06c) text
> > ```
> 
> * [ ]  In which context?
> 
It's blue for all states. Does that answer the question?


> > ```
> > * Can there be space above the CTA in the default band, darker band, and dark band with bg image?
> > ```
> 
> * [ ]  We can do anything we put our minds to 😉 Can you point me to a spec?
> 
Should be 16px. The [band on patternflyelements.org](https://patternflyelements.org/components/band/) uses a row-gap.


> > Icon
> > ```
> > * circled icons should be white
> > ```
> 
> * [ ]  I added some rules here based on which `color` attr is assigned. Is there a spec for that? would appreciate some a11y guidance.
> 
> Screen Shot 2022-04-04 at 16 57 37
> 
We don’t have guidance for all of these on the site, but the moderate (#F0AB00) circle is the only one that would need a black (#151515) icon, according to the web contrast checker.

core/pfe-core/controllers/color-context.scss
```diff
@@ -16,7 +22,25 @@
   --pfe-broadcasted--link-decoration--visited: var(--pfe-theme--link-decoration--visited--on-dark, none);
 }
 
-:host([on=saturated]) {
```
zeroedin, line 19: I'm not sure if this a side effect fix or was intentional but before `:host([on="light|dark|saturated|..."]`  would override `--pfe--broadcasted--text` even when set as a custom CSS value on the parent `<pfe-band>` of a `<pfe-accordion-header>`  

Screenshot from current production [pfe-accordion/demo](https://patternflyelements.com/components/accordion/demo/):
image

[Deploy preview ](https://deploy-preview-1976--patternfly-elements.netlify.app/components/accordion/demo/)for this PR
image

Setting the value `--pfe-broadcasted-text` now seems to cascade correctly 🎉.

bennypowers: :fire: looks like a win to me

elements/pfe-label/demo/pfe-label.html
```diff
@@ -9,14 +9,12 @@ <h2 slot="header">Default</h2>
...
-<pfe-band color="lightest" size="small">
+<pfe-band color-palette="lightest" size="small">
     <h2 slot="header">Color Options</h2>
     <pfe-label color="blue">Blue label</pfe-label>
     <pfe-label color="green">Green label</pfe-label>
     <pfe-label color="orange">Orange label</pfe-label>
-    <pfe-label color="red">
-    Red <span class="visually-hidden-class">Hat</span>&nbsp;label
-    </pfe-label>
+    <pfe-label color="red">Red <span class="visually-hidden-class">Hat</span> label</pfe-label>
```
zeroedin, line 17: Removing the `&nbsp;` from this line causes the label to collapse on itself to `Redlabel` instead of `Red label` to fix wrap the change in another span.
```Suggestion
    <pfe-label color="red"><span>Red <span class="visually-hidden-class">Hat</span> label</span></pfe-label>
```

elements/pfe-accordion/demo/pfe-accordion.html
```diff
@@ -1,325 +1,322 @@
...
-            clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.</p>
-          <p>Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut
-            labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores
-            et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.</p>
-        </pfe-accordion-panel>
-      </pfe-accordion>
-    </article>
+  <pfe-band class="custom-styles" full-width id="custom">
+    <h2>Custom styles</h2>
+    <pfe-accordion id="custom">
```
zeroedin, line 258: Usage of the same `id` attr  twice, change the pfe-band id to `custom-band` instead.

```Suggestion
  <pfe-band class="custom-styles" full-width id="custom-band">
    <h2>Custom styles</h2>
    <pfe-accordion id="custom">
```

elements/pfe-autocomplete/pfe-autocomplete.ts
```diff
@@ -3,12 +3,16 @@ import type { DroplistSelectEvent, PfeSearchDroplist } from './pfe-search-dropli
 import { LitElement, html } from 'lit';
 import { customElement, property, state, query } from 'lit/decorators.js';
 
+import { ColorPalette, ColorTheme } from '@patternfly/pfe-core';
+
 import { ComposedEvent } from '@patternfly/pfe-core';
-import { pfelement, bound, observed } from '@patternfly/pfe-core/decorators.js';
+import { pfelement, bound, observed, colorContextConsumer, colorContextProvider, deprecation, } from '@patternfly/pfe-core/decorators.js';
```
zeroedin, line 9: Noticed a rouge `,` after I commit.  Needs removed.

```Suggestion
import { pfelement, bound, observed, colorContextConsumer, colorContextProvider, deprecation } from '@patternfly/pfe-core/decorators.js';
```

elements/pfe-card/pfe-card.scss
```diff
@@ -170,7 +170,7 @@ $size-small: (
     // Declare the header background color
     background-color: pfe-local(BackgroundColor, $region: header);
     color:            pfe-local($cssvar: Color, $region: header);
-    :host([on="dark"]) & {
+    :host([color-palette^="dark"]) & {
```
bennypowers, line 173: seems legit. I wanted to note that this makes sense b/c @coreyvickery specified that cards should reset their context, i.e. it shouldn't matter what the card is 'on', only it's specified palette.

elements/pfe-tabs/pfe-tab.scss
```diff
@@ -233,12 +233,24 @@
 }
 
 // SATURATED
+:host([variant="earth"][on="saturated"]) {
```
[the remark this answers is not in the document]
bennypowers: @coreyvickery IIRC you gave this a cursory pass on a call we had, so I'm marking this resolved.

AND if i'm wrong about that, OR if you want to make any design tweaks, they are very welcome, but let's do those in new issues.

brianferry (approved):
Looks good to me, we have done the visual regression tests and created issues off of the findings for those.

PR #1989, opened 2022-04-07
feat(create): `--monorepo` flag

bennypowers:
Running the generator in a monorepo, the user should be prompted for a scope name, and these files should be written

- tsconfig.json
- custom-elements-manifest.config.js
- package.json

Running the generator in a single package repo, the user should not be prompted for a scope name, and those three files should not be written

### DIFF
--- a/tools/create-element/main.ts
+++ b/tools/create-element/main.ts
@@ -3,11 +3,14 @@ import { generateElement } from './generator/element.js';
 import Chalk from 'chalk';
 import Yargs from 'yargs';
 import prompts from 'prompts';
+import { readJson } from './generator/files.js';
+import { join } from 'node:path';
 
 export interface BaseOptions {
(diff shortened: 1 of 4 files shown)


tools/create-element/main.ts
```diff
@@ -3,11 +3,14 @@ import { generateElement } from './generator/element.js';
 import Chalk from 'chalk';
 import Yargs from 'yargs';
 import prompts from 'prompts';
+import { readJson } from './generator/files.js';
+import { join } from 'node:path';
```
heyMP, line 7: what's this new magic?  `node:path`

bennypowers: https://nodejs.org/dist/latest-v16.x/docs/api/esm.html#node-imports

heyMP (approved):
👍

PR #1999, opened 2022-04-13
feat(tools): single file bundle for consumers

bennypowers:
## Problem

RHDS uses the `pfe-assets` 11ty plugin from pfe-tools. That plugin in turn generates a single-file bundle of all of PFE, minified, via `singleFileBundle` from pfe-tools.

The problem is that RHDS doesn't have all of PFE installed (and shouldn't have to). Otherwise, users of pfe-tools would always have to have all the pfe packages installed or face fatal errors when building their 11ty sites, and that'd be awkward to say the least.

## Solution

Make deciding which additional packages to bundle the consumer's problem

This PR makes it so that `singleFileBundle`:
1. always bundles every package in the current [mono]repo
2. allows the caller to provide a list of additional packages to bundle (by name)
3. exposes that as an option to pfe-assets 11ty plugin.

Side bonus: removes the on-disk `components.ts` file, just passes it in memory to esbuild instead

## Example
So RHDS would add that plugin like so:

```js
// get all available PFE 2.0 packages
// (except styles, which is only css)
// and bundle them when building the site

const { dependencies } = require(path.join(__dirname, 'package.json'));
const additionalPackages = [
  ...Object.entries(dependencies)
    .map(([k, v]) => k.startsWith('@patternfly') && v === 'next' ? k : false)
    .filter(x => x && x !== '@patternfly/pfe-styles'),
];

/** Copy and manage site assets from the monorepo */
eleventyConfig.addPlugin(pfeAssetsPlugin, {
  prefix: 'rh',
  additionalPackages,
});
```


tools/pfe-tools/esbuild.ts
```diff
@@ -66,6 +66,8 @@ const ALWAYS_EXCLUDE = [
   'pfe-styles',
 ];
 
+const exists = (filePath: string) => stat(filePath).then(() => true, () => false);
```
heyMP, line 69: Is using the `stat()` method better than fs.exists?

tools/pfe-tools/esbuild.ts
```diff
@@ -106,36 +108,56 @@ export function getBasePlugins({ minify }: { minify?: boolean } = {}) {
...
-      COMPONENT_ENTRYPOINTS_CACHE.set(cacheKey, tmpfile);
-      return tmpfile;
+      const imports = await componentDirs.reduce(async (last, dir) => {
+        const acc = await last;
+        const elementPath = join(REPO_ROOT, 'elements', dir);
+        const packageJsonPath = join(elementPath, 'package.json');
+        const isPackage = await exists(packageJsonPath);
+        if (isPackage) {
+          const { name } = JSON.parse(await readFile(packageJsonPath, 'utf8'));
+          return `${acc}\nexport * from '${name}';`;
```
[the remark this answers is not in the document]
bennypowers: the `\n` part is a newline escape
so this reducer takes each component dir, scans it for a package.json, then adds the appropriate export statement on a new line, beginning with the `additionalImports` export statements which were similarly generated 10 lines previously

tools/pfe-tools/esbuild.ts
```diff
@@ -106,36 +108,56 @@ export function getBasePlugins({ minify }: { minify?: boolean } = {}) {
...
-    const prefix = fileURLToPath(new URL('./demo', import.meta.url)).replace(REPO_ROOT, '');
     const result = await esbuild.build({
       absWorkingDir: REPO_ROOT,
       allowOverwrite: true,
       bundle: true,
-      entryPoints: [await componentsEntryPoint({ prefix })],
+      stdin: {
+        contents: await componentsEntryPoint({ additionalPackages: options?.additionalPackages }),
+        resolveDir: REPO_ROOT,
+        sourcefile: 'components.ts',
```
heyMP, line 158: This is now a virtual file, correct?

heyMP (approved):
Ballin!  Approving, just had some pick your brain questions.

tools/pfe-tools/esbuild.ts
bennypowers: yes, [fs.exists is deprecated](https://nodejs.org/dist/latest-v16.x/docs/api/fs.html#fsexistspath-callback)

tools/pfe-tools/esbuild.ts
bennypowers: indeed it is, and isn't that lovely? [the `sourcefile` option is used for stack traces and sourcemaps](https://esbuild.github.io/api/#stdin)

tools/pfe-tools/esbuild.ts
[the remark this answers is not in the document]
bennypowers: this reducer would probably be a bit clearer with [dedented template strings](https://github.com/tc39/proposal-string-dedent) but, you know.

brianferry (approved):

PR #2000, opened 2022-04-15
Version Packages (next)

bennypowers:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

⚠️⚠️⚠️⚠️⚠️⚠️

`main` is currently in **pre mode** so this branch has prereleases rather than normal releases. If you want to exit prereleases, run `changeset pre exit` on `main`.

⚠️⚠️⚠️⚠️⚠️⚠️

# Releases
### Major Changes

-   6a2a0407: [View commit message here](https://gist.github.com/heyMP/200fc0b840690541475923facba393ab)

 ## @patternfly/pfe-sass@2.0.0-next.5

### Major Changes

-   6a2a0407: [View commit message here](https://gist.github.com/heyMP/200fc0b840690541475923facba393ab)

 ## @patternfly/pfe-accordion@2.0.0-next.4

### Major Changes

-   6a2a0407: [View commit message here](https://gist.github.com/heyMP/200fc0b840690541475923facba393ab)

### Patch Changes

-   Updated dependencies [6a2a0407]
-   Updated dependencies [0e9a4130]
    -   @patternfly/pfe-core@2.0.0-next.4
    -   @patternfly/pfe-collapse@2.0.0-next.2
    -   @patternfly/pfe-icon@2.0.0-next.2

 ## @patternfly/pfe-band@2.0.0-next.2

### Major Changes

-   6a2a0407: [View commit message here](https://gist.github.com/heyMP/200fc0b840690541475923facba393ab)

### Patch Changes

-   Updated dependencies [6a2a0407]
    -   @patternfly/pfe-core@2.0.0-next.4

 ## @patternfly/pfe-card@2.0.0-next.3

### Major Changes

-   6a2a0407: [View commit message here](https://gist.github.com/heyMP/200fc0b840690541475923facba393ab)

### Patch Changes

-   Updated dependencies [6a2a0407]
    -   @patternfly/pfe-core@2.0.0-next.4

 ## @patternfly/pfe-cta@2.0.0-next.3

### Major Changes

-   6a2a0407: [View commit message here](https://gist.github.com/heyMP/200fc0b840690541475923facba393ab)

### Patch Changes

-   f66ba580: Add color context support via explicitly defined attributes. See <https://github.com/patternfly/patternfly-elements/pull/1976>
[...]

PR #2001, opened 2022-04-17
fix(cta): explicit color context support

bennypowers:
Color context is broken for CTA on #2000, see https://deploy-preview-2000--patternfly-elements.netlify.app/


brianferry (approved):
Change makes sense, we should do some testing around the formatting that was present in this changeset and see if we can mimic the behavior we were seeing in the actions pipeline.  Otherwise we'll just need to know what the max length is for a changeset.

PR #2005, opened 2022-04-18
fix: cta styling removing color context provider

zeroedin:
## Problem

`pfe-cta` had `color-palette` attribute and the `@colorContextProvider()` decorator applied to it inadvertently causing the variants to display colors incorrectly when inheriting from their parent color context and applying their own overrides.

### Related issues

- #2002

## Solution

Remove the `@colorContextProvider()` decorator and revert back to `color` to a standard property.   Also removing the `@deprecated` decorator as it is in this case not a part of the color context controller.  Update the demo HTML and component scss to reflect this reversion.  

#### Browser requirements

Your component should work in all of the following environments:

- [ ] Latest 2 versions of Edge
- [ ] Internet Explorer 11 (should be useable, not pixel perfect)
- [ ] Latest 2 versions of Firefox (one on Mac OS, one of Windows OS)
- [ ] Firefox 78 (or latest version for Red Hat Enterprise Linux distribution)
- [ ] Latest 2 versions of Chrome (one on Mac OS, one of Windows OS)
- [ ] Latest 2 versions of Safari
- [ ] Android mobile device (such as the Galaxy S9)
- [ ] Apple mobile device (such as the iPhone X)
- [ ] Apple tablet device (such as the iPhone Pro)

### Ready-for-merge Checklist

- [ ] Expected files: all files in this pull request are related to one request or issue (no stragglers or scope-creep).
- [ ] Tests have been updated to cover these changes.
- [ ] Browser testing passed.
- [ ] Changelog updated.
- [ ] Documentation (README.md, WHY.md, etc.) updated or added.
- [ ] Link to the demo recording: []()
- [ ] Approved by designer.

### Merging

Please **squash** when merging and ensure your commit message uses [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) formatting.

**Be sure to share your updates with the [patternfly-elements-contribute@redhat.com](mailto:patternfly-elements-contribute@redhat.com) mailing list!**

Issue #2002, opened 2022-04-18
pfe-cta color context attribute color-palette incorrectly applies styles to variants


zeroedin:
pfe-cta had the color context attribute `color-palette` applied to it as a color context provider.  I believe this might have done inadvertently.

latest example:
![image](https://user-images.githubusercontent.com/13666/163856524-c10647a3-4ec2-4363-b8f0-dabea0cbc2b5.png)

pfe-cta I don't believe needs to be a provider of color context, instead, its prior `color` attribute was separate from the color context code base instead working as a for lack of better terminology tertiary theme attribute.

pfe-cta has a `priority` as well as `variant` attribute which affects its styling.  The`color` attribute was a third attribute to override some of the styles applied by the other two to give more variation.  If I understand it correctly it was not the being used in the same context as the `on` and now `color-palette` as a context for child elements as pfe-cta really only accepts `<a>` and `<button>` tags as the light dom elements.

pfe-cta should reflect the same styles as in the example below:

v1.x example:
![image](https://user-images.githubusercontent.com/13666/163857259-4fbed8b4-6e1d-4703-9e65-218d68ec95c6.png)

My suggestion is to revert back to the `color` attribute as an explicit attribute such as in `pfe-icon` and `pfe-label`, removing the `@colorContextProvider` decorator and the `@deprecation` of the color attribute.  Reverting the demo HTML and scss changes for the component.

bennypowers:
Closed by #2006 
If there are any hanging issues here, let's open a new ticket

PR #2006, opened 2022-04-19
[draft] fix: tied legacy pfe-cta css vars to the new color-context vars

heyMP:
## Problem / Solution

This illustrates how to tie the legacy CSS vars for pfe-cta into the new color context controller `--pfe-context-background-color: var(--pfe-cta--BackgroundColor);`

The `_sharedStyles.scss` generates all of the color-palette and priority overrides via css custom properties.  Since the color context provider sets background-color and only background-color, we have to specifically use that custom property to override background-color.

`color-context.scss`
```scss
:host(:is([color-palette])) {
  background-color: var(--pfe-context-background-color, var(--pfe-theme--color--surface--base));
}
```

### Alternatives

- #2005 
- We could also remove the `background-color` declaration in `color-context.scss` and put in the docs that you also have to implement the `--pfe-context-background-color` manually.

### Related issues

- #2002 

### Preview

- [Link](https://deploy-preview-<pr_number>--happy-galileo-ea79c4.netlify.app/examples/) 

### What has changed and why

- 

### Testing instructions

- [ ] *Test case #1*
  1. Go here
  2. Do this
- [ ] *Test case #2*
  1. Go here
  2. Do this

#### Browser requirements

- [ ] Internet Explorer 11 (should be useable, not pixel perfect)
- [ ] Latest 2 versions of Firefox (one on Mac OS, one of Windows OS)
- [ ] Firefox 78 (or latest version for Red Hat Enterprise Linux distribution)
- [ ] Latest 2 versions of Chrome (one on Mac OS, one of Windows OS)
- [ ] Latest 2 versions of Safari
- [ ] Android mobile device (such as the Galaxy S9)
- [ ] Apple mobile device (such as the iPhone X)
- [ ] Apple tablet device (such as the iPhone Pro)

### Ready-for-merge Checklist

- [ ] Expected files: all files in this pull request are related to one request or issue (no stragglers or scope-creep).
- [ ] Tests have been updated to cover these changes.
- [ ] Browser testing passed.
- [ ] Changelog updated.
- [ ] Documentation (README.md, WHY.md, etc.) updated or added.
- [ ] Approved by designer.

### Merging


elements/pfe-cta/demo/pfe-cta.html
```diff
@@ -178,9 +178,9 @@ <h4 style="color:white;">Color override</h4>
       <pfe-cta priority="primary" color-palette="lightest"><a href="#primary">Primary lightest</a></pfe-cta>
       <pfe-cta priority="secondary" variant="wind"><a href="#secondary-wind">Secondary + wind variant</a></pfe-cta>
     </div>
-  </div>
+  </pfe-band>
 
-  <div class="custom-dark-band resize">
+  <pfe-band color-palette="darkest" class="custom-dark-band resize">
```
zeroedin, line 183: Forgive me if I'm wrong, but isn't the point of the `.custom-dark-band` class to pass the `--context` CSS property and have children reflect that without the need for a parent with a provider?  If you use a `pfe-band` and `color-palette` attribute/provider then you are bypassing that `--context`  usage?

zeroedin: After futher digging I read this comment in `ColorContextConsumer` class cole-context.ts line:211-216

```js
  /**
   * When a color context consumer connects,
   * it requests colour context from the closest context provider,
   * then updates it's host's `on` attribute
   */
 hostConnected() {
```

So the idea of setting `--context` may not work as it will always go further up the chain to find a provider when it is a consumer.  Maybe my understanding of what `--context` is supposed to do was incorrect.

bennypowers: it should get `--context` based on the state of the css cascade wherever the element happens to be at the moment

bennypowers: @heyMP ideally this PR wouldn't change the demo files, what to demonstrate the correctness of the code changes

bennypowers:
I pushed some commits which fix up the context variables for CTA by explicitly defining them. Is is the prettiest code? No. However I'd prefer not to dwell on it, since I'm fixing to refactor this CSS at some point anyway. It would be nice to add some regression tests on this branch once we reach consensus (through manual testing) that it passes acceptance criteria.

I haven't yet addressed the custom `--context` div wrapper, will tackle that next.

I tested this by running two separate dev servers, one with commit c10f6783937626837422c3cc521f0a6c27ae5d77 checked out (i.e. before the context refactor) and one with this branch checked out.

In these screenshots, the left hand side is the old, pre-refactor version, and the right hand side is the PR branch
Screen Shot 2022-04-25 at 11 46 49
Screen Shot 2022-04-25 at 11 47 04

bennypowers:
OK the issue with the `--context` div is that the refactor moved responsibility for deriving the context from the css var to the provider only. This was a performance enhancement as much as it was about simplifying the code. So let's reopen that discussion: Should we stop supporting `--context` without a provider element?

Opposed:
- That means changing the pfe-cta demo file, which is another way of saying making a breaking change to our acceptance criteria
- There may be cases of such 'in the wild' which will break

In Favour:
- `--context` is not documented on patternflyelements.org
- Context is JS-only anyways, so we don't lose by doubling down on custom element registration
- Allows us to simplify context code and improve performance

elements/pfe-cta/demo/pfe-cta.html
```diff
@@ -180,7 +180,7 @@ <h4 style="color:white;">Color override</h4>
     </div>
   </div>
 
-  <div class="custom-dark-band resize">
+  <div color-palette="darkest" class="custom-dark-band resize">
```
zeroedin, line 183: I think `color-palette` attribute was supposed to be removed in the revert commit.

```Suggestion
<div class="custom-dark-band resize">
```

zeroedin:
I think the compromise with `on` might fit the bill here.  While it isn't ideal to have to set the attribute individually when used inside a parent that isn't a color context provider, it provides flexibility and performance by not having to need additional js.

elements/pfe-cta/demo/pfe-cta.html
heyMP: I agree with removing the `--context`.  These approaches sound resonable to me.  In the future, we could always provide a dedicated declarative component for opting in to the ContextProvider / ContextConsumer.

bennypowers:
Ok, taking @heyMP and @zeroedin 's comments into consideration, what I will do is:
- explicitly add `on` to demo examples
- add theming examples to demo pages which place CTAs on context providers, but customize documented theme vars
- add regression tests for CTA

There are a few other examples in the repo where `--context` is set, so i'll examine those, and if I find the same issue there i'll take the same approach

bennypowers:
@heyMP @zeroedin  
- `--context` by itself is no longer supported (it was never documented). Authors that want to 'fake' a context should provide both `--context` and `on` to each element. 
- `on` set in HTML now overrides color context for providers. in other words, manually setting `on` is an escape hatch which opts out of color context calculations.
  - [docs](https://patternflyelements.org/theming/#context) do mention the `on` attribute, but imply that it is 'behind the scenes'. perhaps we should make this explicit in docs?
  - [elsewhere in docs](https://patternflyelements.org/theming/colors/#:~:text=if%20needed%2C%20you%20can%20override%20the%20on%20attribute%20by%20manually%20applying%20the%20pfe-theme%20attribute%20with%20one%20of%20the%20following%20values%3A%20light%2C%20dark%2C%20saturated.) `pfe-theme` is documented as an override for `on`, this should be updated, right?
- adds some test helpers for validating css color strings

bennypowers (approved):
Less griping, tinker more!

Issue #1984, opened 2022-04-05
Padding in imaged card demo


marionnegp:
@bennypowers, [imaged card with side overflow + header in demo](https://patternflyelements.org/components/card/demo/) should have a padding-bottom of 16px for header text.

bennypowers:
It appears this issue is no longer relevant. Please open a new issue if it is still relevant.

PR #2015, opened 2022-04-28
style(card): :art: fixing img with overflow "top" not getting correct top padding

brianferry:
Adding padding-top to img tag when the element has a header and a slotted image with an overflow of top.  Padding-top will be set to pfe-local(spacing);

### Related issues

- (#1984) Padding in imaged card demo

### What has changed and why

Added css to the pfe-card.scss 

- In the body section if a header is present and a slotted img is present with an attribute of overflow="to" we set the padding-top to pfe-local(spacing)

### Testing instructions

Update can be viewed in the pfe-card demo page.


bennypowers:
FYI, the `style` commit type is *not* for CSS changes, but rather for _code style_ refactors, e.g. changing whitespace, linting files, etc. This is confusing, but that's how it is:

https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type

elements/pfe-card/pfe-card.scss
```diff
@@ -231,6 +231,12 @@ $size-small: (
       --pfe-card__overflow--MarginBottom: calc(#{pfe-local(spacing--vertical)} * -1);
     }
 
+    &:is(.has-header){
+      ::slotted(img[overflow~="top"]){
```
bennypowers, line 235: Looks fine. Arguably the spacing should be on the header not the image, but i'm not overly bothered

bennypowers (changes requested):
Please add a changeset

bennypowers:
@marionnegp PTAL at https://deploy-preview-2015--patternfly-elements.netlify.app/components/card/demo/

marionnegp:
Padding looks good to me!

elements/pfe-card/pfe-card.scss
brianferry: Yeah I went back and forth on this and as the heading already had what looked like spacing added onto the bottom of it and the image was the element ignoring that I opted to add the css onto the image itself.

brianferry:
Added the changeset @bennypowers

bennypowers (approved):

PR #2019, opened 2022-05-03
fix(tools): pass options to dev server

bennypowers:
## What I did
pass the user's options to the pfe dev server plugin


zeroedin (commented):
Looks good, did a quick test with pfeDevServerConfig options, and seemed to pass into the dev demo with no issues.

zeroedin (approved):

PR #2027, opened 2022-05-10
fix(create): fix templates for single-package repos

bennypowers:
## What I did

- Generator now provides the option to choose the css flavour for the element, e.g. `scss`, `css`, or `postcss`
- Generator templates for single-package repos (i.e. RHDS) are fixed. internally, generator hardcodes some values for RHDS, this can be updated later as it becomes an issue

## Testing Instructions
- Confirm that the generator works in pfe
```bash
npm run build -w @patternfly/create-element
```
```bash
npm run new
```
```bash
npm start
```
Confirm that the new element works
- Confirm that the generator works in rhds
```bash
cd ../path/to/red-hat-design-system
```
It will be necessary to copy over some of create's dependencies to the rhds node_modules folder.
You may need to update these paths to suit your local setup
```bash
rm -rf node_modules/@patternfly/create-element/                                           
cp -R ../../patternfly/patternfly-elements/tools/create-element/ node_modules/@patternfly/create-element
cp -R ../../patternfly/patternfly-elements/node_modules/execa/ node_modules/execa
cp -R ../../patternfly/patternfly-elements/node_modules/npm-run-path/ node_modules/npm-run-path 
npx @patternfly/create-element -n rh-jazz-hands --overwrite
```


brianferry:
@bennypowers - I am able to run the scripts as they are put in the doc and everything looks like it is working for me.  In the `Red Hat Design System` repo I needed to add the following to my `package.json` to switch it from thinking it was not a monorepo to thinking that it was.  

```
"workspaces": [
    "./elements/*"
  ],
```

qq - Can you manually overwrite this setting if you wanted when you were creating a new element?  Like if for some reason in rhds we wanted to make one component that had scss would we be able to pass in `--css=false` or something similar?

bennypowers:
RHDS isn't a monorepo tho, so you shouldn't need to do that.

heyMP:
@brianferry if you're comfortable with this PR you can probably hit the approve button on this.  It's set to automerge. 👍

brianferry:
This worked for me in RHDS but for PFE I am getting the following error: 

image

I know we've seen this before where the demo link does not line up with where it's at locally but just wanted to make sure we're not reintroducing anything here.

RHDS does not get this same error for me,

bennypowers:
@brianferry looks probably like the in the monorepo path, it should run `npm i` after scaffolding

brianferry:
> @brianferry looks probably like the in the monorepo path, it should run `npm i` after scaffolding

Yep, looks like that was commented out in the create-element file.  I added it back in with the last commit.  If that looks good to you then I think we can merge

bennypowers:
shipit!

brianferry (approved):
LGTM, great work @bennypowers

PR #2030, opened 2022-05-12
feat(tools): single file build entrypoints

bennypowers:
## What I did

Added some new APIs for `@patternfly/pfe-tools/esbuild.js` to facilitate building daughter single-package repos (i.e. RHDS)

`singleFileBuild` now scans `node_modules` for installed
`@patternfly/pfe-*` packages (except `core`, `tools`, `sass`, and
`styles`) and generates an entrypoint file for them.

Users can alternatively pass `componentsEntryPointsContent`, which is
the string contents of a javascript module that exports the desired
components.

These changes make using `singleFileBuild` more useful and ergonomic for
daughter repositories (e.g. RHDS)

## Testing Instructions
1. `npm run build` <- verify that the minified bundle is good by inspecting the built docs site (can verify this in the deploy preview)
2. patch this version into RHDS (on footer branch) and verify that the 11ty site builds and that the generated `rhds.min.js` in works

### DIFF
--- a/tools/pfe-tools/package.json
+++ b/tools/pfe-tools/package.json
@@ -107,6 +107,7 @@
     "esbuild-plugin-lit-css": "^2.0.0",
     "esbuild-plugin-minify-html-literals": "^1.0.1",
     "eslint": "^8.14.0",
+    "execa": "^6.1.0",
     "glob": "^8.0.1",
     "html-include-element": "^0.3.0",
     "koa-send": "^5.0.1",
--- a/.changeset/calm-items-cover.md
+++ b/.changeset/calm-items-cover.md
@@ -1,0 +1,36 @@
+---
+"@patternfly/pfe-tools": minor
+---
+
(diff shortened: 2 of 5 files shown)


tools/pfe-tools/package.json
```diff
@@ -107,6 +107,7 @@
     "esbuild-plugin-lit-css": "^2.0.0",
     "esbuild-plugin-minify-html-literals": "^1.0.1",
     "eslint": "^8.14.0",
+    "execa": "^6.1.0",
```
zeroedin, line 110: I don't see this being used anywhere in this PR.  Is this a leftover from another change either not included or removed?

bennypowers: it's necessary to avoid an error in RHDS, which has it's own version of execa which differs by at least one semver major

brianferry (approved):
Looks good to me, was able to bring it into rhds and was able to build after bringing over execa and npm-run-all dependencies.

PR #2036, opened 2022-05-19
feat(modal): PF1:1

bennypowers:
## What I did
- deprecate the `width` attribute in favour of `variant`
- implement many `--pf-` css variables
- remove _all_ `--pfe-` css variables

## Testing Instructions
- Load up the deploy preview for this branch
- Load up [PatternFly v4 Modal](https://www.patternfly.org/v4/components/modal/)
- Compare and contrast
- Read over the [docs](https://deploy-preview-2036--patternfly-elements.netlify.app/components/modal/) and see if anything's missing or no longer relevant

## Notes to reviewers
This PR is *BREAKING*. It aligns pfe-modal with pfv4. the previously-pfe-specific styles and APIs will be moved over to https://github.com/RedHat-UX/red-hat-design-system/pull/288
So dip into the percy tests to see if anything wild broke

### DIFF
--- a/elements/pfe-modal/demo/demo.css
+++ b/elements/pfe-modal/demo/demo.css
@@ -1,3 +1,25 @@
+@font-face {
+  font-family: "RedHatDisplayUpdated";
+  src: url(https://patternfly.org/v4/fonts/RedHatDisplay-updated-Regular.woff2) format("woff2");
+  font-style: normal;
+  font-weight: 300;
+  text-rendering: optimizeLegibility;
+}
+@font-face {
+  font-family: "RedHatDisplayUpdated";
+  src: url(https://patternfly.org/v4/fonts/RedHatDisplay-updated-Medium.woff2) format("woff2");
+  font-style: normal;
+  font-weight: 400;
+  text-rendering: optimizeLegibility;
+}
+@font-face {
+  font-family: "RedHatDisplayUpdated";
+  src: url(https://patternfly.org/v4/fonts/RedHatDisplay-updated-Bold.woff2) format("woff2");
+  font-style: normal;
+  font-weight: 700;
+  text-rendering: optimizeLegibility;
+}
+
 :host {
   --pfe-modal--MinWidth: 800px;
   --pfe-modal--MaxWidth: 800px;
--- a/elements/pfe-modal/demo/pfe-modal.html
+++ b/elements/pfe-modal/demo/pfe-modal.html
@@ -157,8 +157,8 @@
 </pfe-band>
 
 <pfe-band color-palette="lightest" size="small">
-  <h2 slot="header">pfe-modal-width CSS Custom Property</h2>
-  <pfe-modal style="--pfe-modal-width: 50%">
+  <h2 slot="header"><code>--pf-c-modal-box--Width</code> CSS Custom Property</h2>
+  <pfe-modal style="--pf-c-modal-box--Width: 50%">
     <h2 slot="header">Width 50% header</h2>
     <p>Lorem ipsum dolor sit amet, <a href="#foo">consectetur adipisicing</a> elit, sed do eiusmod tempor incididunt
       ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut
--- a/elements/pfe-modal/demo/pfe-modal.html
+++ b/elements/pfe-modal/demo/pfe-modal.html
@@ -7,7 +7,7 @@
 
 <pfe-band color-palette="lightest" size="small">
   <h2 slot="header">Basic</h2>
-  <pfe-modal>
+  <pfe-modal trigger="simple-modal-header-trigger">
     <h2 slot="header">Simple modal header</h2>
     <p>Lorem ipsum dolor sit amet, <a href="#foo">consectetur adipisicing</a> elit, sed do eiusmod tempor incididunt
       ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut
@@ -20,16 +20,16 @@
     <pfe-button variant="secondary" slot="footer">
       <button>Cancel</button>
     </pfe-button>
-    <pfe-button slot="trigger">
-      <button>Show modal</button>
-    </pfe-button>
   </pfe-modal>
+  <pfe-button id="simple-modal-header-trigger" slot="footer">
+    <button>Show modal</button>
+  </pfe-button>
 </pfe-band>
 
 <pfe-band color-palette="lightest" size="small">
   <h2 slot="header">With description</h2>
-  <pfe-modal>
-    <h2 slot="header">Simple modal header</h2>
+  <pfe-modal trigger="with-description-trigger">
+    <h2 slot="header">Modal with description</h2>
     <p slot="description">A description is used when you want to provide more info about the modal than the title is
       able to describe. The content in the description is static and will not scroll with the rest of the modal body.
     </p>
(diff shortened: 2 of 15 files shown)


zeroedin:
PF Modal doesn't close on a click outside of clicking the dismiss icon.  Should this interaction also be removed?

elements/pfe-modal/pfe-modal.scss
```diff
@@ -1,154 +1,181 @@
...
+}
+
+
+[part=content] {
+  overflow-y: auto;
+  overscroll-behavior: contain;
+
+  max-height: var(--pf-c-modal-box--MaxHeight,
+    calc(100% - var(--pf-global--spacer--2xl,
+      3rem)));
```
zeroedin, line 98: In testing, it seems the overflow wasn't working correctly.  Here is what I did to fix it.

Add `style="--pf-c-modal-box--MaxWidth: 320px;
    --pf-c-modal-box--Width: 100%;"` to the overflow example demo.  

Then change 100% to 100vh for the max-height calc.

```Suggestion
max-height: var(--pf-c-modal-box--MaxHeight,
    calc(100vh - var(--pf-global--spacer--2xl,
      3rem)));
```

elements/pfe-modal/demo/pfe-modal.html
```diff
@@ -5,115 +5,290 @@
...
+  </pfe-modal>
+</pfe-band>
+
+<!-- TODO: wizard -->
+<!-- TODO: dropdown that can break out -->
+<!-- TODO: help popover -->
+<!-- TODO: with form fields -->
+
+
+<pfe-band color-palette="lightest" size="small">
```
zeroedin, line 251: This comment goes along with the comment added to the *.scss about fixing the overflow content example.

```Suggestion
<pfe-band color-palette="lightest" size="small" style="--pf-c-modal-box--MaxWidth: 320px; --pf-c-modal-box--Width: 100%;">
```

bennypowers: When I was testing I preferred to just manipulate the viewport. PF4 demos don't have any controls or built in styles for that. Is that enough for you?

bennypowers:
> PF Modal doesn't close on a click outside of clicking the dismiss icon/cancel button. Should this interaction also be removed?

🤔 I don't wanna, but you're right ;)

one of the limitations of lit is that it's easy to compose a superclass' template into your own, but harder to make fine-grained adjustments to super's template:
```html
<button id="close-button" @click=${this.close}>x</button>
```
Options:
- `renderCloseButton(): TemplateResult` meh, not ideal
- `@click=${this.onClickCloseButton}` and make it protected. Ok, I guess
- `static closeOnOutsideClick = false`, a blunt instrument, but does the job

elements/pfe-modal/demo/demo.css
```diff
@@ -1,3 +1,25 @@
+@font-face {
```
bennypowers, line 1: Maybe we should leave this font loading up to the host page?

bennypowers:
I added some protected optional override methods for rendering the various content slots. It's not the most elegant thing in the world, but it does the job. If we have to do this elsewhere, let's stick to the `renderThingnameSlot` convention.

The reason we need this is to override the colour context for the modal content, without overriding the color context for the slotted trigger.

We could avoid that by not allowing slotted triggers, and using a private context provider, which I'm totally fine with - happy to revert that change if we choose to disallow slotted triggers and use the `trigger` attr instead

wesruv:
Just took a peek from my phone, wonder if there's any efficiency from using a `<dialog>` element, but I'm guessing you may have tried? If so I'd be curious what your experience was.

bennypowers:
> Just took a peek from my phone, wonder if there's any efficiency from using a `<dialog>` element, but I'm guessing you may have tried? If so I'd be curious what your experience was.

I'm definitely interested in exploring this, but it seems to me a drop out of scope at the moment.

elements/pfe-modal/demo/pfe-modal.html
zeroedin: Works for me.

zeroedin (commented):
This looks good to me with the previous suggestions fixed.

heyMP:
> We'll need to put in some escape hatches for `rh-dialog`:
> 
> [According to ux-dot, dialogs should always have lightest color-palette](https://ux.redhat.com/components/modal/#style), in other words they should reset the colour context for their contents.

I like what you have now with the renderSlot overrides but just thinking out loud:

- We could make the trigger "detachable" where it could be moved outside of the parent container.
```
<pfe-band color-palette="darkest">

  <rh-dialog trigger-id="dialog-trigger">
    The footer button needs to inherit the light context of the dialog,
    since it's projected into the dialog window
    <pfe-button slot="footer"><button>Confirm</button></pfe-button>

    But the trigger button renders visually inside the dark band
    So it mustn't inherit the dialog's context
  </rh-dialog>
  
  <pfe-cta id="dialog-trigger"><button>Open</button></pfe-cta>

</pfe-band>
```

elements/pfe-modal/pfe-modal.ts
```diff
@@ -39,48 +39,78 @@ export class ModalOpenEvent extends ComposedEvent {
...
-export class PfeModal extends LitElement {
+export class PfeModal extends LitElement implements HTMLDialogElement {
   static readonly shadowRootOptions = { ...LitElement.shadowRootOptions, delegatesFocus: true };
 
   static readonly version = '{{version}}';
 
   static readonly styles = [style];
 
+  /** Should the dialog close when user clicks outside the dialog? */
+  protected static closeOnOutsideClick = false;
```
heyMP, line 86: should this default to true?

heyMP: sorry, just saw that this is the behavior of PF.

elements/pfe-modal/demo/demo.css
bennypowers: ok i just now realized that I'm *doing* this in the host page 🙈

bennypowers:
Ok this is what I want to do:
1. Remove the `trigger` slot. This is a breaking change, but the benefits outweigh the losses IMO and migration to `trigger` attr isn't that hard (all in HTML, no need to call JS methods)
2. Thereby remove the render overrides
3. Defer the question on what to do about non-modal dialogs until RHDS (or someone else) asks it (YAGNI)

bennypowers:
@wesruv I looked into it a drop. Def. want to adopt `<dialog>` at some point, but not just yet. I believe there are some a11y concerns with certain AT UAs not being ready for `<dialog>`

elements/pfe-modal/pfe-modal.ts
```diff
@@ -74,7 +73,7 @@ export class ModalOpenEvent extends ComposedEvent {
  * @cssprop {<color>} --pf-c-modal-box--BackgroundColor - {@default #fff}
  * @cssprop --pf-c-modal-box__title--FontFamily - default font family for header-slotted headings
  */
-@customElement('pfe-modal') @pfelement()
+@customElement('pfe-modal')
```
heyMP, line 76: Do we need `@pfelement()`?

bennypowers: `@pfelement` does 3 things:
1. adds `PFElement` className and `pfelement` attr to the host element on connect
2. adds optional perf tracking
3. enqueues body autoreveal

(1) PF doesn't do, not sure why we'd need to here, and if it is needed, let the user add them in their HTML
(2) likewise, this isn't a pf4 thing, so why would we need it? we can solve that problem otherwise in rhds, I'd think, and leave `<pfe-*>` for UI
(3) same same, this is kind of an app concern, we don't need component code to get involved

heyMP: Gotcha. 👍

wesruv:
> I'm definitely interested in exploring this, but it seems to me a drop out of scope at the moment.

Totally agree!

heyMP (approved):
LGTM

Issue #2057, opened 2022-07-12
[1:1] pfe-label


zeroedin:
# Component name: `<pfe-label>`

## Design specification updates

[PatternFly v4 Label](https://www.patternfly.org/v4/components/label)

PR #2058, opened 2022-07-13
feat(label)!: PF1:1

zeroedin:
## What I did
- Removed all SCSS syntax from label stylesheet, was very little in 2.0 implementation
- Implemented `--pf-*` vars per [PFv4 docs](https://www.patternfly.org/v4/components/label/)
- Added `variant` attribute for `filled` `outline` style variants
- Added `is-compact` boolean attribute for compact style version
- Removed `@pfelement` decorator from pfe-label

## Choices 
- Used pf style layout including the border styles on ::before pseudo element, we opted not to do this in 2.0, but decided to revert that for the 1:1 implementation.
- `is-truncated` *not* implemented (for now) 
- Link events *not* implemented (for now)
- Closable icon *not* implemented (for now)

Fixes: #2057


elements/pfe-label/pfe-label.ts
```diff
@@ -1,95 +1,102 @@
...
-   * @default grey
-   */
-  @property({ reflect: true }) color?: 'grey' | 'blue' | 'green' | 'orange' | 'red' | 'purple' | 'cyan' = 'grey';
+  static readonly styles = [...BaseLabel.styles, styles];
 
-  /**
-  * Sets the value for a icon
-  */
-  @property({ reflect: true }) icon?: string | null = null;
+  @property({ reflect: true, type: Boolean, attribute: 'is-compact' }) isCompact = false;
```
bennypowers, line 93:
```suggestion
  @property({ reflect: true, type: Boolean }) compact = false;
```

[guide](https://github.com/RedHat-UX/red-hat-design-system/wiki/Custom-Elements-API-Style-Guide#host-attributes). True, PF react uses `isFoo`, but I think we're justified here in diverging slightly for the sake of clarity, and this is how I approached it for button.

bennypowers:
@zeroedin please add changeset and demo

brianferry:
Hey Steven, 

started taking a look at this PR and it looks great, just a few things stood out to me...

Looking at the style differences between the demo (locally) and https://www.patternfly.org/v4/components/label/ I noticed a few things:

- Font color for Cyan and Orange colors for seem off for the filled labels
- In pfe4 it appears that all of the outline variant border colors are gray whereas in our demo they match the color of the label
- Purple label w/ icon, the icon appears darker? 

For the compact icons can we create demo subcategories for those as well, outline / filled / icons like we do with the full variants? 

Great job on this PR man, very compact and code is easy to read and understand 😀

bennypowers:
still needs a changeset :D

zeroedin:
> still needs a changeset :D

Doh, I'll get that added in the morning;)

.changeset/green-berries-collect.md
```diff
@@ -0,0 +1,17 @@
+---
+"@patternfly/pfe-label": major
+---
+
+Changes to `<pfe-label>` to align it with [PatternFly v4](https://patternfly.org/components/label).
+
+NEW:
+Added `variant` attribute for `filled` and `outline` style variants
+Added `is-compact` boolean attribute for compact style version
```
bennypowers, line 9:
```suggestion
Added `compact` boolean attribute for compact style version
```

bennypowers:
a few last minute changes:
- updated the changeset
- `ifDefined` lit directive is _only_ for binding an attribute value if it's defined, otherwise removing the attribute. When you want to conditionally render something, use ternary
- added `removable` and `truncated` booleans
- added delegatesFocus to shadow root init for pfelabel, which can have a close button
- widened the return type of our render overrides. lit can compose almost anything into a template (at least, lit declares its types such), so best to not be too strict here, since we might want to render the empty string

bennypowers (approved):
Labels give them MANIA

PR #2066, opened 2022-07-24
fix(tools): fixes for docs pages

bennypowers:
## What I did
- package the njk templates with pfe-tools
- add a backreference to the module for every declaration when generating docs

Blocks https://github.com/RedHat-UX/red-hat-design-system/pull/435


bennypowers:
@zeroedin your fix lgtm. if you like it if you like it put a green on it

PR #2072, opened 2022-07-27
fix(tools): load pfv4 fonts in demo 

bennypowers:
## What I did
- nice up the dev server demo, with some internal refactors
- use PFv4 fonts in dev server
- remove the site.tagPrefix option from same, move it to the root of the options objs
- add a hamburger icon to pfe-icon (don't learn from the implementation here, we'll eventually replace it with something nicer)
- fix/improve some component demos

## Notes to reviewers
the dev server demos and the netlify dp demos should be visually equivalent, except for the main header


zeroedin:
Not sure if this is really a blocking issue.  But when you click on a component demo from the nav, a new page loads, and focus returns to the top of the nav on the first nav item, not the top of the page itself.  I feel like this is most likely an accessibility error.   After page load I would assume focus is on the document, usually followed by a skip-nav link which could skip the header links (logo/github) and then place focus on the main navigation.  This gives the user the option to click a header link after page load.  There is also a visual style applied with `:focus-visible` which almost makes it look like an active page indicator.

bennypowers:
@zeroedin I spun that out to a new issue, let's merge this to unblock the next two prs

zeroedin (approved):
Stamp of Approval

PR #2080, opened 2022-07-31
fix(jump-links): shouldUpdateLightDOM

### LINKED ISSUE #2079: typescript error in jumplinks
https://github.com/patternfly/patternfly-elements/runs/7568139954?check_suite_focus=true#step:6:1010

#2073 revealed a typescript error in https://github.com/patternfly/patternfly-elements/blob/cd12cef723427a87e46524614582195a001ad00f/elements/pfe-jump-links/pfe-jump-links-nav.ts#L1029-L1033

> error TS2774: This condition will always return true since this function is always defined. Did you mean to call it instead?

bennypowers:
## What I did
- fix a predicate that updates light dom in jump links
- thereby fix #2079

### DIFF
--- a/.changeset/warm-fishes-listen.md
+++ b/.changeset/warm-fishes-listen.md
@@ -1,0 +1,5 @@
+---
+"@patternfly/pfe-jump-links": patch
+---
+
+Fix predicate which updates light DOM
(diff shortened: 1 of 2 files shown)


bennypowers:
Sunday, and this one-liner is blocking all releases in this repo, so I needed to merge w/o review

PR #2081, opened 2022-07-31
fix(tools): template file in tarball

bennypowers:
## What I did
Include the demo template file in the npm tarball


bennypowers:
Like #2080, this small fix has to go out today, so i'll merge w/o review

PR #2101, opened 2022-08-19
docs(datetime): fixing href & script tags in datetime docs

brianferry:
Link to the stylesheet is broken for datetime. Script tag brings in < tag when spaces are put between lines so I went ahead and removed those. 

https://deploy-preview-2101--patternfly-elements.netlify.app/components/datetime/


bennypowers (approved):

bennypowers:
Nice catch, @brianferry

PR #2107, opened 2022-08-23
chore(autocomplete): add missing dependency

### LINKED ISSUE #2102: [bug] `pfe-autocomplete` depends on `pfe-button`, but not in package.json
@bennypowers `pfe-autocomplete` also has a dependency on `pfe-button`, we should add it to it's package.json as well.  Otherwise this looks good to go.

_Originally posted by @zeroedin in https://github.com/patternfly/patternfly-elements/issues/2097#issuecomment-1222549017_

bennypowers:
## What I did
Add pfe-button as a dependency to pfe-autocomplete

Closes #2102

### DIFF
--- a/.changeset/cool-coats-press.md
+++ b/.changeset/cool-coats-press.md
@@ -1,0 +1,8 @@
+---
+"@patternfly/pfe-button": patch
+"@patternfly/pfe-icon": patch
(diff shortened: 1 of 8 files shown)


heyMP (approved):
LGTM 🤘

PR #2136, opened 2022-09-01
fix(tools): deslugify demos paths

bennypowers:
## What I did
- add a `deslugify` helper to the config module in tools. This function takes a slug like `call-to-action` and returns a tag name like `rh-cta`. It caches the config file in memory for performace, but if config changes, any long-running tasks (e.g. dev server) will need a restart
- fixes the Manifest demo metadata getter so it deslugifies demo urls.


zeroedin (approved):
Very nice, dig how it reuses the .pfe.config.json so can be used in the inheriting repos

PR #2138, opened 2022-09-05
feat(icon)!: icon 1:1

bennypowers:
Let it be known ahead of time: there will be rebasings

## Demo

:zap: :mage: [Behold!](https://deploy-preview-2138--patternfly-elements.netlify.app/components/icon/demo/) :eye: :crystal_ball: 

### Dependent Components
- [pfe-accordion](https://deploy-preview-2138--patternfly-elements.netlify.app/components/accordion/demo/) (don't mind changing up the icon here, but I'd rather avoid bikeshedding b/c anyways we're 1:1-ing the accordion element)
- [pfe-button](https://deploy-preview-2138--patternfly-elements.netlify.app/components/button/demo/)
- [pfe-clipboard](https://deploy-preview-2138--patternfly-elements.netlify.app/components/clipboard/demo/)
- [pfe-label](https://deploy-preview-2138--patternfly-elements.netlify.app/components/label/demo/)
- [pfe-modal](https://deploy-preview-2138--patternfly-elements.netlify.app/components/modal/demo/)
- [pfe-select](https://deploy-preview-2138--patternfly-elements.netlify.app/components/select/demo/)

### Accessibility Considerations
The shadow container for icon will be removed from the a11y tree if the `label` attribute is unset.

## Ready-for-merge Checklist

- [x] Changesets added.
- [x] Documentation (README.md, docs pages) updated or added.
- [x] Dependent elements updated
- [x] commitlint errors can be safely ignored this time


heyMP:
@bennypowers, @zeroedin and I are seeing this error on node -v 16.

```
   npm run start

> start
> npm-run-all build:tools build:icons build:analyze --parallel dev:* watch:docs:analyze


> build:tools
> npm run build --workspace @patternfly/pfe-tools


> @patternfly/pfe-tools@1.0.0-next.29 build
> tsc -b .


> build:icons
> ts-node --esm --project tsconfig.icons.json ./scripts/icons/build.ts

SyntaxError: Unexpected identifier
    at Loader.moduleStrategy (node:internal/modules/esm/translators:146:18)
ERROR: "build:icons" exited with 1.
```

We tried removing package-lock.json, removing node_modules, pinning esbuild to `0.15.5`.  The only solution that worked for me was changing to node -v 18.

bennypowers:
hmm

```
$ node -v
v16.15.1
$ npm run build:icons

> build:icons
> ts-node --esm --project tsconfig.icons.json ./scripts/icons/build.ts

(node:65545) ExperimentalWarning: Importing JSON modules is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
$
```

bennypowers:
do you have a global tsnode?
if i had to guess, i'd say it's probably something in the toolchain choking on the import assertions in that file

zeroedin:
> global tsnode

I do not have a global tsnode.  The only global packages I have installed in the nvm version are: node and yarn.

bennypowers:
well, I did a couple of things:

```bash
cd patternfly-elements
git stash
git stash
git checkout main
git checkout feat/pf-icon-1-1
git pull
npm run clean:nuke
npm ci

# install missing esbuild binaries, and write them to package lock
npm i -w tools/pfe-tools esbuild --include=optional
git add tools/pfe-tools/package.json package-lock.json
git commit -m "chore: include optional esbuild deps"
git push
```

With those done, I was able to run the dev server both on linux and on macos

```bash
npm run clean:nuke
npm ci
npm start
```

scripts/icons/icons.ts
```diff
@@ -0,0 +1,62 @@
+import type { IconPack } from '@fortawesome/free-regular-svg-icons';
+
+import { fas as FAS } from '@fortawesome/free-solid-svg-icons';
+import { far as FAR } from '@fortawesome/free-regular-svg-icons';
+import { fab as FAB } from '@fortawesome/free-brands-svg-icons';
+
+import pfIcons from '@patternfly/patternfly/icons/pf-icons.json' assert { type: 'json' };
```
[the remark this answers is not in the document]
heyMP: Wow @brianferry awesome find!

heyMP:
@bennypowers 

> Try running that last snippet

I ran those steps and unfortunitely it didn't fix the issue.  Running on `node v16.11.0`

Is it possible you are on a version greater than version 16?

bennypowers:
16.15.1

heyMP:
Wow, updating to 16.15.1 fixed it for me 😂

bennypowers:
thanks for the  help debugging. bumped the nvmrc

scripts/icons/icons.ts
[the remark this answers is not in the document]
bennypowers: this should be resolved with the new `.nvmrc`. make sure to `nvm use`

brianferry:
@bennypowers - Seeing an issue on the docs page on the demo site where the icons don't appear to be showing on the main docs page but they do show on the /demo page.  Thinking it might be the relative paths of the icons?

bennypowers:
alright give that another look

bennypowers:
@heyMP @zeroedin @brianferry thoughts on the non-bubbling `error` event?

alternatives:
- bubble it but require users to handle it before it throws when it reaches the window 
- rename it `icon-error` `load-error` `fail` etc
- keep it like this, but make users who need to catch them for analytics fire their own events

elements/pfe-icon/pfe-icon.ts
```diff
@@ -1,224 +1,157 @@
...
-  /**
-   * Whether to draw a circular background behind the icon.
-   */
-  @property({ type: Boolean, reflect: true }) circled = false;
+  /** Size of the icon */
+  @property({ reflect: true }) size: 'sm'|'md'|'lg'|'xl' = 'sm';
 
-  @property({ type: Boolean, reflect: true }) block = false;
+  /** Accessible label for the icon */
+  @property({ reflect: true }) label?: string;
```
bennypowers, line 83: remove, always aria-hidden=true the shadow icon, document to use `aria-label(ledby)`

cc Greg Gibson

elements/pfe-icon/docs/pfe-icon.md
```diff
@@ -6,198 +6,186 @@
 </style>
 
 {% renderOverview %}
```
bennypowers, line 8: add accessibility notes (aria-label(ledby))

elements/pfe-icon/pfe-icon.ts
bennypowers: cc @hellogreg

elements/pfe-icon/docs/pfe-icon.md
bennypowers: cc @hellogreg

elements/pfe-band/test/pfe-band.spec.ts
```diff
@@ -4,6 +4,8 @@ import { hexToRgb, getColor } from '@patternfly/pfe-tools/test/hex-to-rgb.js';
 import { PfeBand } from '@patternfly/pfe-band';
 import { setViewport } from '@web/test-runner-commands';
 
+import '@patternfly/pfe-tools/test/stub-logger.js';
```
brianferry, line 7: Not sure what this is doing?  I don't see a console logger or a logger function inside the pfe-band component.

elements/pfe-tabs/demo/pfe-tabs.html
```diff
@@ -450,7 +450,7 @@ <h2>Content 2</h2>
           <p>Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est
             laborum.</p>
         </pfe-tab-panel>
-        <pfe-tab role="heading" slot="tab" id="tab3">Tab 3</pfe-tab>
+        <pfe-tab role="heading" slot="tab" id="tab3-a">Tab 3</pfe-tab>
```
brianferry, line 453: Was the -a added here for a test before?

scripts/icons/build.ts
```diff
@@ -0,0 +1,18 @@
+import type { IconSpec } from './icons.js';
+
+import { mkdir, writeFile } from 'node:fs/promises';
+import * as SETS from './icons.js';
+
+// const getContent = (name: string, { height, width, path, xOffset = 0, yOffset = 0 }: IconSpec) =>
```
brianferry, line 6: Should this be removed?

elements/pfe-card/test/pfe-card.spec.ts
```diff
@@ -2,6 +2,8 @@ import { expect, html, aTimeout } from '@open-wc/testing';
 import { createFixture } from '@patternfly/pfe-tools/test/create-fixture.js';
 import { hexToRgb, getColor } from '@patternfly/pfe-tools/test/hex-to-rgb.js';
 
+import '@patternfly/pfe-tools/test/stub-logger.js';
```
brianferry, line 5: Same as the comment on pfe-band, not sure if this is doing something I'm not recognizing here but it doesn't look like the component itself logs or uses the logger class

elements/pfe-button/demo/pfe-button.html
```diff
@@ -47,8 +47,8 @@ <h2>Disabled</h2>
   <pfe-button disabled warning><button>Warning disabled</button></pfe-button>
 
   <h3>Link variant</h3>
-  <pfe-button disabled variant="link" icon="fa-plus-circle"><button>Link disabled</button></pfe-button>
-  <pfe-button disabled variant="link" icon="fa-external-link-square" icon-position="right"><button>Link disabled</button></pfe-button>
+  <pfe-button disabled variant="link" icon="plus-circle"><button>Link disabled</button></pfe-button>
+  <pfe-button disabled variant="link" icon="link" icon-position="right"><button>Link disabled</button></pfe-button>
```
brianferry, line 51: Do we think it would be worth giving them the option to also determine how the icon loads since we now have the loading option?  For example, with button you can now do something like 

```html
<pfe-button variant="link" icon="attention-bell" icon-set="patternfly" icon-position="right">
    <button>Bell</button>
  </pfe-button>
```

But we could also support something like 

```html
<pfe-button variant="link" icon="attention-bell" icon-set="patternfly" icon-position="right" icon-loading="eager">
    <button>Bell</button>
  </pfe-button>
```

brianferry:
@bennypowers this is great, love a lot of the upgrades & updates that have been made in the PR.  Especially like how clean / concise the icons are now.  Left a few question / minor notes in here but I think these changes are great

elements/pfe-band/test/pfe-band.spec.ts
bennypowers: What this does is stub out the Logger methods which call console methods. That way, we can still test them, but they don't pollute the test results output

elements/pfe-tabs/demo/pfe-tabs.html
bennypowers: 🤔 can't remember. It's probably because there's a duplicate id on the page

scripts/icons/build.ts
bennypowers: Probably

elements/pfe-button/demo/pfe-button.html
bennypowers: Perhaps it can wait until a user asks for it

elements/pfe-icon/README.md
```diff
@@ -24,77 +24,62 @@ import '@patternfly/pfe-icon';
...
 
-```javascript
-window.PfeConfig = {
-  IconSets: []
-};
+```js
+// Workaround for bundled pfe-icon: make icon imports absolute, instead of relative to the bundle
+import { PfeIcon } from '/pfe.min.js';
+PfeIcon.getIconUrl = (set, icon) =>
+  new URL(`/assets/icons/${set}/${icon}.js`, import.meta.url);
```
mwcz, line 74: There's no chance of race condition here, right?  Importing pfe-icon has a side effect of running `customElements.define("pfe-icon", ...)`, but `PfeIcon.getIconUrl` is guaranteed to run before any  elements start upgrading.  Pretty sure that's the case.

mwcz: I sure:tm: it's good.  `PfeIcon.getIconUrl = ...` will be on the current stack, but any `PfeIcon` constructors run by element upgrades will happen on new stacks further down the queue.  Only if the decorator could somehow cause an element to upgrade immediately would there be a problem and I don't think there's a way to do that.

elements/pfe-icon/README.md
```diff
@@ -24,77 +24,62 @@ import '@patternfly/pfe-icon';
...
 
-#### Updating an existing icon set
-
-To updating an existing icon set, you use the same `addIconSet` function.  The first input which is the icon set namespace is required, as is the new path.  You can optionally pass in a new function for parsing the icon names into filenames.
+## Loading
 
-```javascript
-PfeIcon.addIconSet("local", "https://hosted-icons.com/");
-```
+Icons load _lazily_ by default, meaning that the browser will not attempt to fetch and render the
```
mwcz, line 80: Lazy load by default, nice improvement!

mwcz:
I finally took a look at this today.  Thanks @bennypowers for the reminder.

It looks like the two main high level changes are the SVG loading mechanism (raw SVG -> ESM), and the SVG injection mechanism (SVG `<image>` tag -> embed in shadow DOM).  It all looks good to me so consider this a :+1:.  Here's some old thoughts from the dawn of pfe-icon, in case they're interesting or useful.

## Loading mechanism

Old pfe-icon loads raw SVGs in order to have a very low barrier when adding new or custom icons.  I use that pretty frequently, so needing to embed the SVGs each in an ES module adds some friction.  I spoke with @bennypowers about this and the idea of splitting the converter out into its own package seemed like a good one (TBD after this PR).  I'm picturing a `svg2esm` package that could be run easily on the command line or imported as a library.

## Injection mechanism

The ability to render the icon as an "in-tree" SVG so to speak, rather than in an SVG `<image>` element, certainly makes coloring the icon easier.  

Though, loading them in an `<image>` has some benefits and tradeoffs as well:

1. `<image>` protects the page from scripts embedded in SVGs
2. `<image>` prevents pfe-icon's shadow DOM CSS from applying to the SVG's inner elements, and prevents CSS embedded in the SVG to leak out into the shadow DOM (pro and con! as this also makes it harder to style/colorize the icon, hence why the old method used filter+feFlood)
3. `<image>` is handled by the browser similarly to HTML's `` in that the same-origin policy is relaxed.  Including an SVG with `fetch` or [`<use>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use) (or importing an ES module) requires CORS headers for cross-origin imports.  Not hard to add, but a frequent stumbling block.

bennypowers:
interesting point about cors. I think shadow root provides the right level of encapsulation here. if svg has scripts or styles that interfere with the limited dom in the shadow root, that's a smell. you provide unminified, weird svg - that's on you.

as for `<script>`, i'm [pretty sure](https://lit.dev/playground/#gist=c21e109ed9a9fe9ca80073e8e3ed641b) that won't run anyways

what does the cli for the hypothetical converter look like?

```sh
svg2esm ./icons/**/*.svg -o ./build/icons/
```

brianferry:
Looks at the code everything looks good, was able to build & serve locally.  I'll do a follow up for the pfe-accordion icon just so we have it in there.

brianferry (approved):

PR #2140, opened 2022-09-07
fix(tools): update dev server ts target

bennypowers:
## What I did
- update the typescript target in dev server config. this allows top-level await and other features


kylebuch8 (approved):
Longtime Gone. Thankful Me.

bennypowers:
lols "one-liner"

PR #2141, opened 2022-09-07
perf: lazy tooltip

bennypowers:
## What I did
I identified a performance issue with tooltip, which blocks the main thread for about 200ms for each tooltip added to the dom. This PR defers initializing popperjs until the tooltip is first show. Rendering 3000 tooltip elements on a page went from 40s-180s to ~5s.

This should be considered a hotfix patch until we can replace popper with something more modern: floating-ui or our own homebrew.

## Testing Instructions
- codepen demo https://codepen.io/bennyp/pen/XWqbEgo?editors=1010
- revert all commits except the demo commit and run the demo
- unrevert commits and compare performance

### DIFF
--- a/.changeset/weak-dragons-lay.md
+++ b/.changeset/weak-dragons-lay.md
@@ -1,0 +1,6 @@
+---
+"@patternfly/pfe-core": minor
+"@patternfly/pfe-tooltip": minor
+---
+
+Improves performance of floating DOM (tooltip) by lazily initializing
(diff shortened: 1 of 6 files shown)


bennypowers:
depends on #2140

brianferry (approved):
Looks good to me, definitely noticing the improvements locally.  Took around 10s to run pre commits and afterwards less than a second.

brianferry:
Only thing I noticed was the performance demo on the preview doesn't look like it can find the module for `lit`, is this an issue with sub-categories of demos right now @bennypowers?

bennypowers:
yeah something fishy is going on: view-source:https://deploy-preview-2141--patternfly-elements.netlify.app/components/tooltip/demo/performance

I'm not satisfied with the demo stuff. Might be best to require demos to be `demo-slug/index.html` etc

PR #2170, opened 2022-10-11
feat(switch): implement face

bennypowers:
## What I did
- implement FACE in pfe-switch
- remove `reversed`, use `:has` instead
- remove label attrs, use `<label>` instead
- rename has-check-icon to show-check-icon
- fallback to pfe tokens

## Not yet done
- [ ] more FACE stuff like value, validations, etc
- [x] fix polyfill and modules shims loading order for demo
- [ ] tests

## Limitations
element-internals-polyfill will probably require some workarounds for fieldsets (alternatively, just wait for safari to implement)
:has requires workarounds for non-supporting browsers. What we could do is imperatively toggle classnames or data attrs or something.


bennypowers:
@WickyNilliams helpfully pointed out that since the shadow inputs are aria-hidden, screen-reader cursor can't navigate to it. most likely the best solution will  be to remove the shadow controls altogether and imperatively wire up the listeners

bennypowers:
:thinking: we could add and optional slot for `label` which would style it automatically and work with `reversed`

bennypowers:
ok i think this is ready to review

Issue #2117, opened 2022-08-25
[1:1]: pfe-clipboard => pfe-clipboard-copy


bennypowers:
# Component name: `<pfe-clipboard>`

To be renamed `<pfe-clipboard-copy>`.

## Design specification updates

[PatternFly v4 Clipboard Copy](https://www.patternfly.org/v4/components/clipboard-copy)

Note: the RHDS clipboard element will likely be very different so consider well what to put in the base class.

PR #2173, opened 2022-10-12
feat(clipboard-copy): add clipboard-copy

heyMP:
Issue: https://github.com/patternfly/patternfly-elements/issues/2117


elements/pfe-clipboard-copy/pfe-clipboard-copy.scss
```diff
@@ -1,3 +1,272 @@
+:root {
```
bennypowers, line 1: we shouldn't need to pull all this in. better to fall back

elements/pfe-clipboard-copy/pfe-clipboard-copy.ts
```diff
@@ -20,83 +28,61 @@ export class PfeClipboardCopy extends LitElement {
...
+        'pf-m-expanded': this.expanded,
+        'pf-m-inline': this.variant === 'inline-compact',
+        'pf-m-block': this.block
+        })}
+      >
+        ${this.variant === 'inline-compact' ?
+          html`
+            ${this.code ?
+              html`
+                <code class="pf-c-clipboard-copy__text pf-m-code">
```
bennypowers, line 58: we'd shouldn't use BEM in shadow DOM

elements/pfe-clipboard-copy/pfe-clipboard-copy.scss
heyMP: I put that stuff in there just to play around with different workflows for doing 1:1 conversion.  After seeing some other examples I'm starting with the plain BaseClass for functionality and will add the PF styles later.

elements/pfe-clipboard-copy/pfe-clipboard-copy.ts
heyMP: Roger 👍

bennypowers (approved):
Great work, @kylebuch8, @heyMP, and @nikkimk!
 
I simplified the shadow template, removed a bunch of javascript, employed "shadow classes" pattern for managing state, added demos, and refactored test.

Well done, all. Looping Generates Tremendous Magnetism

PR #2177, opened 2022-10-14
chore: fix netlify build by adding plugin-local-install-core

kylebuch8:
Attempting to fix the Netlify build by following the guidance here:
https://github.com/netlify/build/issues/1640

The issue with the local plugin is that the dependencies are not installed. The additional plugin should fix the issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -138,3 +138,6 @@ core/pfe-core/demo/*
 /test
 browserstack.err
 !docs/demo/web-dev-server.demo.config.js
+
+# Local Netlify folder
+.netlify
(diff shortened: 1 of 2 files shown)


bennypowers (approved):
Nice solution, @kylebuch8

bennypowers:
So close! https://app.netlify.com/sites/patternfly-elements/deploys/634ae308fd11820008302f75

kylebuch8:
I feel like if we just point at the actual npm package rather than using the local version, it should just work.

bennypowers:
But we don't publish this as an npm package

kylebuch8:
It looks like we published it 8 months ago.

https://www.npmjs.com/package/netlify-plugin-github-actions

bennypowers:
Oh lol ok yeah sure

PR #2181, opened 2022-10-17
chore: remove netlify visual regression workflow

bennypowers:
Previous attempts to fix the netlify build all failed. At this point the best thing to do is remove the visual-regression workflow trigger (which we're anyways not really using) so that we can deploy from main.


kylebuch8 (approved):
LGTM

Issue #2182, opened 2022-10-18
pfe-accordion: need to update pfe-accordion to match new pfe-icon spec


brianferry:
## Description of the issue

With the new pfe-icon spec the updated pfe-accordion will need to be updated to include the set along with the icon text string that matches.

### Impacted component(s)

- pfe-accordion
- pfe-accordion-header

bennypowers:
closed by #2183

PR #2183, opened 2022-10-18
fix(accordion): :bug: updating the icon in BaseAccordionHeader

brianferry:
Updating the icon in BaseAccordionHeader as it was built for the previous pfe-icon spec and this will follow the new one with a customizable set attribute. 

### Related issues

- #2182

### What has changed and why

- Changed the default icon string that is used in BaseAccordionHeader along with adding an icon-set attribute to the component for compability with pfe-icon.
- Updated the pfe-jump-links.js because it is throwing an error currently in the console log that I noticed while testing.

### DIFF
--- a/.changeset/odd-lamps-rescue.md
+++ b/.changeset/odd-lamps-rescue.md
@@ -1,0 +1,5 @@
+---
+"@patternfly/pfe-accordion": patch
+---
+
+Updating pfe-accordion to match the new pfe-icon spec and include a customizable icon-set.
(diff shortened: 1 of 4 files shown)


bennypowers (changes requested):
Good stuff. Let's make a few changes to polish up the HTML

.changeset/odd-lamps-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@patternfly/pfe-accordion": patch
+---
+
+Updating pfe-accordion to match the new pfe-icon spec and include a customizable icon-set.
```
bennypowers, line 5:
```suggestion
Uses latest version of `<pfe-icon>` internally.
```

bennypowers (approved):
les gents sont toujours meillures

PR #2185, opened 2022-10-19
fix(tabs): many refactors

bennypowers:
### What I did
- use ElementInternals
- many refactors (see commit messages)
- some perf improvements (i.e. use a single listener on window)
- rename `tab-expand` to `expand`
- move all the pf css to the pfe-*.scss files
- add attrs to customize the aria-label for scroll buttons
- remove some dependencies
- use pfe-icon


bennypowers (commented):
This is a good change, but perhaps even better would be using the web test runner commands package to test the computed a11y tree. There's some code for that in the switch pr, so maybe its best to leave a Todo comment here until that's landed.

zeroedin (approved):
Merged the inflight code from my branch and updated tests, otherwise....

Leverages Good Tools Magnificently

PR #2190, opened 2022-10-20
fix(label): move base styles to pfe

zeroedin:
## What I did

1. Moved base styles to pfe, as these aren't exported causing issue downstream.

## Testing Instructions
1. `npm start` or view deploy preview

## Reviewer notes
CSS didn't change just combined into the pfe-label.scss file.  But just in case look for any possible discrepancies.

### DIFF
--- a/elements/pfe-label/pfe-label.scss
+++ b/elements/pfe-label/pfe-label.scss
@@ -1,3 +1,17 @@
+:host {
+  position: relative;
+  padding-top: var(--pf-c-label--PaddingTop, var(--pf-global--spacer--xs, 0.25rem));
+  padding-left: var(--pf-c-label--PaddingLeft, var(--pf-global--spacer--sm, 0.5rem));
+  padding-bottom: var(--pf-c-label--PaddingBottom, var(--pf-global--spacer--xs, 0.25rem));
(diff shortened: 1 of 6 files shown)


elements/pfe-label/pfe-label.scss
```diff
@@ -1,3 +1,17 @@
+:host {
+  position: relative;
```
bennypowers, line 2: IMO things like this can stay in BaseStyles.scss, just not the pf vars

elements/pfe-label/pfe-label.scss
```diff
@@ -12,6 +26,154 @@
   max-width: var(--pf-c-label__text--MaxWidth, 16ch);
 }
 
+pfe-icon, ::slotted(pfe-icon) {
+  color: currentColor;
+}
+
+:host,
```
bennypowers, line 33: e.g, this whole rule can stay in BaseStyles

bennypowers (changes requested):
Please also add a changeset

elements/pfe-label/pfe-label.scss
bennypowers: see pfe-switch pr for an example of what i mean

bennypowers (approved):

marionnegp (approved):

Issue #2195, opened 2022-10-21
pfe-accordion: gap when using bordered styles between headers


brianferry:
## Description of the issue

When using the bordered variant of the pfe-accordion the panel headers have a 1px gap between the sidebars instead of flowing into one another.  Needs to be addressed via CSS.

### Impacted component(s)

- pfe-accordion

PR #2196, opened 2022-10-21
[pfe-accordion]: Updating stylesheets, package.json exports, and gap 

brianferry:
## <Pull request description>

- Updating the accordion styles to match the format where the variable specific styles go into pfe-accordion.scss instead of BaseAccordion.scss
- Updating exports for package.json
- Address sidebar gap when bordered style is used

### Related issues

- #2195


zeroedin (approved):
Legos Great To Manipulate

PR #2206, opened 2022-10-27
fix(switch): workaround for double-event firing in ff

### LINKED ISSUE #2209: [bug]switch: clicking switch wrapped by a label does not update the switch
Go to https://patternflyelements.org/components/switch/demo/#nested-label

Click the switch and notice that the switch does not update. If you click the label, everything works. But clicking the switch should also work.

It works correctly in Firefox but not in Chrome. It appears that clicking the switch in Firefox causes two clicks on the switch and just one click in Chrome.

bennypowers:
## What I did
I noticed that in Firefox (only), when switch is nested inside a `<label>`, then clicking the switch fires the onClick handler twice with two different events.

This PR checks the event for the existence of `event.explicitOriginalTarget` ([ff only](https://developer.mozilla.org/en-US/docs/Web/API/Event/explicitOriginalTarget)), and if it exists, runs some heuristics in order of increasing performance burden to check if this event is the doppelgänger event
Closes #2209

## Notes to reviewers
This patch contains some mildly weird javascript (assignment expressions returning their RHS operand). This is so that performance-sensitive work only gets done when needed inside the workaround.

## Testing instructions

load the demo page on all three browsers and check that clicking on the label and on the control toggles the state. So that means for each browser checking that both the first instance of switch, the unlabeled instance of switch, and the nested-in-label instance of switch all work as expected.

Note that in safari, clicking the label anywhere doesn't toggle the state, but let's leave that for a subsequent issue.


brianferry:
Looks good to me in FF version 102.4.0esr (64-bit) and Chrome 106.0.5249.119.  Only oddness I'm seeing is in Safari like you mentioned and doesn't seem tied to the work done in this ticket.  👍

brianferry (approved):
Leaves Gather Together Meticulously

bennypowers:
see https://github.com/calebdwilliams/element-internals-polyfill/pull/86 for the safari issue (tldr it appears to be a polyfill bug)

Issue #2219, opened 2022-11-01
pfe-accordion: accessibility voiceover broken on collapse event


brianferry:
## Description of the issue

When collapsing an accordion with voice reader enabled the collapse event does not trigger a response from the voice reader where the expand event does.

### Impacted component(s)

- pfe-accordion

PR #2220, opened 2022-11-01
fix(accordion): :wheelchair: Fixing pfe-accordion voice accessibility bug on collapse

brianferry:
### Related issues

- #2219


hellogreg (approved):

elements/pfe-accordion/BaseAccordion.scss
```diff
@@ -0,0 +1,5 @@
+:host {
+    transition-property: box-shadow, border;
```
bennypowers, line 2: fascinating! how'd you arrive at this solution?

brianferry: Deconstructing and reconstructing the pre-pfe1:1 accordion and figuring out what was left that made it work alongside the demos at https://www.w3.org/WAI/ARIA/apg/example-index/accordion/accordion gave me a good amount of working examples to figure out what was causing the voiceover to work

PR #2221, opened 2022-11-03
fix(switch): making sure toggle isn't called when switch is disabled

### LINKED ISSUE #2215: [bug] switch: disabled switches are still clickable in Firefox
## Steps to reproduce
- Go to https://patternflyelements.org/components/switch/demo/#disabled in Firefox
- Click on the disabled switch
- Notice how the switch is still clickable and changing from on to off

kylebuch8:
Fixes issue in Firefox where when `pfe-switch` is disabled, the switch can still be toggled. This PR exits the `#toggle` method if `pfe-switch` is disabled.

Fixes #2215.

### Related issues
- #2215

### Preview
- Go to https://deploy-preview-2221--patternfly-elements.netlify.app/components/switch/demo/#disabled in Firefox
- Try to click on the switch to make it toggle

### DIFF
--- a/elements/pfe-switch/BaseSwitch.ts
+++ b/elements/pfe-switch/BaseSwitch.ts
@@ -84,6 +84,10 @@ export abstract class BaseSwitch extends LitElement {
   }
 
   #toggle() {
+    if (this.disabled) {
+      return;
+    }
+
     this.checked = !this.checked;
     this.#updateLabels();
     this.dispatchEvent(new Event('change', { bubbles: true }));
--- a/.changeset/giant-spies-love.md
+++ b/.changeset/giant-spies-love.md
@@ -1,0 +1,4 @@
+---
+"@patternfly/pfe-switch": patch
+---
+Prevented clicks and other interactions when the switch is disabled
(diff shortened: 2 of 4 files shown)


elements/pfe-switch/BaseSwitch.ts
```diff
@@ -84,6 +84,10 @@ export abstract class BaseSwitch extends LitElement {
   }
 
   #toggle() {
+    if (this.disabled) {
```
bennypowers, line 87: this is woke
but bespoke? put `pointer-events: none` on the disabled host. use

```css
:host([disabled]),
:host(:disabled) {
  pointer-events: none;
}
```

bennypowers (changes requested):

elements/pfe-switch/BaseSwitch.ts
kylebuch8: Are you suggesting that I remove the early return in favor of the CSS change? Or are you suggesting that I add this too?

bennypowers: i think the css would do the trick on it's own, but i don't object to the early return

bennypowers: make sure to put it in base styles since it's not pfe-specific

bennypowers (approved):
Lonely geezers trade magazines (don't call hr on me)

Issue #2046, opened 2022-06-20
[1:1] pfe-card


bennypowers:
# Component name: `<pfe-card>`

## Design specification updates

[PatternFly v4 Card](https://www.patternfly.org/v4/components/card)

We may defer implementation of `expanded` and `selectable` for now.

methomps:
Discussion thread around card: https://github.com/orgs/RedHat-UX/discussions/361

brianferry:
@marionnegp @coreyvickery @markcaron 

For `pfe-card` it looks like for 1:1 our coverage should be

- Basic Card (title, body, footer)
- Cards that can include any combination of the following
  - Title Header
  - Image Header
  - Body
  - Footer
- Compact Card
- Large Card

Do we want to also support for the first pass the following options: 

Borders / Drop Shadows:
- Flat (no drop-shadow, does have a border)
- Rounded (sets a border radius)
- Plain (No border or drop shadow)

Styling of the internals of the card itself:
- Full-Height (Body takes up full height)
- Card w/ dividers between sections (similar to how accordion can include borders between sections)

Actions
**What are they?** pf v4 includes a menu and a checkbox item, presumably could be anything and would have to be provided functionality by the user?
**What do they include?**
- Card w/ image and **actions**
- Card w/ title and **actions**
- Card w/ **actions** and no title

As mentioned by @bennypowers previously we will likely need to defer expanded and selectable for now. 

Keeping in mind too that whatever we add in here could be pfe specific if we think rhds is not going to need it.

brianferry:
Added a branch @ feat/pfe-card-pfv4 for this work https://github.com/patternfly/patternfly-elements/tree/feat/pfe-card-pfv4

coreyvickery:
@markcaron @marionnegp Seems fine for 1:1, are we missing anything else?

bennypowers:
Closed by #2222

PR #2222, opened 2022-11-03
feat(card): pfe-card 1:1 

brianferry:
## <Pull request description>

Updating the pfe-card component to match the styles and attributes of the pfv4 card component.

### Related issues

- #2046


elements/pfe-card/pfe-card.ts
```diff
@@ -27,123 +15,64 @@ import style from './pfe-card.scss';
  * @slot - Any content that is not designated for the header or footer slot, will go to this slot.
  * @slot footer
  *       Use this slot for anything that you want to be stuck to the base of the card.
- *       This region is bottom-aligned.
- *
- * @slot pfe-card--header - {@deprecated use header}
```
bennypowers, line 32: we need to list the header slot

elements/pfe-card/BaseCard.ts
```diff
@@ -0,0 +1,45 @@
+import { LitElement, html } from 'lit';
+import { property } from 'lit/decorators.js';
+import { SlotController } from '@patternfly/pfe-core/controllers/slot-controller.js';
+
+import { classMap } from 'lit/directives/class-map.js';
+
+import style from './BaseCard.scss';
+
+export abstract class BaseCard extends LitElement {
```
bennypowers, line 9: this should list slots and parts

elements/pfe-card/pfe-card.ts
```diff
@@ -27,123 +15,64 @@ import style from './pfe-card.scss';
...
-   * Sets color theme based on parent context
-   */
-  @colorContextConsumer()
-  @property({ reflect: true }) on?: ColorTheme;
-
-  /** Optionally adjusts the padding on the container. Accepts: `small`. */
-  @property({ reflect: true }) size?: 'small';
+ * Optionally allow the card to take up the full height of the parent element.
+ */
+  @property({ type: Boolean, reflect: true }) fullHeight = false;
```
bennypowers, line 70: the attr for this should be `full-height`

elements/pfe-card/test/pfe-card.spec.ts
```diff
@@ -80,302 +56,142 @@ describe('<pfe-card>', function() {
...
-    const element = await createFixture<PfeCard>(dynamicHeaderFooter);
-
-    const header = document.createElement('h2');
-    header.setAttribute('slot', 'pfe-card--header');
-    header.textContent = 'Card Header';
+  describe('size', () => {
+    describe('is unset', () => {
+      it('should have default sizing for card parts', async function() {
+        const element = await createFixture<PfeCard>(TEMPLATES.card1);
+        const cardHeader = element.shadowRoot!.querySelector<HTMLSlotElement>('[part="header"]')!;
```
bennypowers, line 63: it's better to avoid testing based on shadow dom structure. can you check the computed size of the whole element or it's slotted content instead?

elements/pfe-card/test/pfe-card.spec.ts
```diff
@@ -80,302 +56,142 @@ describe('<pfe-card>', function() {
...
-  it(`should render a header and footer when content for those deprecated slots are added dynamically`, async function() {
-    const element = await createFixture<PfeCard>(dynamicHeaderFooter);
-
-    const header = document.createElement('h2');
-    header.setAttribute('slot', 'pfe-card--header');
-    header.textContent = 'Card Header';
+  describe('size', () => {
+    describe('is unset', () => {
+      it('should have default sizing for card parts', async function() {
+        const element = await createFixture<PfeCard>(TEMPLATES.card1);
```
bennypowers, line 62: @hellogreg which accessibility-related things should we be testing here?

please see https://modern-web.dev/docs/test-runner/commands/#accessibility-snapshot for an idea of what we can do - basically get a snapshot of the AOM for a given element then assert things on in

bennypowers (approved):
Lookin' good, Brian

A few things to change, we'll have this merged soon

elements/pfe-card/test/pfe-card.spec.ts
bennypowers: sure thing. please schedule a call to go over the a11y snapshot facility at your convenience

elements/pfe-card/BaseCard.ts
brianferry: Great callout, updating now

elements/pfe-card/pfe-card.scss
```diff
@@ -1,278 +1,75 @@
...
+  box-shadow: var(--pf-c-card--BoxShadow, var(--pf-global--BoxShadow--sm, 0 0.0625rem 0.125rem 0 rgba(3, 3, 3, 0.12), 0 0 0.125rem 0 rgba(3, 3, 3, 0.06)));
 }
 
-/* @deprecated */
-:host([color]),
-:host([color-palette]),
-:host([on]) {
-  background-color:    var(--pfe-card--BackgroundColor, var(--pfe-context-background-color));
+:host([size="compact"]) {
+  --_pf-c-card__body--FontSize: var(--pf-c-card--size-compact__body--FontSize, var(--pf-global--FontSize--sm, .875rem));
```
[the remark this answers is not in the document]
brianferry: Yes they're ours, I can simplify them since they are pretty long though I thought we were keeping the variable styles outside of BaseCard?  These are also not relevant for rh-card since they go along with the variants present in pfe-card if that makes sense?

elements/pfe-card/demo/pfe-card.html
```diff
@@ -4,98 +4,39 @@
...
 <link rel="stylesheet" href="/core/pfe-styles/pfe.min.css">
 <link rel="stylesheet" href="/elements/pfe-card/pfe-card--lightdom.min.css">
-<link rel="stylesheet" href="./demo.css">
+<link rel="stylesheet" href="/core/pfe-styles/pfe.min.css">
+<link rel="stylesheet" href="/elements/pfe-card/demo/demo.css" />
 <script type="module" src="./pfe-card.js"></script>
 
-<pfe-card color-palette="lightest" border class="button-series">
+<div>
+  <input id="compact" type="checkbox" class="card_controls" value="compact" />
```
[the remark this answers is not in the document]
brianferry: Good point, switched (lol) those over now.

elements/pfe-card/pfe-card.ts
brianferry: Ahh yeah that makes sense too, updated & updated styles / demos

elements/pfe-card/test/pfe-card.spec.ts
```diff
@@ -195,4 +195,19 @@ describe('<pfe-card>', function() {
...
+      const element = await createFixture<PfeCard>(TEMPLATES.card1);
+
+      expect(element).to.be.accessible();
+      expect(element.shadowRoot).to.be.accessible();
+    });
+
+    it('should have an article element wrapper in the shadow dom', async () => {
+      const element = await createFixture<PfeCard>(TEMPLATES.card1);
+
+      expect(element.shadowRoot?.querySelector('article')).not.to.be.undefined;
```
bennypowers, line 210: This should not query the shadow root. Use the a11y snapshot instead

bennypowers (changes requested):

elements/pfe-card/test/pfe-card.spec.ts
brianferry: So when I was looking at using a11ysnapshot there did not appear to be a way to tell whether or not the element has an article element present.  If the article element is present or it's not I see a role of 'WebArea' so I'm not sure if that would be the best test case here.  

Here's the output I'm seeing

```
{
        role: 'WebArea',
        name: '',
        children: [
          { role: 'heading', name: 'Card 1', level: 2 },
          { role: 'text', name: 'This is pfe-card.' },
          { role: 'text', name: 'Text in footer' }
        ]
      }
```

Though there's also the possibility I'm just not using the plugin correctly lol

bennypowers: https://github.com/modernweb-dev/web/issues/2083
it's not you. 

Do your best until then, but try to avoid querying the shadow root

elements/pfe-card/pfe-card.scss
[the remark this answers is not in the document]
bennypowers: what's important in Base* styles isn't that they don't have variables, but that they don't have patternfly-specific variables, but 'private' props that are related to functionality are ok, as long as they don't reference pfe defaults. if they use pfe defaults, keep them here.

bennypowers (approved):
Legitimately Great Taskwork, Man

PR #2230, opened 2022-11-09
fix(switch): adding repo and file list to package.json

brianferry:
Noticed an issue while working with the new pfe-switch where the published package included the .ts and .scss files.  Added those to ignore from package.json.  

Example on unpkg: https://unpkg.com/browse/@patternfly/pfe-switch@2.0.0-next.0/


bennypowers (approved):
good catch. can't wait to roll these all into a single package!

PR #2233, opened 2022-11-11
feat(code-block): adding pfe-code-block

### LINKED ISSUE #2093: [1:1] pfe-code-block
# Component name: `<pfe-code-block>`

> The code-block component exists in PFv4 and PFE. It will be updated in PFE and migrated to RHDS.

## Design specification

[PatternFly v4 Code Block](https://www.patternfly.org/v4/components/code-block)

kylebuch8:
# Code Block
Adding `pfe-code-block` to PatternFly Elements. It should match https://www.patternfly.org/v4/components/code-block.

Preview: https://deploy-preview-2233--patternfly-elements.netlify.app/components/code-block/

Closes #2093


kylebuch8:
Part of this PR _should_ be to delete the old `pfe-codeblock` but it's being used in some of the documentation pages. Thoughts on how we should proceed? The new `pfe-code-block` doesn't have the same features and syntax highlighting as the original version.

kylebuch8:
Also, the copy to clipboard functionality isn't a 1-to-1. We need pfe-clipboardcopy to be completed before it will be an exact match. I don't think it's a blocker.

bennypowers:
> Part of this PR _should_ be to delete the old `pfe-codeblock` but it's being used in some of the documentation pages. Thoughts on how we should proceed? The new `pfe-code-block` doesn't have the same features and syntax highlighting as the original version.

we should remove every instance from the docs and use markdown fenced code blocks instead. There's no reason to run client-side javascript for something we can do in 11ty

elements/pfe-code-block/BaseCodeBlock.ts
```diff
@@ -0,0 +1,38 @@
+import { LitElement, html, nothing } from 'lit';
+import { property, queryAssignedNodes } from 'lit/decorators.js';
+import styles from './BaseCodeBlock.scss';
+
+export abstract class BaseCodeBlock extends LitElement {
+  static readonly styles = [styles];
+
+  @property({ type: Boolean, reflect: true }) expanded = false;
+  @property({ type: Boolean }) showExpandButton = false;
+  @queryAssignedNodes({ slot: 'expandable-code', flatten: true }) expandableCodeNodes!: Array<Node>;
```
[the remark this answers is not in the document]
bennypowers: alternately, put everything in `PfeCodeBlock`, then we can decide later on what to do without making a breaking change.

elements/pfe-code-block/demo/pfe-code-block.html
```diff
@@ -0,0 +1,39 @@
+<link rel="stylesheet" href="./demo.css"/>
+<script type="module" src="./pfe-code-block.js"></script>
+
+<article>
+  <h1>Code Block</h1>
+  <h2>Basic</h2>
+  <pfe-code-block id="code">
+    <span slot="code">apiVersion: helm.openshift.io/v1beta1/
```
bennypowers, line 8: can we assume js is always on? if not, why not ape zero-md and use `<script type="{{mimetype}}">`?

elements/pfe-code-block/package.json
```diff
@@ -36,39 +39,12 @@
...
-  ],
-  "repository": {
-    "type": "git",
-    "url": "github:patternfly/patternfly-elements",
-    "directory": "elements/pfe-codeblock"
-  },
-  "bugs": {
-    "url": "https://github.com/patternfly/patternfly-elements/issues"
-  },
-  "files": [
```
bennypowers, line 52: we want to include ts sources and demos, etc?

bennypowers (changes requested):

bennypowers (approved):

PR #2239, opened 2022-11-17
fix(tabs): correct typo and made imports uniform

zeroedin:
## What I did

 - Fixed `<pfe-tabs>` CSS imports and a typo that caused an issue with the child repository having incorrect styles


heyMP (approved):
Good change 👍

PR #2256, opened 2022-12-08
fix(tabs): export inset type and add inset as property

zeroedin:
## What I did

- Added inset as a decorated property 
- Exports InsetVariant type

### DIFF
--- a/elements/pfe-tabs/BaseTabs.ts
+++ b/elements/pfe-tabs/BaseTabs.ts
@@ -164,14 +164,7 @@ export abstract class BaseTabs extends LitElement {
   }
 
   override render() {
-    const classes = {
-      'inset-sm': this.inset === 'sm',
-      'inset-md': this.inset === 'md',
-      'inset-lg': this.inset === 'lg',
-      'inset-xl': this.inset === 'xl',
-      'inset-2xl': this.inset === '2xl',
-      'inset-page': this.inset === 'page'
-    };
+    const classes = { [`inset-${this.inset}`]: true };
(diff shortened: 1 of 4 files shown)


elements/pfe-tabs/BaseTabs.ts
```diff
@@ -74,6 +83,8 @@ export abstract class BaseTabs extends LitElement {
 
   id: string = this.id || getRandomId(this.localName);
 
+  @property({ reflect: true }) inset?: InsetVariant;
```
heyMP, line 86: Do we need this property to be reflected to the attribute?

zeroedin: After discussion in chat, you are correct.

elements/pfe-tabs/BaseTabs.ts
```diff
@@ -152,9 +164,17 @@ export abstract class BaseTabs extends LitElement {
...
 
   override render() {
+    const classes = {
+      'inset-sm': this.inset === 'sm',
+      'inset-md': this.inset === 'md',
+      'inset-lg': this.inset === 'lg',
+      'inset-xl': this.inset === 'xl',
+      'inset-2xl': this.inset === '2xl',
+      'inset-page': this.inset === 'page'
+    };
```
heyMP, line 174: We could shorthand this to:

```suggestion
    const classes = { [`inset-${this.inset}`]: true };
```

heyMP (approved):
Great use of classMap to solve this use case 👍

elements/pfe-tabs/BaseTabs.ts
zeroedin: I really phoned that one in at the end of the day huh? 😆

PR #2257, opened 2022-12-08
feat(tooltip): remove popperjs dependency, replace with floating-ui/dom

### LINKED ISSUE #2258: [bug] tooltip: double tab index when wrapped around focusable element
## Description of the issue

When pfe-tooltip is wrapped around a focusable element like anchor or button, there is an extra focus effect.

### Impacted component(s)

- pfe-tooltip

### Steps to reproduce

1. Go to https://patternflyelements.com/components/tooltip/
2. Tab through the buttons on the demo page.
3. You'll see that there is a focusable item for the tooltip invoker AND one for the button

### Expected behavior

I don't think tooltips should have there own tabindex.  They should only monitor focusin, focusout and similar events.

Reference: 

> A tooltip provides extra information about a form field, a link, a button, or other focusable element. It must be triggered by both focus and hover events and remains on the screen as long as the trigger has the focus. The focus does not move to the tooltip.

[https://dequeuniversity.com/library/aria/tooltip](https://dequeuniversity.com/library/aria/tooltip)

### Screenshots

Bug:

https://user-images.githubusercontent.com/3428964/206749810-cacea7fb-4c37-4b51-a5e8-29c27df6d038.mp4

Expected behavior:

https://user-images.githubusercontent.com/3428964/206749890-c2b6dc85-c753-42fb-85e7-51ba524db913.mp4

### Proposal

```patch
diff --git a/elements/pfe-tooltip/BaseTooltip.ts b/elements/pfe-tooltip/BaseTooltip.ts
index 8f5615919..24054fa27 100644
--- a/elements/pfe-tooltip/BaseTooltip.ts
+++ b/elements/pfe-tooltip/BaseTooltip.ts
@@ -63,7 +63,8 @@ export abstract class BaseTooltip extends LitElement {
   }

   #addListeners() {
-    this.addEventListener('focus', this.show);
+    this.addEventListener('focusin', this.show);
+    this.addEventListener('focusout', this.hide);
     this.addEventListener('blur', this.hide);
     this.addEventListener('tap', this.show);
     this.addEventListener('click', this.show);
@@ -74,7 +75,7 @@ export abstract class BaseTooltip extends LitElement {
   override render() {
     const { initialized } = this.#domController;
     return html`
-      <div id="invoker" part="invoker" role="tooltip" tabindex="0" aria-labelledby="tooltip">
+      <div id="invoker" part="invoker" role="tooltip" aria-labelledby="tooltip">
         <slot></slot>
       </div>
       <div id="tooltip" aria-hidden=${!this.#isOpen} class=${classMap({ initialized })}>
```

brianferry:
### What has changed and why

Currently tooltip has a dependency on the [popperjs library](https://popper.js.org/) that while it works it has some issues for us, notably
- Performance
  - `@popperjs/core` has some  
- The `process.env.NODE_ENV` issue 
  - Currently when `pfe-tooltip` or `rh-tooltip` are being used we need to make sure the `process.env.NODE_ENV` variable is set to `production` so the @popperjs library can function. 

As a result, we want to make the switch to using `@floating-ui/dom` which will provide all the same functionality as `@popperjs/core` but without the `NODE_ENV` issue and with some small performance enhancements.

image

### Performance Tests

| Test        | Current          | New ✨  |
| ------------- |:-------------:| -----:|
| Rendering 10,000 elements      | 0.d5478s | 0.4255s |
| Triggering .show() on 10,000 elements      | 117s (29.434s for 2500, x4)      |   6.744s |
| Triggering .click() back and forth between two tooltips 2,000 times | 1.12s     |    0.37s |

### Testing Instructions

Closes #2258. Ensure this PR addresses accessibility concerns raised there.


bennypowers (approved):
Let's go to Madrid!

PR #2260, opened 2022-12-12
refactor(tabs): abstract base class fields

### LINKED ISSUE #2254: [bug] pfe-tabs missing inset property
Missing inset property.  pfe-tabs.ts should have a `@property` of `inset`.  Styling was added for this functionality so it appears to work visually but no property currently exists.  This might have been an oversight in thinking of moving it from `BaseTab.ts` to `pfe-tabs.ts` while developing.

zeroedin:
## What I did

- Refactor BaseClasses to use abstract fields instead of `@property` decorator
- closes #2254


elements/pfe-tabs/BaseTab.ts
```diff
@@ -23,11 +23,9 @@ export abstract class BaseTab extends LitElement {
   @queryAssignedElements({ slot: 'icon', flatten: true })
   private icons!: Array<HTMLElement>;
 
-  @observed
-  @property({ reflect: true, type: Boolean }) active = false;
+  abstract active: boolean;
```
bennypowers, line 26: you could move the jsdoc blocks to base, and they'll inherit to the subclasses

elements/pfe-tabs/BaseTab.ts
```diff
@@ -23,11 +23,9 @@ export abstract class BaseTab extends LitElement {
...
 
-  @observed
-  @property({ reflect: true, type: Boolean }) disabled = false;
+  abstract disabled: boolean;
 
   #internals = this.attachInternals();
```
bennypowers, line 30: out of scope for this PR, but it could be fun to move this to internals controller or a subclass thereof, then apply those to the subclasses

bennypowers (commented):
looks good. excited to see these changes coming on

bennypowers (approved):
LARPers getting their makeup

PR #2265, opened 2022-12-12
refactor(label): abstract base class fields

zeroedin:
### What I did

- Abstracted base class property fields
- Removed reflected properties in `pfe-tabs.ts` opting for classMap() usage

### DIFF
--- a/elements/pfe-label/pfe-label.ts
+++ b/elements/pfe-label/pfe-label.ts
@@ -9,6 +9,22 @@ import '@patternfly/pfe-button';
 
 import styles from './pfe-label.scss';
 
+export type LabelVariant = (
+  | 'filled'
+  | 'outline'
+);
+
+export type LabelColor = (
+  | 'blue'
+  | 'cyan'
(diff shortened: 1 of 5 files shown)


elements/pfe-label/pfe-label.ts
```diff
@@ -118,6 +149,12 @@ export class PfeLabel extends BaseLabel {
     `;
   }
 
+  willUpdate() {
+    super.willUpdate();
+    const pfeClasses = { compact: this.compact, truncated: this.truncated };
+    this.classes = { ...this.classes, ...pfeClasses };
+  }
+
```
zeroedin, line 157: @bennypowers curious to hear your opinion on this approach of adding classes to the class map from the child class.

bennypowers: if it was up to me i'd rather just do it in `render()`, but this is fine

bennypowers (approved):
LGTM but tests are failing

PR #2266, opened 2022-12-14
refactor(accordion)!: perf and API improvements

bennypowers:
## What I did
- Move stuff from BaseAccordion to PfeAccordion
- Update and reorganize
- Remove code from `*Accordion*`
- refactor same


bennypowers:
@marionnegp there are margins on the panel contents in the demo in this pr, that's because in keeping with semantic html best practises I wrapped the demo panel contents in `<p>`. **so**, even though this isn't a pixel-perfect match, I think there's room for the pr to pass design review here.

image

elements/pfe-accordion/BaseAccordionHeader.ts
```diff
@@ -36,140 +34,102 @@ export abstract class BaseAccordionHeader extends LitElement {
...
-    super();
-    this.addEventListener('click', this._clickHandler);
-  }
+  #logger = new Logger(this);
 
   override connectedCallback() {
     super.connectedCallback();
+    this.addEventListener('click', this.#onClick);
     this.hidden = true;
     this.id ||= getRandomId('pfe-accordion');
```
brianferry, line 51: Probably want to update this to `getRandomId('accordion')` like `BaseAccordionPanel.ts`

bennypowers: good call. will send a fix

bennypowers: used the `localName`

elements/pfe-accordion/demo/pfe-accordion.html
```diff
@@ -1,221 +1,188 @@
...
+
   <pfe-accordion id="bordered-example" bordered>
     <pfe-accordion-header>
       <h3>Item one</h3>
     </pfe-accordion-header>
     <pfe-accordion-panel>
-      Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore
-      magna
-      aliqua.
+      <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</p>
```
brianferry, line 139: Noticing something here with the `large` attribute turned on, the `::after` border is not going up the whole way on the element.  I can see two ways currently that fixes it, one is with removing the `<p></p>` tags around the content inside the panel and the other is removing the `--pf-c-accordion__panel-body--PaddingTop: 0` inside `pfe-accordion-panel.scss`.

elements/pfe-accordion/pfe-accordion-panel.ts
```diff
@@ -1,4 +1,4 @@
-import { customElement, } from 'lit/decorators.js';
+import { customElement, property, } from 'lit/decorators.js';
```
brianferry, line 1: don't need the additional comma here

brianferry:
This looks great @bennypowers, really like the setup for the tests and the layout of the files.  I'll need to do some testing too with rh-accordion to see if I see any changes there but this is awesome.

elements/pfe-accordion/BaseAccordionHeader.ts
```diff
@@ -36,140 +34,102 @@ export abstract class BaseAccordionHeader extends LitElement {
...
-      this._generatedHtag = undefined;
+    if (header !== this.#generatedHtag) {
+      this.#generatedHtag = undefined;
     }
 
     this.headingTag = header?.tagName.toLowerCase() ?? 'h3';
     this.headingText = header?.textContent?.trim() ?? '';
 
+    do {
+      await this.updateComplete;
```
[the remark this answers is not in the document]
bennypowers: What we're doing here is checking to see if `updateComplete` resolves as `false`, which indicates that another update was triggered during the previous update cycle. In which case, we want to wait on the now-enqueued next update. We should do that because we overrode `getUpdateComplete` to include the child pairs' update cycles as well

elements/pfe-accordion/BaseAccordion.ts
```diff
@@ -155,121 +115,66 @@ export abstract class BaseAccordion extends LitElement {
...
+    this.expandedSets.delete(index);
     header.expanded = false;
+    await header.updateComplete;
   }
 
-  private async _collapsePanel(panel: BaseAccordionPanel) {
-    if (!panel) {
-      this.#logger.error(`Trying to collapse a panel that doesn't exist`);
+  async #collapsePanel(panel: BaseAccordionPanel) {
+    if (!panel.expanded) {
```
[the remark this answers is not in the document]
bennypowers: I don't believe so, but try it and see if it helps simplify code

elements/pfe-accordion/BaseAccordion.ts
```diff
@@ -40,113 +33,80 @@ export class AccordionCollapseEvent extends ComposedEvent {
...
-   *   </pfe-accordion-panel>
-   * </pfe-accordion>
-   * ```
-   *
-   * To expand multiple sets, you can dash separate indexes: ?lorem-ipsum=1-2.
-   */
-  @observed
-  @property({ type: Boolean }) history = false;
+  /** When true, only one accordion panel may be expanded at a time */
+  @property({ reflect: true, type: Boolean }) single = false;
```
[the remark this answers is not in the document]
bennypowers: Perhaps it should indeed be abstract. Subclasses might want to do something like `single="page"`

elements/pfe-accordion/demo/pfe-accordion.html
bennypowers: Content should be structured for semantics

We could remove margin block start on first child, block end on last child, but that feels like overreach to me

WDYT @marionnegp ?

bennypowers:
I have a couple changes lined up for accordion, based on this PR

brianferry:
@bennypowers - Moved the single attribute into the child component since rh-accordion won't be needing it at this time (it was defaulting to true), just needed to be able to grab the headers from `pfe-accordion` to do so

elements/pfe-accordion/demo/pfe-accordion.html
bennypowers: I'd like to merge this and rh-accordion asap, so if this isn't a blocker, please drop an approval here

brianferry:
@bennypowers are there still changes you want to make to this branch or are we good to update & merge?

bennypowers:
i'm g2g

brianferry (approved):
Great work on this @bennypowers as always 👏

PR #2269, opened 2022-12-19
fix(create): typescript error

bennypowers:
## What I did

- Placate TypeScript in create cli's main entrypoint
- Update the package-lock

Should unblock #2267


heyMP (approved):
Listen Good Typscript to Monorepos. 👂

PR #2270, opened 2022-12-19
chore: enable array.at in tsconfig

bennypowers:
## What I did

Enable es2022 language features (for `array.at`)

Should unblock #2267


heyMP (approved):
👍

PR #2283, opened 2022-12-26
feat(jump-links)!: reimplement in line with PFv4

bennypowers:
## What I did
- added `@patternfly/pfe-core/controllers/scroll-spy-controller.js`
- reimplemented pfe-jump-links with above, and with styles and APIs in line with PFv4 designs

## Testing Instructions
Run the demo locally, and check the [Deploy Preview](https://deploy-preview-2283--patternfly-elements.netlify.app/components/jump-links/demo/)
- is it accessible? 
- how quickly can you break the scroll spy? under what conditions? give clear repros, please!
- compare the styles and behaviour to [PFv4](https://www.patternfly.org/v4/components/jump-links/) and to the [Scroll spy demo](https://www.patternfly.org/v4/components/jump-links/react-demos/scrollspy-with-subsections/) ([Deploy Preview](https://deploy-preview-2283--patternfly-elements.netlify.app/components/jump-links/demo/scrollspy-with-subsections/))

### Public APIs/HTML
Keep in mind that the primary goal here is to *replicate the pfv4* dx, not to provide a strictly html-first element. Nonetheless, the controller tries not to presume anything about the available DOM, so that it could be useful in implementing a light-dom-heavy version of jump-links in the near future (i.e. rh-jump-links or rh-table-of-contents, whathaveyou).

Try to think ahead to what you'd need to implement an HTML-first / noscript-y version of this, and think if the controller as-is would be enough.

### Assumptions made
There are, as far as i can tell, two places where the controller makes assumptions about it's host's light DOM, both regarding this concept of "link children". Link children are the elements in the controller host's light-dom which link to the scrollable content headings. when the content scrolls up and down the page, the link child which represents the current heading should become "active". 

1. we assume that link children have an `href` attribute which contains a hash reference to its heading (i.e. `<a href="#foo">` or `<pfe-jump-links-item href="#bar">`
2. we assume that it's sufficient to toggle some attribute on the link child when its heading scrolls into view.

If you find these assumptions to be too restrictive, please speak up.


bennypowers:
blocked by #2296

bennypowers:
> For keyboard users, groups of focusable elements should use a roving tabindex pattern so that they can skip the group without tabbing through every item, and use the arrow keys to navigate between items. (See [Navigation Menubar Example | APG | WAI | W3C](https://www.w3.org/WAI/ARIA/apg/example-index/menubar/menubar-navigation))
> 
> FWIW I had to make [`RovingTabindexController`](https://github.com/RedHat-UX/red-hat-design-system/blob/feat/rh-audio-player/lib/RovingTabindexController.ts) pattern because toolbars and menus also need this pattern. If it's useful in this case, perhaps you could take a look, recommend any changes, and decide if we should move should move it to `patternfly-elements`. Or you can riff off of it directly in the jump links.

I think we should port it to here. Can we open a new issue in the 1:1 milestone and merge this so we can move on with the other pre-release chores?

core/pfe-core/controllers/scroll-spy-controller.ts
```diff
@@ -0,0 +1,166 @@
...
+  activeAttribute?: string;
+
+  /**
+   * The root node to query content for
+   * @default the host's root node
+   */
+  rootNode?: Node;
+}
+
+type HREFElement = Element & { href: string };
```
heyMP, line 23: TS question: There is an existing `HTMLAnchorElement` type.  Is there a benefit to going more generic with the `Element` type?

elements/pfe-jump-links/custom-elements-manifest.config.js
```diff
@@ -1,5 +1,5 @@
 import { pfeCustomElementsManifestConfig } from '@patternfly/pfe-tools/custom-elements-manifest.js';
 
 export default pfeCustomElementsManifestConfig({
-  globs: ['pfe-*.ts'],
+  globs: ['pfe-*.ts', 'Base*.ts'],
```
heyMP, line 4: We could add this as a patch note in the changeset.

elements/pfe-jump-links/pfe-jump-links.ts
```diff
@@ -1,17 +1,81 @@
...
 /**
- * Jump links act as persistent navigation that consists of a list of anchor links.
- * Selecting a link moves a user to content that corresponds with the link selected.
- * A link is displayed as active when the content it links to is visible in the browser window.
- *
- * @summary Moves a user to content when a link is selected
- * @cssprop --pfe-jump-links-panel--offset You can control offset in your styling layer as well. This value can be set directly on the component inside a style attribute, e.g. style="--pfe-jump-links-panel--offset: 100;" or using the appropriate selector in another file. Please note that adding an attribute will take precedence over a css value. At the moment only integer values passed to this key are valid. No other values are supported. This means that passing "300px", "2rem","calc(100% - 12px)" will all result in JavaScript errors. You should pass a number that correlates to pixels. To read about the offset attribute, see above.
+ * **Jump links** allow users to navigate to sections within a page.
+ * @fires toggle - when the `expanded` disclosure widget is toggled
+ * @slot - Place pfe-jump-links-items here
```
heyMP, line 14: Do we want to document the PF CSS props here?

heyMP:
@bennypowers I think the API is pretty solid.  I could see a scenario were we could make a more declarative version where we don't rely on the anchor id's but on a wrapper child element.

```html
<pfe-jump-links-section>
  <p>first section...</p>
</pfe-jump-links-section>

<pfe-jump-links-section>
  <p>second section...</p>
</pfe-jump-links-section>
```

As we discussed in our co-working meeting today, I think there's room to make the scrollspy more accurate in determining what section the user is most likely focusing on.  But I agree this could come in a future iteration.

core/pfe-core/controllers/scroll-spy-controller.ts
bennypowers: Yes, we'd generally prefer to use the widest type possible here, since this controller should work with the most number of cases.

In our case specifically, pfe-jump-links-item does not implement the whole HTMLAnchorElement interface, but it does have an href property.

This is one of the assumptions I listed above - that link elements have an href property. One way to avoid this requirement would be to add some `getHash` option, which would default to `el => el.href`

elements/pfe-jump-links/pfe-jump-links.ts
bennypowers: Yes, we should

elements/pfe-jump-links/custom-elements-manifest.config.js
bennypowers: since i was able to put everything in the controller, we don't actually export a base class.

composition > inheritance :muscle:

bennypowers:
Thanks @nikkimk ! This should probably get a changeset for adding roving tabindex controller

elements/pfe-jump-links/pfe-jump-links.ts
```diff
@@ -1,17 +1,143 @@
-import { PfeJumpLinksNav } from './pfe-jump-links-nav.js';
-export { PfeJumpLinksPanel } from './pfe-jump-links-panel.js';
-export { PfeJumpLinksNav };
+import { html, LitElement } from 'lit';
+import { customElement, property, queryAssignedElements } from 'lit/decorators.js';
 
-import { pfelement } from '@patternfly/pfe-core/decorators.js';
-import { customElement } from 'lit/decorators.js';
+import { ScrollSpyController } from '@patternfly/pfe-core/controllers/scroll-spy-controller.js';
+import { RovingTabindexController } from '../../core/pfe-core/controllers/roving-tabindex-controller.js';
```
brianferry, line 5: Should this be updated to pull from `@patternfly/pfe-core/controllers/roving-tabindex-controller.js` instead of `../../core/pfe-core/controllers/roving-tabindex-controller.js`?  & the same for line 6.

brianferry:
For the scrollspy demo locally I am not able to navigate with my keyboard to any of the headings other than 1 & 5 (first and last) using the arrow keys.  Not sure if that is expected or not but didn't see a disabled attribute on those elements.

zeroedin:
> For the scrollspy demo locally I am not able to navigate with my keyboard to any of the headings other than 1 & 5 (first and last) using the arrow keys. Not sure if that is expected or not but didn't see a disabled attribute on those elements.

Not sure what you are seeing @brianferry but I am able to navigate fine with the arrows, home, and end key all working as expected.

elements/pfe-jump-links/pfe-jump-links.ts
zeroedin: Fixed in https://github.com/patternfly/patternfly-elements/pull/2283/commits/fe724763f7e34d451e43232c755387fcb5937c34

.changeset/unlucky-doors-cheer.md
```diff
@@ -0,0 +1,9 @@
+---
+"@patternfly/pfe-core": minor
+---
+
+✨ Added `ScrollSpyController`
```
zeroedin, line 5: Need a changeset entry here for the addition of `roving-tabindex-controller` as well

.changeset/unlucky-doors-cheer.md
```diff
@@ -0,0 +1,9 @@
...
+
+`ScrollSpyController` sets an attribute (`active` by default) on one of it's
+children when that child's `href` attribute is to a hash reference to an IDd
+heading on the page.
```
zeroedin, line 9: Need a changeset entry here for the addition of `roving-tabindex-controller` as well

```Suggestion
✨ Added `ScrollSpyController`
✨ Added `RovingTabindexController`

- `ScrollSpyController` sets an attribute (`active` by default) on one of it's
children when that child's `href` attribute is to a hash reference to an IDd
heading on the page.
- `RovingTabindexController` implements roving tabindex, as described in WAI-ARIA practices.  
```

nikkimk (approved):

PR #2284, opened 2022-12-26
chore: fix ts build

bennypowers:
## What I did
futz with the typescript configs and fix some typescript errors in various test files

should unblock release (see https://github.com/patternfly/patternfly-elements/actions/runs/3780052452/jobs/6425817091)

this was cherry-picked from #2283


tools/pfe-tools/tsconfig.json
```diff
@@ -7,14 +7,8 @@
     "dev-server/bundle.ts"
   ],
   "compilerOptions": {
-    "target": "es2022",
```
heyMP, line 10: We had to bump this to es2022 to support `Array.at`.  Do you think this will cause a regression at all?

bennypowers: Might. Let's put it back

tools/pfe-tools/tsconfig.json
```diff
@@ -7,14 +7,8 @@
     "dev-server/bundle.ts"
   ],
   "compilerOptions": {
```
bennypowers, line 9:
```suggestion
  "compilerOptions": {
    "target": "es2022",
```

bennypowers, line 9:
```suggestion
  "compilerOptions": {
    "target": "es2022",
```

PR #2292, opened 2023-01-02
feat(progress-stepper)!: 1:1

### LINKED ISSUE #2052: [1:1] pfe-progress-stepper

bennypowers:
## What I did
- remove `pfe-progress-steps` and add `pfe-progress-stepper`

closes #2052 
closes #2128
depends on #2296


marionnegp (commented):
Took a quick look today and found a few things.

- [ ] The circles should be 24px (not including border) based on PF's. I'm guessing they're larger in PFE because of it's using rems.
- [ ] Circles in the compact version should be 14px (not including border).
- [ ] Icons in the circles aren't showing up for me.

bennypowers:
> Took a quick look today and found a few things.
> 
>     * [ ]  The circles should be 24px (not including border) based on PF's. I'm guessing they're larger in PFE because of it's using rems.
> 
>     * [ ]  Circles in the compact version should be 14px (not including border).
They use the same rem value as pfv4, and the same token name, so I think these are unrelated
```css
--pf-c-progress-stepper--m-compact__step-icon--Width: 1.125rem;
```

elements/pfe-progress-stepper/pfe-progress-step.ts
```diff
@@ -0,0 +1,86 @@
...
+      <div id="main" class="${classMap({ compact })}">
+        <slot id="title"></slot>
+        <slot id="description" name="description" ?hidden="${!hasDescription}">${this.description}</slot>
+      </div>
+    `;
+  }
+
+  updated(changed: PropertyValues<this>) {
+    super.updated?.(changed);
+    this.#internals.ariaCurrent = String(!!this.current);
```
brianferry, line 78: Looks like we need to wrap this with a checker to make sure we don't get an infinite loop here, it's currently causing the demos not to load for me.

elements/pfe-progress-stepper/pfe-progress-step.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+}
```
zeroedin, line 80: As brian said, there is an issue here causing an infinite loop.

Here was my fix, although I'm not convinced it is ideal either
```Suggestion
  attributeChangedCallback(name:string, _old:string | null, value:string | null) {
    super.attributeChangedCallback(name, _old, value);
    if (name === 'current') {
      this.#internals.ariaCurrent = String(!!this.current);
    }
  }
```

elements/pfe-progress-stepper/pfe-progress-step.ts
```diff
@@ -0,0 +1,86 @@
...
+  danger: { icon: 'circle-exclamation' },
+  warning: { icon: 'triangle-exclamation' },
+  info: { icon: 'resources-full', set: 'patternfly' },
+}));
+
+/**
+ * @slot -
+ *       Short description of the current step.
+ * @slot description
+ *       Longer decscription of the current step.
```
zeroedin, line 24:
```Suggestion
 *       Longer description of the current step.
```

elements/pfe-progress-stepper/pfe-progress-stepper.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  /** Whether to use the compact layout */
+  @observed(function(this: PfeProgressStepper) {
+    this.querySelectorAll('pfe-progress-step').forEach(step => step.requestUpdate());
+  })
+  @property({ type: Boolean, reflect: true }) compact = false;
+
+  #internals = new InternalsController(this, {
+    role: 'progressbar',
+    ariaValueNow: this.value,
```
zeroedin, line 37: ariaValueNow needs a `string` value here.

```Suggestion
    ariaValueNow: this.value.toString(),
```

zeroedin (commented):
Other than the one show stopper, and a couple of other small fixes, this is looking great!

elements/pfe-progress-stepper/pfe-progress-step.ts
bennypowers: I wasn't able to reproduce this infinite loop, neither on the dev server, nor on the 11ty demo page.

Nonetheless, I pushed a commit which gates the assignment. PTAL

elements/pfe-progress-stepper/pfe-progress-step.ts
bennypowers: probably better just to check for changes in `updated`, lit likes to take control of ce lifecycle callbacks.

elements/pfe-progress-stepper/pfe-progress-step.ts
brianferry: Looks good on my side now, thank you!

zeroedin (approved):
Lemongrass Ginger Tofu Marinade

PR #2296, opened 2023-01-03
feat(core): internals controller improvements

bennypowers:
## What I did
- proxy `role` and `aria*` props in `ElementInternals` to `InternalsController`
- wrap the host's `form*Callback` methods in `InternalsController`
- add `formDisabled: boolean` to `InternalsController`
- disable BaseButton if it's disabled either by attr or by form, without requiring the reflected attr when form is disabled.

## Testing Instructions
Check out button's `form controls` demo and make sure everything's kosher there

### DIFF
--- a/core/pfe-core/controllers/internals-controller.ts
+++ b/core/pfe-core/controllers/internals-controller.ts
@@ -3,10 +3,25 @@ import type { ReactiveController, ReactiveControllerHost } from 'lit';
 export class InternalsController implements ReactiveController {
   #internals: ElementInternals;
 
+  disabled = false;
+
   constructor(
     public host: ReactiveControllerHost & HTMLElement,
   ) {
     this.#internals = host.attachInternals();
+    const { formDisabledCallback } = host;
+    Object.defineProperty(Object.getPrototypeOf(host), 'formDisabledCallback', {
+      configurable: true,
+      value: (disabled: boolean) => {
+        formDisabledCallback?.call(host, disabled);
+        this.formDisabledCallback(disabled);
+      }
+    });
+  }
+
+  formDisabledCallback(disabled: boolean) {
+    this.disabled = disabled;
+    this.host.requestUpdate();
   }
 
   hostConnected?(): void
--- a/core/pfe-core/controllers/internals-controller.ts
+++ b/core/pfe-core/controllers/internals-controller.ts
@@ -1,10 +1,6 @@
 import type { ReactiveController, ReactiveControllerHost } from 'lit';
 
 export class InternalsController implements ReactiveController, ARIAMixin {
-  #internals: ElementInternals;
-
-  disabled = false;
-
   declare role: ARIAMixin['role'];
   declare ariaAtomic: ARIAMixin['ariaAtomic'];
   declare ariaAutoComplete: ARIAMixin['ariaAutoComplete'];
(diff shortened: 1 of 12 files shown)


core/pfe-core/controllers/internals-controller.ts
```diff
@@ -1,12 +1,85 @@
...
+
+  static protos = new WeakMap();
+
   constructor(
     public host: ReactiveControllerHost & HTMLElement,
+    options?: Partial<ARIAMixin>
   ) {
     this.#internals = host.attachInternals();
+    // proxy the internals object's aria prototype
+    for (const key of Object.keys(this.#internals)) {
```
heyMP, line 64: I can't get this `Object.keys(this.#internals)` to return anything but an empty array.  Is this working for you?

core/pfe-core/controllers/internals-controller.ts
```diff
@@ -1,12 +1,85 @@
...
+  declare ariaValueMax: ARIAMixin['ariaValueMax'];
+  declare ariaValueMin: ARIAMixin['ariaValueMin'];
+  declare ariaValueNow: ARIAMixin['ariaValueNow'];
+  declare ariaValueText: ARIAMixin['ariaValueText'];
+
   #internals: ElementInternals;
 
+  /** True when the control is disabled via it's containing fieldset element */
+  get formDisabled() {
+    return this.host.matches(':disabled');
```
heyMP, line 53: Will this trigger a re-render for the host element?

core/pfe-core/controllers/internals-controller.ts
```diff
@@ -1,12 +1,85 @@
 import type { ReactiveController, ReactiveControllerHost } from 'lit';
 
-export class InternalsController implements ReactiveController {
+function isARIAMixinProp(key: string): key is keyof ARIAMixin {
+  return key === 'role' || key.startsWith('aria');
+}
+
+export class InternalsController implements ReactiveController, ARIAMixin {
+  declare role: ARIAMixin['role'];
```
heyMP, line 8: TS question.  Is there a reason we hare to `declare` these when we have already indicated that this class implements `ARIAMixin`

core/pfe-core/controllers/internals-controller.ts
bennypowers: No, see above comments. It's a platform limitation, you can't hijack the callback at runtime. Alternatives are a matching class decorator or a mixin.

core/pfe-core/controllers/internals-controller.ts
bennypowers: `implements` just tells the compiler to complain if they aren't there

core/pfe-core/controllers/internals-controller.ts
bennypowers: Maybe we should switch it back to reading the prototype

bennypowers: this should be fixed now

core/pfe-core/controllers/internals-controller.ts
bennypowers: FYI this is what that would look like

```ts
import type { ReactiveElement } from 'lit';
import { InternalsController } from '../controllers/internals-controller.js';

function isReactiveElementClass(
  klass: Function // eslint-disable-line @typescript-eslint/ban-types
): klass is typeof ReactiveElement {
  return typeof klass['addInitializer' as keyof typeof klass] === 'function';
}

const instances = new WeakMap();

/**
 */
export function internals(instance?: ReactiveElement): InternalsController
export function internals(options?: Partial<ARIAMixin>): ClassDecorator
export function internals(optsOrInstance: unknown): InternalsController|ClassDecorator {
  if (optsOrInstance instanceof HTMLElement) {
    return instances.get(optsOrInstance);
  } else {
    return function(klass) {
      if (!isReactiveElementClass(klass)) {
        throw new Error(`@internals may only decorate ReactiveElements. ${klass.name} is does not implement ReactiveElement.`);
      }

      klass.addInitializer(instance => {
        const controller = new InternalsController(instance, optsOrInstance);
        instance.addController(controller);
        instances.set(instance, controller);
      });
    };
  }
}
```

```ts
@internals()
export class BaseButton extends LitElement {
  // ...
  #internals = internals(this);
  // ...
}
```

A mixin is arguably more honest here

heyMP: I see what you're saying.  So end users would be required to use a mixin over a controller.

heyMP (approved):
👍

PR #2297, opened 2023-01-03
docs: band padding & node engine increment in package.json 

brianferry:
1. Changed the engines included in the package.json since we are currently on node v18.12.1
2. Updated the styles for the `.band` class as they weren't showing up with the correct width for me in Firefox or Chrome.  Open to suggestions on this one as I'm not sure if it's necessary to keep the code or to have it only apply for screens >= 768px.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "url": "github:patternfly/patternfly-elements"
   },
   "engines": {
-    "node": ">=14 <=18"
+    "node": ">=16 <=19"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "url": "github:patternfly/patternfly-elements"
   },
   "engines": {
-    "node": ">=14 <=17"
+    "node": ">=14 <=18"
```
bennypowers, line 11:
```suggestion
    "node": ">=16 <=19"
```

bennypowers (changes requested):

PR #2305, opened 2023-01-10
fix(create-element): remove pfelement decorator

bennypowers:
## What I did
- remove `@pfelement` decorator from create-element templates
- remove static version string from same
- update demo css and js

closes https://github.com/RedHat-UX/red-hat-design-system/issues/577

### DIFF
--- a/tools/create-element/templates/element/element.ts
+++ b/tools/create-element/templates/element/element.ts
@@ -1,18 +1,14 @@
 import { LitElement, html } from 'lit';
 import { customElement } from 'lit/decorators.js';
 
-import { pfelement } from '@patternfly/pfe-core/decorators.js';
-
 import styles from '<%= cssRelativePath %>';
 
 /**
  * <%= readmeName %>
  * @slot - Place element content here
  */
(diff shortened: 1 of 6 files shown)


tools/create-element/templates/element/element.ts
```diff
@@ -1,18 +1,14 @@
...
 import styles from '<%= cssRelativePath %>';
 
 /**
  * <%= readmeName %>
  * @slot - Place element content here
  */
-@customElement('<%= tagName %>') @pfelement()
+@customElement('<%= tagName %>')
 export class <%= className %> extends LitElement {
-  static readonly version = '{{version}}';
```
zeroedin, line 14: I know we are moving toward removing individual published versions for the elements,  so this will would push us in that direction 👍

zeroedin (approved):
I think this is an appropriate fix given we are moving toward single-project version publishing.  When I had explored this issue earlier I went down the rabbit hole of making the generation templates customizable in a child repo.  We can maybe explore that as well in a future PR.

Lackluster Greek Takeout, Moussaka

PR #2310, opened 2023-01-13
refactor(tabs): remove imperative modification

zeroedin:
## What I did
- Removed imperative modification of the `tabindex` attribute


brianferry (approved):
Looks good and I learned something new

bennypowers:
Nice solution! You could also use ifDefined in these cases. Good work

PR #2315, opened 2023-01-16
fix(tools): typescript composite

zeroedin:
# What I did

`pfe-tools`: Set typescript compilerOptions `composite: true`

`pfe-tooltip`: Added return type for anonymous function for content in constructor


heyMP (approved):
@bennypowers I'm going to go ahead an approve this PR so it fixes the build errors.  Steven and I were guessing that there might be a better TS config solution.

PR #2316, opened 2023-01-18
chore!: elements

bennypowers:
## What I did
A lot, turns out:
### Infrastructure
- [x] removed percy
- [x] removed last vestiges of sass/scss
- [x] removed lerna, npm-run-all, added wireit
- [x] add 11ty and dev server to `npm start`, generally made running repo scripts a smoother experience
- [x] upgrade to 11ty 2

### Packages
- [x] combined elements into single package
- [x] removed `pfe-sass` and `pfe-styles`
- [x] change all element scss to css

### Docs
- [x] update element screenshots
- [x] add 'landing page' to dev server
- [x] add tabs to homepage with framework snippets

## Stuff that works
- [x] `npm run build`
- [x] `npm start` (here)
- [ ] `npm start` (in rhds)
- [x] `npm run docs`
- [x] ~~`npm run docs --watch`~~ do `npm start`
- [x] generator (here)
- [x] generator (in rhds)

[deploy preview](https://deploy-preview-2316--patternfly-elements.netlify.app/)


nikkimk (commented):
Getting `package.json:80:5 Script "build:create" not found in the scripts section of this package.json. "build:create"` when I do an install.

bennypowers:
Thanks, @nikkimk. Until https://github.com/google/wireit/issues/644, all the wireit targets have to be listed in the `scripts` blocks, so I fixed that up

.gitignore
```diff
@@ -45,20 +47,8 @@ elements/**/*.d.ts
 core/**/*.d.ts
 tools/**/*.d.ts
 
-elements/**/__*.scss
-core/**/__*.scss
-tools/**/__*.scss
-
-elements/**/*.css
-core/**/*.css
 tools/**/*.css
```
heyMP, line 50: I think we should remove this line as well, right?

bennypowers: good catch ty

brianferry:
Potential Individual Areas 

- dotfiles & ci/cd 
- pfe-core
- docs
- Elements/ folder (broken up by different components
- Generator
- Create Element
- Wireit

eyevana:
I'll handle the review of elements/pfe-a*/pfe-b*/pfe-c* :)

brianferry:
I'll take a look at the core/* folder

nikkimk:
I'll take elements/pfe-i* to /pfe-p*

nikkimk (approved):

nikkimk:
I'll take docs too

nikkimk (approved):

brianferry:
It looks like the docs might not be working correctly, on the deploy preview I see the top level "All Components" category but nothing under it.  And in the demo files themselves I'm not seeing the importMap being used.  Is this expected? Wanted to double check that the core/* changes worked with the pfe-button control demo

bennypowers:
@brianferry docs should be better now

elements/pfe-card/demo/pfe-card.html
```diff
@@ -1,12 +1,5 @@
-<link rel="stylesheet" href="/core/pfe-styles/red-hat-font.min.css">
```
eyevana, line 1: @bennypowers removing these stylesheets was intentional? Just want to confirm.

eyevana: It also looks like all the files under `pfe-autocomplete` were removed.

eyevana:
Ok, taking the remaining elements, so pfe-s* & pfe-t*.

elements/pfe-card/demo/pfe-card.html
eyevana: Maybe they're just missing from the `elements/pfe.ts` file?

bennypowers: yes, we're removing `pfe-styles` package. RHDS supercedes it.

as well, pfe-autocomplete, select, and dropdown are not quite ready for 1:1, so we removed them. progress-stepper should still be there

eyevana: Roger that. The rest looks good to me! :)

eyevana: Like my assigned "rest", not the "rest rest" :)

brianferry:
@bennypowers - Noticing a few things I'm trying to work through locally:

- Demos don't seem to be loading correctly (this might be expected since we're changing this to a monorepo so the importmap might not work until it's published?)
- Running into issues locally with the package.json commands that use `ttsc --build . . .` but if I run `npx ttsc --build . . .` it runs locally
- Stemming from the previous problem, having trouble locally running `npm run watch:docs`
- Noticing in the demos/ pages a MIME type error on imported `.css` files
image
  - Also noticed that these files are deleted in the current project so maybe those are not a problem, but I'm curious if css files in general are an issue currently

On the other hand, the demo looks correct with the button in the `npm run start` mode so I think that is correct there.  Love the wireit config and I think that'll end up being awesome to work in.  Really like the more streamlined approach to the folder structure too.

bennypowers:
@brianferry try
```
git clean -dfx
npm ci
```

Then try again

brianferry:
@bennypowers - So it works & builds once, if I run

1. `git clean -dfx`
2. `npm ci`
3. `npm run start`

This works, but if I stop the service and run `npm run start` again I get the following message

image

The `npm run start` command seems to start both the `localhost:8080` and `localhost:8000` service and the 8000 service looks good but the 8080 is where the doc demos don't seem to be pulling in the correct files for me.

bennypowers:
Might be misconfigured wireit cache

bennypowers:
ok @brianferry should be fixed now

brianferry:
@bennypowers - Pushed an update to the branch which should fix the component imports on the demo pages, only thing left on those is that the icons aren't importing (they look like they might need an import map fix or maybe a url fix? The js is there, it just isn't point from `/icons/fab/angle-right.js` -> `/components/icon/icons/fab/angle-right.js`

bennypowers:
@brianferry PTAL at e9dbac1655796e0d9d800bd87b7bac6feb69b97c

brianferry:
@bennypowers - Icons & Demos look great, core looks good, `npm run start` and dependencies are all running as expected.   Create-element looks good to me too 👍. Docs looks good.  Very tiny commit for the demo button styling on the "All Elements" page.

bennypowers:
Thanks, @brianferry !

brianferry (approved):

nikkimk:
Looking at elements/pfe-s* to /pfe-t*

zeroedin (approved):
These changes look and work great, there is a noticeable performance increase in launching the dev server.   Nice Work!

bennypowers:
This will need a squash-and-merge when it's ready

nikkimk (approved):
The rest of elements looks good.

PR #2322, opened 2023-01-27
fix(core): roving tabindex controller arrow up/down keys set focus in…

nikkimk:
## What I did

- switched Arrow Up and Arrow Down in `roving-tabindex-controller`
- made `roving-tabindex-controller`'s updateActiveItem public

## What was changed and why
- In Jump Links, Tabs, and anything else that uses the controller, the arrow up and arrow down keys were moving counterintuitively. (Up was moving to the next item and down to the previous, when it should have been vice versa.)
- In tabs, an active item could already be set when the page loads. Now the active item can be updated with the controller without moving focus to the active tab.

## Testing Instructions
From `jump-links/demo/scrollspy-with-subsections`...

1. Use TAB key to focus on first jump link.
2. Use the ARROW DOWN key to move to the second link.
3. Use ARROW UP to focus on the first jump link.

### DIFF
--- a/.changeset/tasty-ties-try.md
+++ b/.changeset/tasty-ties-try.md
@@ -1,0 +1,6 @@
+---
+"@patternfly/pfe-core": patch
+---
+
+
+Fixed `roving-tabindex-controller.js` so active item can be set without setting focus
--- a/.changeset/tasty-ties-try.md
+++ b/.changeset/tasty-ties-try.md
@@ -1,6 +1,0 @@
----
-"@patternfly/pfe-core": patch
----
-
-
-Fixed `roving-tabindex-controller.js` so active item can be set without setting focus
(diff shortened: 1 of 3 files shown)


zeroedin:
I would assume we'd want a changeset to be added for this change.

zeroedin (approved):
LGTM

.changeset/tasty-ties-try.md
```diff
@@ -0,0 +1,6 @@
+---
+"@patternfly/pfe-core": patch
+---
+
+
+Fixed `roving-tabindex-controller.js` so active item can be set without setting focus
```
heyMP, line 6: Combine these two changesets into one.

zeroedin (approved):

heyMP (approved):
Looks great!

PR #2328, opened 2023-01-30
fix(tabs): remove inset attribute

zeroedin:
## What I did

 - Removed incorrect implementation of inset attribute, opting to use `[part="tabs-container]` to style insets with `--pf-*` css tokens.
 
 
 ### Explanation
 Patternfly Tabs has an `inset` property that takes a map key/value pair of `breakpoint: insetClass`
 
 ```jsx
<Tabs
  inset={{
    default: 'insetNone',
    md: 'insetSm',
    xl: 'insetLg',
    2xl': 'inset2xl'
  }}
>
```

I originally misunderstood how `inset` prop was implemented and implemented it as a single value `inset="sm"`.  

To fix this the cleanest route seemed to be to remove the `inset` property API and just have the `breakpoint: insetClass` key-value pairs exist as standard CSS through using the `::part(tabs-container)` which was already available rather than add more complexity to the API.

```css
@media and screen (min-width: {{breakpoint}}) {
  .inset-sm::part(tabs-container) {
    --pf-c-tabs--inset: var(--pf-global--spacer--sm, 0.5rem);
    --pf-c-tabs--m-vertical--inset: var(--pf-global--spacer--sm, 0.5rem);
    --pf-c-tabs--m-vertical--m-box--inset: var(--pf-global--spacer--sm, 0.5rem);
  }
}
``` 

This does however require the implementor to add these classes to both the component:

```html
<pfe-tabs class="inset-class"></pfe-tabs>
```

and their implementation CSS:

```css
.inset-class::part(tabs-container) {
  /* CSS properties here */
}
@media and screen (min-width: {{breakpoint}}) {
  .inset-class::part(tabs-container){
    /* CSS properties changes at this breakpoint */
  }
}
```

### Alternative Solution

The other solution I believe would be to use an object as the attribute value which I found clunky, but if deemed necessary we can implement

```html
<pfe-tabs inset='{"default": "insetNone", "md": "insetSm", "xl": "insetLg", "2xl": "inset2xl"}'></pfe-tabs>
```

Generates the classes in the shadow dom
```html
...
 <div part="tabs-container" class="pf-m-inset-none pf-m-inset-sm-on-md pf-m-inset-lg-on-xl pf-m-inset-2xl-on-2xl">...</div>
...
```

All combos of CSS classes such as `pf-m-inset-sm-on-md` would then be served in shadow dom CSS.


zeroedin:
Closing due to merge conflicts.  Reimplemented:  #2329

PR #2336, opened 2023-02-02
feat(tools): add pfe cli

bennypowers:
## What I did
- add `pfe` cli
  - [x] `generate|new`: same as create-element
  - [x] `analyze`: generate custom-elements.json
  - [x] create React wrappers with `@lit-labs/react` or some such ([npm](https://npm.im/@lit-labs/react)) 
  - [x] add demos to CEM in analzyer post-processing
  - [x] lint changesets
- move guts of the generator to pfe-tools


bennypowers:
I'm pretty happy with how the analyzer works so far. It could use some TUI polish, and is still missing the demos, but once the next version of lit analyzer releases we'll be in range

PR #2337, opened 2023-02-02
fix(elements): add entrypoints to exports map

bennypowers:
## What I did
- add our element entrypoints to package.json. this blocks release
- simplify the bundle script


scripts/bundle.js
```diff
@@ -1,41 +1,62 @@
...
+import { join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { litCssPlugin } from 'esbuild-plugin-lit-css';
+import Glob from 'glob';
 import CleanCSS from 'clean-css';
 
-import { litCssPlugin } from 'esbuild-plugin-lit-css';
+const glob = promisify(Glob);
+
+const resolveDir = join(fileURLToPath(import.meta.url), '../..//elements');
```
brianferry, line 13: Should this be `'../../elements'` instead of  `'../..//elements'`?

bennypowers: thanks fixed

heyMP (approved):
LGTM 🚀

PR #2343, opened 2023-02-03
fix(tools)!: removes the esbuild plugins and helpers

bennypowers:
## What I did
- Removed esbuild helpers from tools
- moved test runner, dev server and cem/a helpers to their respective folders
- update tools readme
- mark `pfeCustomElementsManifestConfig` as deprecated so we can remove it in a subsequent release without breaking
### Why?
The esbuild helpers had gotten more complex and had built in assumptions which no longer obtained. rather than prescribing build scripts to try and cover every case for every user, it's much better for users to write their own esbuild scripts or rollup configs. furthermore, the esbuild API is in flux, so maintaining these plugins as part of our public API opens us up to maintenance burden and our users to dependency churn.

Our cem/a config relies on a series of patches to an out of date version of cem/a. We should not recommend these in our public api. Better to remove them, and the dependency on cem/a, and either fork cem/a or use the upcoming lit-labs analyzer

It's better to remove these now rather than in a breaking change after RC


tools/pfe-tools/README.md
```diff
@@ -7,57 +7,21 @@ Tools and utilities for building PatternFly Elements and other design systems.
...
-
-Helpers for working with esbuild in design system monorepos
+- Helpers for collating and rendering [custom-elements manifests][cem] in 11ty
+  sites.
+- Various 11ty utility plugins
 
 ## test
 
-Helpers for testing web components using [web test runner](https://modern-web.dev/docs/test-runner/overview/)
+Helpers for testing web components using [web test runner][wtr]
```
zeroedin, line 20: TIL:  silly, but I didn't know you could reference links like this in markdown.

brianferry:
Builds & demos all looking good from here, readme for tools looks good here too

tools/pfe-tools/test/config.ts
```diff
@@ -1,12 +1,11 @@
-import type { PfeDevServerConfigOptions } from './dev-server';
 import type { TestRunnerConfig } from '@web/test-runner';
 
 import { playwrightLauncher } from '@web/test-runner-playwright';
 import { summaryReporter, defaultReporter } from '@web/test-runner';
 import { junitReporter } from '@web/test-runner-junit-reporter';
 import { a11ySnapshotPlugin } from '@web/test-runner-commands/plugins';
 
-import { pfeDevServerConfig } from './dev-server.js';
+import { pfeDevServerConfig, type PfeDevServerConfigOptions } from '../dev-server/config.js';
```
zeroedin, line 8: Another TIL:  I was unaware you could add a type to a regular import line, very useful for keeping things concise.

zeroedin (changes requested):
`npm run new` fails.  Needs a `wireit` command in package.json.  Will submit commit.

```json
"new": "wireit",
...
"new": {
  "command": "./tools/create-element/bin/main.js",
  "dependencies": [
    "build:create"
  ]
},
```

package.json
```diff
@@ -225,6 +225,12 @@
         "build:icons"
       ]
     },
+    "new": {
+      "command": "./tools/create-element/bin/main.js",
+      "dependencies": [
+        "build:create"
```
brianferry, line 231: Looks like `npm run new` is hanging after the `What is the element's name?` message

zeroedin (approved):

package.json
zeroedin: This fix definitely works, we couldn't figure out why the chaining `&&` was working but the `dependencies` in wireit command weren't working in a similar fashion.  Nice fix brian.

bennypowers:
Thanks @brianferry and @zeroedin. Like the fix!

Issue #2324, opened 2023-01-30
pf-avatar missing dark variant image implementation


zeroedin:
Patternfly v4 has a dark variant avatar image:  https://github.com/patternfly/patternfly-elements/issues/2323

The current version of avatar does not implement this feature.  The dark avatar svg is available in the component, the attribute and image swap need to be added.

bennypowers:
@anujsingla asked to grab this

anujsingla:
I have made changes for this issue - https://github.com/patternfly/patternfly-elements/pull/2357
closing this issue.

PR #2357, opened 2023-02-07
feat(avatar): add dark variant image

anujsingla:
## What I did

1. Added dark variant avatar for pf-avatar.
2. Added demo in the pf-avatar HTML file. 

fixes - https://github.com/patternfly/patternfly-elements/issues/2324

### DIFF
--- a/elements/pf-avatar/BaseAvatar.ts
+++ b/elements/pf-avatar/BaseAvatar.ts
@@ -29,7 +29,7 @@ export class BaseAvatar extends LitElement {
   @property({ reflect: true }) size: 'sm'|'md'|'lg'|'xl' = 'sm';
 
   /** Whether or not the Avatar image is dark */
-  @property({ type: Boolean, reflect: true, attribute: 'isdark' }) isDark = false;
+  @property({ type: Boolean, reflect: true }) dark = false;
 
   render() {
     return this.src != null ? html`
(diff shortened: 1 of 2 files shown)


elements/pf-avatar/BaseAvatar.ts
```diff
@@ -28,15 +28,17 @@ export class BaseAvatar extends LitElement {
...
     return this.src != null ? html`
       <img
         size=${this.size}
         alt=${this.alt ?? ''}
         @load="${(e: Event) => this.dispatchEvent(new AvatarLoadEvent(e))}"
         src=${this.src}>
-    ` : isDark ? html`
+    ` : this.isDark ? html`
       <svg xmlns="http://www.w3.org/2000/svg" style="enable-background:new 0 0 36 36" viewBox="0 0 36 36">
         <style>.st1,.st2{fill-rule:evenodd;clip-rule:evenodd;fill:#6a6e73}.st2{fill:#4f5255}</style><path d="M0 0h36v36H0z" style="fill:#212427"/>
```
bennypowers, line 43: this is fine, but it would be better to move the styles to svg attributes

even better would be to change the attribute values using the `svg` template tag, rather than switching off of `dark` and returning a completely new template.

there is an experimental `spread` helper but it's maybe not yet ready for production

bennypowers (changes requested):
Looks good! a couple small changes, and please add a changeset:

```md
---
"@patternfly/elements": minor
---
`<pf-avatar>`: added `dark` attribute (boolean)
```

bennypowers (approved):

PR #2358, opened 2023-02-07
feat(tools): add minify option to ts css-imports

bennypowers:
## What I did

1. Added `minify` option to css-imports typescript transform and enabled it in build config
2. Fixed `inline` option of css-imports transformer to allow ts sources to import multiple css sheets

## Testing Instructions

Say you had these files:
- a.css
  ```css
  a { color: var(--a); }
  ```
- b.css
  ```css
  b { color: var(--b); }
  ```
- m.ts
  ```ts
  import a from './a.css';
  import b from './b.css';
  export {a, b};
  ```

Try building using ttsc with this config:

```json
{
  "compilerOptions": {
    "module": "es2020",
    "moduleResolution": "Node",
    "strict": true,
    "target": "es2020",
    "plugins": [
      {
        "transform": "@patternfly/pfe-tools/typescript/transformers/css-imports.cjs",
        "inline": true,
        "minify": true
      }
    ]
  }
}
```
Before: 
```js
import { css } from "lit";
const a = css `a { color: var(--a); }`;
import { css } from "lit";
const b = css `b { color: var(--b); }`;
export { a, b };
```
After:
```js
import { css } from "lit";
const styles = css `a{color:var(--a);}`;
const styles = css `b{color:var(--b);}`;
export { a, b };
```

You can try this by building rhds. `rh-context-provider.ts` imports 2 stylesheets.

### DIFF
--- a/.changeset/curvy-windows-poke.md
+++ b/.changeset/curvy-windows-poke.md
@@ -1,0 +1,5 @@
+---
+"@patternfly/pfe-tools": patch
+---
+
+Allowed TypeScript modules to import multiple CSS modules when inlining CSS.
--- a/.changeset/itchy-roses-vanish.md
+++ b/.changeset/itchy-roses-vanish.md
@@ -1,0 +1,5 @@
+---
+"@patternfly/pfe-tools": minor
+---
+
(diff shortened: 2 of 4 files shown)


brianferry (approved):
Looks great, tested importing two stylesheets in pf-jump-links

PR #2360, opened 2023-02-07
fix(tools): find demos

bennypowers:
## What I did

1. find demos in both pfe and rhds repos

## Testing Instructions

1. analyze here and in rhds - manifest should include demos references


zeroedin (approved):
Tested locally and reviewed on DP.   👍 🚀

PR #2364, opened 2023-02-08
chore: post-checkout clear wireit cache

bennypowers:
## What I did

1. add a post-checkout git hook which clears the wireit cache when checking out a branch


.husky/post-checkout
```diff
@@ -0,0 +1,19 @@
...
+# Branch checkout, not file checkout
+# Not a clone operation
+# Old ref and new ref are different (i.e. actually switched branches)
+if   [ -z "$CI" ] \
+  && [ -n "$1" ] \
+  && [ "$3" = "1" ] \
+  && [ "$1" != "$3" ]; \
+then
+  NO_FORMAT="\033[0m"
+  C_YELLOW="\033[38;5;11m"
```
zeroedin, line 12: This color is hard to read on a light background terminal.  Not sure how much we want to get into this but assuming we should try to abide by accessibility principles here too, not quite sure how to handle this given a people can customize their terminal experiences, maybe a dark background on the text block itself?


Maybe something along these lines would work in both light and dark user preferences?
image

zeroedin: /cc @nikkimk

bennypowers: these are ansi colours, so i think it's up to the term theme

PR #2373, opened 2023-02-13
fix(tabs): make showOverflow protected

zeroedin:
## What I did

1. Made `showScrollButtons` protected in order to be able to read it from a child class.

## Notes:  

By making `showScrollButtons` protected I can read it in a child class downstream. This will allow the child class to know if the overflow has occurred or not.

### DIFF
--- a/elements/pf-tabs/BaseTabs.ts
+++ b/elements/pf-tabs/BaseTabs.ts
@@ -1,5 +1,6 @@
 import { LitElement, html } from 'lit';
 import { property, query, queryAssignedElements } from 'lit/decorators.js';
+import { classMap } from 'lit/directives/class-map.js';
 
 import { RovingTabindexController } from '@patternfly/pfe-core/controllers/roving-tabindex-controller.js';
 import { getRandomId } from '@patternfly/pfe-core/functions/random.js';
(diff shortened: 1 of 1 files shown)


bennypowers (commented):
Rather than loading this onto the base class, can it go on the container in the implementing class?

bennypowers:
Maybe we can move that logic into a controller and make the controller an implementation concern in *-tabs

zeroedin:
> Rather than loading this onto the base class, can it go on the container in the implementing class?

I would definitely like to do this elsewhere.  Wasn't ideal doing this here, rather the first idea that came to mind. The other thought that was maybe more appropriate was making `this.#showScrollButtons` publically readable, and reading the value to set the class in RHDS instead.

zeroedin:
> 



> Maybe we can move that logic into a controller and make the controller an implementation concern in *-tabs

Quite possible.  Going to see what the design specs say.  There is a possibility to need a container query controller as well to dictate what spacing is needed and at what size of the container, not just if the container is overflowed or not.  At that point, then it is a bigger change than what this draft PR is currently reflecting.

zeroedin:
Closed superseded by: https://github.com/patternfly/patternfly-elements/pull/2375

PR #2384, opened 2023-02-23
feat(tools) eslint-plugin

### LINKED ISSUE #2383: Lit decorator imports should be tree-shaken in source
### Discussed in https://github.com/patternfly/patternfly-elements/discussions/2382

<div type='discussions-op-text'>

<sup>Originally posted by **kylebuch8** February 22, 2023</sup>
I have a humble request. Can we be more specific when we import other modules into our components?

The reason I ask is because as we try to use the components and I take a look at the network calls to see what is impacting our page loads, I'm seeing a bunch of files that are included that are never used. For example, in this screenshot, all of the `query` decorators are imported even though they aren't being used.

image

In our component files, you'll see this a lot.
```
import { customElement, property } from 'lit/decorators.js';
```

And yes, that's nice to be able to do that and it's a convenience. However, what that ends up doing is importing all of the decorators that may not even be used. The `lit/decorators.js` file ends up importing all of these other files.

```
export * from '@lit/reactive-element/decorators/custom-element.js';
export * from '@lit/reactive-element/decorators/property.js';
export * from '@lit/reactive-element/decorators/state.js';
export * from '@lit/reactive-element/decorators/event-options.js';
export * from '@lit/reactive-element/decorators/query.js';
export * from '@lit/reactive-element/decorators/query-all.js';
export * from '@lit/reactive-element/decorators/query-async.js';
export * from '@lit/reactive-element/decorators/query-assigned-elements.js';
export * from '@lit/reactive-element/decorators/query-assigned-nodes.js';
```

When I try to create an optimized bundle, I end up including a bunch of unused code that increases the payload size. I thought that using something like Rollup would do some tree shaking and eliminate the unused code, but it doesn't seem to do that.

So, can we start writing code like this?

```
import { LitElement } from 'lit-element/lit-element.js';
import { html } from 'lit/html.js';
import { customElement } from 'lit/decorators/custom-element.js';
import { property } from 'lit/decorators/property.js';
import { state } from 'lit/decorators/state.js';
```
</div>

bennypowers:
Closes #2383, #2382 

## What I did

1. added `@patternfly/eslint-plugin-elements`
2. add `no-lit-decorators-index-import` to above
3. lint sources with same

## Testing Instructions

1. check demos of affected elements


heyMP (approved):
Legitimate Goodness Terminated Middlemen

heyMP:
A-Mazing!

Screenshot 2023-02-24 at 12 53 06 PM
Screenshot 2023-02-24 at 12 52 40 PM
Screenshot 2023-02-24 at 12 33 22 PM

PR #2410, opened 2023-03-07
fix(core): resolves #2408 for roving-tabindex-controller

### LINKED ISSUE #2408: roving-tabindex-controller ignores for expandable items
## roving-tabindex-controller ignores for expandable items

For components with complex navigation, such as navigation with expandable submenus, roving-tabindex controller has the following issues:
- listeners are set to host when only the items should have listeners.
- vertical navigation is disabled for `aria-expanded=true` items when navigation should be enabled and include navigation to any children that are on its items list.
- currently allows navigation for `aria-expanded=false` items when it should skip children

### Steps to reproduce

1. Go to the [rh-menu demo](https://deploy-preview-679--red-hat-design-system.netlify.app/components/menu/demo/).
2. Using a keyboard focus on the menu button while closed.
3. Press the down arrow. Nothing should happen.
4. Press enter to open the menu.
5. Press the down arrow. Focus should be set on the active subitem.

### Expected behavior

1. Go to the [rh-menu demo](https://deploy-preview-679--red-hat-design-system.netlify.app/components/menu/demo/).
2. Using a keyboard focus on the menu button while closed.
3. Press the down arrow. 
4. If item is part of a larger roving tabindex group, like a parent menu, focus should be on the next sibling. If item is *not* part of a larger roving tabindex group, nothing should happen. 
5. Press enter to open the menu.
6. Press the down arrow. Focus should be set on the active subitem.

nikkimk:
## What I did

1. Added an optional parameter to `roving-tabindex-controller`'s `initItems()` to limit `keydown` listener to a specific container element instead of host
2. Removed `aria-expanded=true` from `horizontalyOnly` keyboard navigation

## Testing Instructions

Since this is a controller, the way I had to test it was to temporarily do the following:
1. Copy the updated `roving-tabindex-controller.js` to `pf-tabs` and `pf-jumplinks` element folders.
2. Change the import path in `pf-tabs/BaseTabs.js` and `pf-jumplinks/pf-jumplinks.js` to `./roving-tabindex-controller.js`.
3. Determine that these changes do not break tabs and jumplinks.
4. Undo your changes.

### DIFF
--- a/core/pfe-core/controllers/roving-tabindex-controller.ts
+++ b/core/pfe-core/controllers/roving-tabindex-controller.ts
@@ -15,6 +15,9 @@ export class RovingTabindexController implements ReactiveController {
   /** active focusable element */
   #activeItem?: HTMLElement;
 
+  /** closest ancestor containing items */
+  #itemsContainer?: HTMLElement;
+
   /** array of all focusable elements */
   #items: HTMLElement[] = [];
 
@@ -97,8 +100,8 @@ export class RovingTabindexController implements ReactiveController {
     const horizontalOnly =
         !item ? false
       : item.tagName === 'SELECT' ||
-        item.getAttribute('aria-expanded') === 'true' ||
         item.getAttribute('role') === 'spinbutton';
+
 
     switch (event.key) {
       case 'ArrowLeft':
@@ -188,7 +191,7 @@ export class RovingTabindexController implements ReactiveController {
   /**
    * from array of HTML items, and sets active items
    */
-  initItems(items: HTMLElement[]) {
+  initItems(items: HTMLElement[], itemsContainer?: HTMLElement) {
     this.#items = items ?? [];
     const focusableItems = this.#focusableItems;
     const [focusableItem] = focusableItems;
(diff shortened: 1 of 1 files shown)


core/pfe-core/controllers/roving-tabindex-controller.ts
```diff
@@ -188,17 +191,28 @@ export class RovingTabindexController implements ReactiveController {
   /**
    * from array of HTML items, and sets active items
    */
-  initItems(items: HTMLElement[]) {
+  initItems(items: HTMLElement[], itemsContainer?: HTMLElement) {
```
bennypowers, line 194: needs a changeset

.changeset/fresh-pumas-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@patternfly/pfe-core": patch
```
bennypowers, line 2:
```suggestion
"@patternfly/pfe-core": major
```

core/pfe-core/controllers/roving-tabindex-controller.ts
```diff
@@ -188,17 +191,28 @@ export class RovingTabindexController implements ReactiveController {
...
+        this.#itemsContainer.removeEventListener('keydown', this.#onKeydown.bind(this));
+      }
+      this.#itemsContainer = itemsContainer || this.host;
+      this.#itemsContainer.addEventListener('keydown', this.#onKeydown.bind(this));
+    }
   }
 
   hostConnected() {
-    this.host.addEventListener('keydown', this.#onKeydown.bind(this));
+    this.#itemsContainer = undefined;
```
bennypowers, line 216: this will blow away the state

```js
const el = document.createElement('has-rti-controller');
$('#container').append(el); // ok
el.rtiController.initItems(...items); // alright
$('#other-container').append(el); // oops... hostConnected()
```

PR #2412, opened 2023-03-07
docs: add htmlexample shortcode

zeroedin:
## What I did

1. Added `{% htmlexample %}` short code to display HTML example code in a disclosure.

## Testing Instructions

1. View [Deploy preview](https://deploy-preview-2412--patternfly-elements.netlify.app/) and look at the usage of the shortcodes for hiding and displaying the HTML examples.

## Notes to Reviewers

I think we might need to apply some style to the `<details>` generated by the shortcode.  Ideas welcome!


brianferry:
> I think we might need to apply some style to the <details> generated by the shortcode. Ideas welcome!

Might be worth making a pfe disclosure element as well as an rh-disclosure element down the line both using the `<details>` element.

zeroedin:
image

@brianferry the fact that when you quoted this bit of text it actually turned into a details element is priceless.

brianferry (approved):
Lemurs Galloping Towards Munchies

Issue #2349, opened 2023-02-06
BaseAvatar properties should be abstract


bennypowers:
## Description of the issue

`BaseAvatar` implements its own `size` property
https://github.com/patternfly/patternfly-elements/blob/905465f719414ce54596e9c88fa20e4dc6334831/elements/pf-avatar/BaseAvatar.ts#L28-L29
this could present a problem in the future, 
```ts
class JazzyAvatar extends BaseAvatar {
  @property() size?: 'louis'|'miles'|'john'|'duke';
}
```

>    Property 'size' in type 'JazzyAvatar' is not assignable to the same property in base type 'BaseAvatar'.
     Type '"louis" | "miles" | "john" | "duke" | undefined' is not assignable to type '"sm" | "md" | "lg" | "xl"'.
       Type 'undefined' is not assignable to type '"sm" | "md" | "lg" | "xl"'.

BaseAvatar's `size` property should be an abstract optional string

bennypowers:
@anujsingla if you want to close this in the PR for #2324 that's fine
cc @Nouveau this will be relevant to work on rh-avatar, but shouldn't break anything - maximum you'll need to make sure rh-avatar implements `size`

PR #2414, opened 2023-03-08
fix(avatar)!: make size property as abstract

anujsi:
## What I did

1. Made size property and BaseAvatar class as abstract

fixes - #2349

### DIFF
--- a/elements/pf-avatar/BaseAvatar.ts
+++ b/elements/pf-avatar/BaseAvatar.ts
@@ -16,7 +16,7 @@ export class AvatarLoadEvent extends Event {
  * @summary For displaying a user's avatar image
  */
 
-export class BaseAvatar extends LitElement {
+export abstract class BaseAvatar extends LitElement {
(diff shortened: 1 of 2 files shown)


elements/pf-avatar/BaseAvatar.ts
```diff
@@ -26,15 +26,15 @@ export class BaseAvatar extends LitElement {
   @property({ reflect: true }) alt?: string = 'Avatar image';
 
   /** Size of the Avatar */
-  @property({ reflect: true }) size: 'sm'|'md'|'lg'|'xl' = 'sm';
+  abstract size?: string;
```
bennypowers, line 29: this is, unfortunately, a breaking change, and will thus require a changeset with a `major` version. That also means we're unlikely to merge this quickly

bennypowers (changes requested):

bennypowers:
Closed by  #2630

PR #2424, opened 2023-03-13
docs: updating dev server header to match live site 

brianferry:
## What I did

1. Updated dev server header to match 11ty doc site styles
2. Added `repoHost` to `PfeConfig` which enables customizing the dev server repository host name and icon

## Testing Instructions

1. Checkout branch, run `npm run start`
2. Add a `.pfe.config.json` with the following JSON, and the Icon should change to GitLab.  Adding a blank string for `repoHost` should result in a generic Git logo.
```json
{
  "site": {
    "repoHost": "GitLab"
  }
}
```

## Notes for Reviewers

1. Ensure that the new header works across varying screen sizes.
2. Suggestions welcome on if the configuration of `repoHost` should exist in the `site` block or in at the same level as `sourceControlURLPrefix` and `demoURLPrefix` in `.pfe.config.json` see [RHDS's config](https://github.com/RedHat-UX/red-hat-design-system/blob/main/.pfe.config.json) as a reference of a filled out JSON config file.


tools/pfe-tools/dev-server/demo.css
```diff
@@ -44,6 +44,57 @@ body.menu-open {
...
+}
+
+.logo-bar > a img {
+  height: 100%;
+}
+
+.logo-bar svg {
+  height: 100%;
+}
+
```
eyevana, line 81: Active, focus, and hover state in mobile should be white as well:

```suggestion

.logo-bar #hamburger {
  --pf-c-button--m-plain--Color: #fff;
  --pf-c-button--m-plain--active--Color: #fff;
  --pf-c-button--m-plain--focus--Color: #fff;
  --pf-c-button--m-plain--hover--Color: #fff;
}

```

tools/pfe-tools/dev-server/index.html
```diff
@@ -24,28 +24,54 @@
...
-                aria-controls="sidebar"
-                style="--pf-c-button--m-plain--Color:white"
-                label="Expand Main Navigation">
-        <pf-icon icon="bars"></pf-icon>
+    <div class="logo-bar">
+      <pf-button
+        id="hamburger"
+        plain
+        aria-controls="sidebar"
+        style="--pf-c-button--m-plain--Color:white"
```
eyevana, line 39: Won't need this anymore if previous suggestion accepted in `demo.css`:
```suggestion
```

tools/pfe-tools/config.ts
```diff
@@ -35,9 +37,10 @@ export interface PfeConfig {
 const SITE_DEFAULTS: Required<SiteOptions> = {
   description: 'PatternFly Elements: A set of community-created web components based on PatternFly design.',
   favicon: '/brand/logo/svg/pfe-icon-blue.svg',
```
eyevana, line 39: I know favicon was mentioned during office hours this morning, but I don't remember was discussed. This is using the old logo. Should we update to `favicon.ico` under `/docs/images/`?

eyevana (approved):
I updated the config with no issues. If I had to choose, I would probably place `repoHost` under `site` as well, but it feels like it could go both ways I'm also wondering if we could just extract the host from the sourceControlURLPrefix instead of adding another property in the config file? I had some minor CSS suggestions too, but overall lgtm!

bennypowers:
- can we infer the icon from the `sourceControlUrlPrefix` instead?
- can we add an option (either a string or `(tagName: string) => string)` that returns the pathname) to customize the path name for elements? that would let us easily config `/elements/` vs `/components/` or `/patterns/` etc.

brianferry:
> * can we infer the icon from the `sourceControlUrlPrefix` instead?
> * can we add an option (either a string or `(tagName: string) => string)` that returns the pathname) to customize the path name for elements? that would let us easily config `/elements/` vs `/components/` or `/patterns/` etc.

Can you elaborate a little more on #2? I see in `PfeConfig` we have `elementsDir` which looks like it would let us switch the directory like you're mentioning or were you thinking of something different?

brianferry:
Added the logic to pull the icon from the url and removed the repoHost config var

Issue #2416, opened 2023-03-10
[bug] Accordion fixed panel variant


zeroedin:
## Description of the issue
The fixed panel variant of the Accordion does not behave as intended.  The panel is not open on load.

### Impacted component(s)

- Accordion

### Steps to reproduce

1. View [Accordion demo](https://patternflyelements.org/components/accordion/demo/) compare to functionality on https://www.patternfly.org/v4/components/accordion/

### Expected behavior

The fixed panel should be open on load

### Screenshots

![image](https://user-images.githubusercontent.com/13666/224362238-9f41cd3b-7f78-49a3-9086-76e39ffcc35e.png)

Issue #2397, opened 2023-02-28
[bug][docs] Fix numerous docs style issues


markcaron:
There are multiple style issues throughout PFE doc pages:

- [ ] Add new site header style (@markcaron)
  - [ ] Add new PFE logo
  - [ ] Fix v1/v2 toggle to be consistent in style
  - [ ] Have v1/v2 toggle go directly to All Components (not individual components)
- [ ] Page headings
  - [ ] Add back in gray background behind page headings
  - [ ] Move overview description into page heading
- [ ] Add max-width to content
- [ ] Bands
  - [ ] Fix spacing between "bands"
  - [ ] Remove unnecessary `background-image` SVGs from bands
- [ ] Components
  - [ ] Accordion
    - [ ] API content/sections doesn't match PF (should mimic)
  - [ ] Clipboard copy
    - [ ] Icon doesn't match PatternFly's
    - [ ] Icon size
    - [ ] Bottom border doesn't lineup is both versions
- [ ] Font changes on in-page demos, like [accordion's](https://patternflyelements.org/components/accordion/) constructed stylesheet (i.e., "Red Hat Text" vs "RedHatTextUpdated")

zeroedin:
> * [ ]  Fix v1/v2 toggle to be consistent in style
> * [ ]  Have v1/v2 toggle go directly to All Components (not individual components)

I believe we are deferring these due to symlinking to the `v1` docs and not directly maintaining their templates.

zeroedin:
> * [ ]  Icon size

The icon sizes are not the same as PFv4.  Currently, this icon is showing up at 10px which is the default size of no size variant give (eg. size="md").  

`pf-icon` API does not allow for a 16px icon, even though this is the size that PFv4 itself uses, this issue has come up in several places including `pf-tabs` and downstream in RHDS. 

[Patternfly icon ](https://www.patternfly.org/v4/components/icon) clearly states the same intent.  But then in use cases through out a size of 16px is used in components.  However, these components are not directly using the PFIcon component like we are in web components.

zeroedin:
> [ ] Font changes on in-page demos, like [accordion's](https://patternflyelements.org/components/accordion/) constructed stylesheet (i.e., "Red Hat Text" vs "RedHatTextUpdated")

[...]

PR #2427, opened 2023-03-14
fix(accordion): fixing font-family styling, fixed height demo 

brianferry:
## What I did

1.  Updated the font-family for `pf-accordion` to "Red Hat Text"
2. Updated the demo page to expand the fixed height demos first panel automatically on load

## Related Issues

- #2397 
- #2416

### DIFF
--- a/elements/pf-accordion/pf-accordion-header.css
+++ b/elements/pf-accordion/pf-accordion-header.css
@@ -53,7 +53,7 @@
   font-family:
     var(--pf-c-accordion__toggle--FontFamily,
       var(--pf-global--FontFamily--redhat-updated--heading--sans-serif,
-        "RedHatTextUpdated",
+        "Red Hat Text",
         helvetica,
         arial,
         sans-serif));
--- a/.changeset/silent-bats-crash.md
+++ b/.changeset/silent-bats-crash.md
@@ -1,0 +1,5 @@
+---
+"@patternfly/elements": patch
+---
+
+`<pf-accordion>`: fixed font family from "RedHatTextUpdated" to "Red Hat Text".  Updated fixed height demo to auto expand on load.
--- a/.changeset/silent-bats-crash.md
+++ b/.changeset/silent-bats-crash.md
@@ -2,4 +2,4 @@
 "@patternfly/elements": patch
 ---
 
-`<pf-accordion>`: fixed font family from "RedHatTextUpdated" to "Red Hat Text".  Updated fixed height demo to auto expand on load.
+`<pf-accordion>`: fixed font family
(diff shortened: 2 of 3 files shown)


elements/pf-accordion/pf-accordion-header.css
```diff
@@ -53,7 +53,7 @@
   font-family:
     var(--pf-c-accordion__toggle--FontFamily,
       var(--pf-global--FontFamily--redhat-updated--heading--sans-serif,
-        "RedHatTextUpdated",
```
bennypowers, line 56: this is actually incorrect?

https://www.patternfly.org/v4/components/accordion/ shows:
```css
font-family: "RedHatTextUpdated", "Overpass", overpass, helvetica, arial, sans-serif
```

bennypowers (changes requested):

elements/pf-accordion/pf-accordion-header.css
brianferry: Taken from https://github.com/patternfly/patternfly-elements/issues/2397, we're not sending out a RedHatTextUpdated from the issue itself so we would need to update this to Red Hat Text.

bennypowers: but accordion needs to match up with pf styles. better to load the updated font on our site rather than diverge, IMO

bennypowers: I can see this both ways. In the past, we have said "if we can do better than the pfv4 components...", So I think you could definitely make the argument that this is one of those cases

On the other hand, the source material clearly specifies the "updated" font face. It seems hard to justify making this change just for our docs site, when we could just lost that font on the site. Pfe is supposed to be *visually* agnostic to upstream design choices so long as it doesn't affect accessibility

brianferry: @markcaron thoughts?  I agree with what Benny is saying here that pfe should be functionally the same as pf w/r/t styling but wanted to see your thoughts on updating it to RedHatText vs keeping it as RedHatTextUpdated?

bennypowers: but why diverge from patternfly at all? that's why we have the RHDS font tokens

bennypowers: > can we not add both to the font stack?

*we* can, but *PatternFly* doesn't, so we *shouldn't*. PFE is not supposed to have opinions on things like font stacks, and I don't want to maintain a "tweaks file" which specifies where and how PFE diverges from PatternFly design specs. If we want to change this in the font stack, we have to make a PR to PF core.

bennypowers:
I'm going to close this because really it's out of scope.
- If we want to change the font stack, we need to do that upstream
- If we want to change the fonts only on our website, we should do that in /docs

PR #2435, opened 2023-03-23
fix(accordion): fixing header tag, fixing heading text attribute  

brianferry:
## What I did

1. Added an `anonymous` slot checker in the `pf-accordion-header` component. 
  - This checks for whether an `anonymous` slot has been passed in for the header
    - DOES NOT have an `anonymous` slot = we add the header tag to the shadow dom
    - DOES have an `anonymous` slot = we DO NOT add the header tag to the shadow dom
 2. We have / had an issue with the `heading-text` attribute not showing correctly in the header, this has been fixed
 3. Adding a demo to show the attributes being used instead of the `anonymous` slot

## Testing Instructions

1. [DP](https://deploy-preview-2435--patternfly-elements.netlify.app/components/accordion/demo/)
2. Open the developer console and inspect the accordion demos
3. Verify that the accordion headers with slotted in content do not have nested header tags in the shadow dom
4. Verify that the accrdion headers in the demo that have text / tags passed in via attributes have the nested header tags in the shadow dom. 
5. Verify that the text passed in through `heading-text` is showing in the header as expected

### DIFF
--- a/elements/pf-accordion/BaseAccordionHeader.ts
+++ b/elements/pf-accordion/BaseAccordionHeader.ts
@@ -8,6 +8,7 @@ import { BaseAccordion } from './BaseAccordion.js';
 import { ComposedEvent } from '@patternfly/pfe-core';
 import { getRandomId } from '@patternfly/pfe-core/functions/random.js';
 import { Logger } from '@patternfly/pfe-core/controllers/logger.js';
+import { SlotController } from '@patternfly/pfe-core/controllers/slot-controller.js';
 
 import style from './BaseAccordionHeader.css';
 
@@ -40,6 +41,8 @@ export abstract class BaseAccordionHeader extends LitElement {
 
   #logger = new Logger(this);
 
+  #slots = new SlotController(this, null);
+
   override connectedCallback() {
     super.connectedCallback();
     this.addEventListener('click', this.#onClick);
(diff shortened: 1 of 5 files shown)


elements/pf-accordion/BaseAccordionHeader.ts
```diff
@@ -71,10 +79,23 @@ export abstract class BaseAccordionHeader extends LitElement {
   renderAfterButton?(): TemplateResult;
 
   override render(): TemplateResult {
+    const hasAnonymousSlot = this.#slots.getSlotted()?.length > 0;
     const tag = unsafeStatic(this.headingTag);
     const ariaExpandedState = String(!!this.expanded) as 'true'|'false';
     return staticH`
-      <${tag} id="heading">
+      ${!hasAnonymousSlot ?
+        staticH`
```
bennypowers, line 87: see https://github.com/RedHat-UX/red-hat-design-system/pull/787

bennypowers (changes requested):

elements/pf-accordion/BaseAccordionHeader.ts
bennypowers, line 1: @brianferry please review this last refactoring commit and if it's good, resolve this comment to merge

bennypowers (approved):

elements/pf-accordion/BaseAccordionHeader.ts
bennypowers: also, please add a changeset which describes the user-facing changes

bennypowers: this should also be tested against rh-accordion

brianferry: @heyMP - 

image

Pulled in the new updated pf-accordion files & built & ran it w/ `main` for footer, no longer seeing the shadow-root h3s, wanting to confirm everything looks good to you too

bennypowers (approved):

PR #2438, opened 2023-03-24
fix(accordion): large panel accent not covering whole side panel 

brianferry:
## What I did

1.  Added `overflow:hidden` to the panel so that the accent will span the whole height of the div in cases where the slotted content has margins / padding.

### DIFF
--- a/.changeset/wild-falcons-do.md
+++ b/.changeset/wild-falcons-do.md
@@ -2,4 +2,4 @@
 "@patternfly/elements": patch
 ---
 
-`pf-accordion-panel`: - removing position relative from body class which was causing the accent color to not fill the entire panel
+`pf-accordion-panel`: - adding overflow hidden to the body tag for an issue where the border would not span the whole height of the panel.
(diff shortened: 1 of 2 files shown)


.changeset/wild-falcons-do.md
```diff
@@ -0,0 +1,8 @@
+---
+"@patternfly/elements": patch
+---
+
+`pf-accordion`: 
+  - fixed issue where accent would not display full height if the following conditions were met:
+    - `pf-accordion` was set to `large` 
+    - `pf-accordion-panel` slotted content had padding or margins
```
bennypowers, line 8:
```suggestion
`<pf-accordion>`: fixed issue where accent would not display full height if the following conditions were met:
    - `pf-accordion` was set to `large` 
    - `pf-accordion-panel` slotted content had padding or margins 
```

bennypowers:
Is there a demo case to illustrate the problem?

bennypowers (approved):

PR #2455, opened 2023-03-28
feat: pf-pagination (bottom variant)

eyevana:
## What I did

1.

## Testing Instructions

1.

## Notes to Reviewers

1.

### DIFF
--- a/elements/pf-button/pf-button.css
+++ b/elements/pf-button/pf-button.css
@@ -570,7 +570,7 @@ button:hover {
 :host([block]) button {
   display: flex;
   width: 100%;
-  justify-content: var(--pf-c-button--JustifyContent, center);
+  justify-content: center;
 }
 
 /******************************
--- a/elements/pf-pagination/pf-pagination.css
+++ b/elements/pf-pagination/pf-pagination.css
@@ -112,7 +112,6 @@
 
 #container #options-menu #menu-list .menu-item {
   white-space: nowrap;
-  --pf-c-button--JustifyContent: start;
   --pf-c-button--PaddingTop: var(--pf-c-options-menu__menu-item--PaddingTop, var(--pf-global--spacer--sm, .5rem));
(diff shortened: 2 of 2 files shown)


bennypowers:
PTAL at the prior art in rh-pagination, and see how much you can move to shared code like a controller or some such

eyevana:
@bennypowers yep, I'll compare the two implementations now.

elements/pf-pagination/pf-pagination.ts
```diff
@@ -0,0 +1,309 @@
...
+          <pf-button id="menu-toggle"
+                     aria-expanded=${this._expanded}
+                     aria-haspopup="listbox"
+                     icon="caret-${this._expanded ? 'up' : 'down'}" 
+                     icon-position="right"
+                     part="menu-toggle"
+                     plain
+                     type="button"
+                     @click=${this.#toggleExpanded}>
+              <span><b>${this.#firstOfPage()}-${this.#lastOfPage()}</b> ${TITLES.ofWord} <b>${this.count}</b></span>
```
nikkimk, line 108: `${this.count}` doesn't seem to be working in demo.

nikkimk (commented):
Overall a good draft. I have a few suggestions marked. 

(And it gave me an opportunity to really, really have to think about one of those tricker accessibility questions, so thanks much for helping me get better at my work!)

elements/pf-button/pf-button.css
```diff
@@ -570,7 +570,7 @@ button:hover {
 :host([block]) button {
   display: flex;
   width: 100%;
-  justify-content: var(--pf-c-button--JustifyContent, center);
```
bennypowers: why this change?

eyevana: I removed this and opened that separate PR to add the CSS var, but opting for a `part` on the button instead.

PR #2457, opened 2023-03-28
docs: set text color to white on hover

### LINKED ISSUE #2453: [bug][docs] Get Started CTA on homepage inaccessible focus+hover state
Even though it's highly unlikely that there will be a focus + hover state at the same time, it can still happen and it looks weird. Can we make the text white?

image

eyevana:
Closes #2453

## What I did

1. Fixed get started CTA color on hover 

## Testing Instructions

1. Open the docs link and hover over the get started CTA (also set focus + hover in dev tools) and verify the text color is white.

## Notes to Reviewers

1. I don't think we're using `pfe-cta` in the docs anymore? Should we remove those custom variable overrides?

### DIFF
--- a/docs/home.css
+++ b/docs/home.css
@@ -44,14 +44,17 @@
   --pf-cta--Color: #06c;
 }
 
+#home-header a.cta.primary:hover, 
+#home-header a.cta.primary:focus {
+  --pf-cta--BackgroundColor: transparent;
+}
+
 #home-header a.cta.primary:hover {
   --pf-cta--Color--hover: var(--pf-theme--color--white, #fff);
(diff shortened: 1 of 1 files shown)


brianferry:
image

Now I'm not seeing any text when tab-focused & hovered, I think the focus should be the same color as the hover? color: var(--pf-theme--color--accent, #0066cc);

eyevana:
@brianferry check out the latest. Should be fixed now. Thank you! :)

brianferry:
This looks good to me!

brianferry (approved):

zeroedin (approved):

PR #2458, opened 2023-03-28
fix(accordion): removing auto focus on expanded in the header

### LINKED ISSUE #2454: [bug][docs] Accordion page load sets focus further down the page
Not sure why, but the [accordion page](https://patternflyelements.org/components/accordion/) jumps the focus down the page automatically when it loads.

brianferry:
## What I did

1. Removed auto focus on expanded for the accordion header

Closes #2454 

This is so that if you have an expanded accordion down the page it doesn't automatically scroll the user down to the accordion.

## Testing Instructions

1. Check out the DP and verify the accordion docs page doesn't auto focus on the expanded header in the demo docs

### DIFF
--- a/.changeset/stale-boxes-watch.md
+++ b/.changeset/stale-boxes-watch.md
@@ -2,4 +2,15 @@
 "@patternfly/elements": patch
 ---
 
-`accordion` - removed auto focus on the header when the expanded attribute is added. This is being removed so that if a page loads with an expanded accordion it does not scroll the user to it automatically.
+`accordion` - removed auto-focus on accordion header when the header contains the `expanded` attribute.  
+
+**Example**
+```
+  <pf-accordion>
+    <pf-accordion-header expanded>
+      Header Item
+    </pf-accordion-header>
+  </pf-accordion>
+```
+
+Previously, if you set the expanded attribute to expand the header when the component loaded it would also focus the element on the page for the user.  This has been removed.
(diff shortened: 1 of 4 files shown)


brianferry:
Actually not sure about this one, it seems like the focus should be on the header when it is expanded but I'm not sure if we want it to ignore focus when it's expanded on load or if we just want to update the docs for now and decide if we want to fix in a future update.  Thoughts @markcaron @zeroedin @bennypowers ?

zeroedin:
> Actually not sure about this one, it seems like the focus should be on the header when it is expanded

On page load focus / `document.activeElement` should be at the body element of the page.  When a url#hash property (URI fragment) is present, then the reference to the anchor should be the first tabbable item.  Is that what you are trying to portray with `:focus` here when an accordion header is expanded?  Or maybe I'm misunderstanding the intent/problem, if so let me know.

From my understanding when an accordion header is `expanded` on a page it doesn't necessarily mean the user wants to immediately navigate to that item which is what setting focus there does.  In contrast, when an url#hash is used then the intent is explicit by the user that they want to navigate to that spot.  

Even still when the url#hash is used I don't think you'd set focus rather you'd want to set maybe tabindex, and only if needed, but I think given the basic structure of the HTML  (no iframes etc) the browser should handle that for us.  @nikkimk does this sound correct / am I explaining that correctly?

zeroedin:
I'm pretty sure the only change that needs to be made is the removal of `header.focus()` in the `expand()` method in BaseAccordion.ts L:375.  The click or keypress events are enough to give focus, don't need to explicitly set it on expand.

bennypowers (approved):

PR #2470, opened 2023-04-04
feat(button): expose `button` part

eyevana:
## What I did

1. I added a new CSS variable to override the positioning of the `block` variant. I'm using a list of `pf-button`s in [`pf-pagination`](https://deploy-preview-2455--patternfly-elements.netlify.app/components/pagination/demo/), under the per-page menu, and I need to set the justify-content property to `start` instead of `center`. 

## Testing Instructions

1. Verify that `center` is the default position on the `block` variant.
2. Set `--pf-c-button--block--JustifyContent` to something other than center and verify the change reflects correctly.

## Notes to Reviewers

1. Let me know if the variable name I chose works. I followed Patternfly's naming convention, but I wasn't sure if there was a separate one I should follow for variables that we define in our project.


elements/pf-button/pf-button.css
```diff
@@ -570,7 +570,7 @@ button:hover {
 :host([block]) button {
   display: flex;
   width: 100%;
-  justify-content: center;
+  justify-content: var(--pf-c-button--block--JustifyContent, center);
```
bennypowers, line 573: can we do this with a ::part instead i wonder? Would that be better?

heyMP: I think this would be a good case for `::part`.  @eyevana if we find ourselves adding CSS variables for individual properties on a single element in the shadowdom then I would propose we add a shadow part.

pf-button/BaseButton.ts
```js
  override render() {
      <button part="button" ...> ... </button>
    `;
  }
```

styles.css
```css
pf-button[block]::part(button) {
  justify-content: center;
  margin-block-start: 3.14159px;
}
```

eyevana: Gotcha! I'll update my changes now and add a part instead. Thanks for the feedback.

eyevana: ping =) sorry to be annoying, I just need this little itty bitty change for the work I'm doing in cp.

bennypowers: not annoying, thanks for keeping reviewers focused.

elements/pf-button/BaseButton.ts
bennypowers, line 1: This needs a `@csspart` jsdoc tag

bennypowers (changes requested):

.changeset/plenty-clouds-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+"@patternfly/elements": minor
+---
+
+`<pf-button>`: expose `button` part for styling the internal button element
```
eyevana, line 5: Thank you for adding this @bennypowers...my bad for rushing >.<

bennypowers (approved):

Issue #2445, opened 2023-03-27
[docs]: align component page header and page css to patternfly.org


bennypowers:
Compare

https://patternflyelements.org/components/accordion/

https://www.patternfly.org/v4/components/accordion/

- the background colour for the main body of the page should be grey
- the overview text should be within the header

depends on #2444

PR #2471, opened 2023-04-10
docs: fix align component page header and css

anujsingla:
## What I did

1. Added background color grey of the body
2. Added Overview text in the header. 

Closes https://github.com/patternfly/patternfly-elements/issues/2445

## Testing Instructions

1. https://deploy-preview-2471--patternfly-elements.netlify.app/components/accordion/


elements/pf-accordion/docs/pf-accordion.md
```diff
@@ -90,7 +90,7 @@
   </pf-accordion>
   ```
 
-  ### Single Expanded Behavior
+  ### Single Expanded Behavior {class="anuj"}
```
zeroedin, line 93: Maybe the solution here is to mock up the PatternFly components pages even closer by breaking these out into a h3 Examples and each being a subsection of `{% band header="Single Expanded Behavior", level="4" %}` instead of lumping them all under "Usage" as we currently are doing.  This would give you the background separation on each without having to directly modify `band.njk`.

```Suggestion
{% endband %}

anujsingla: Thanks for the suggestions @zeroedin 
I have added div as a parent element for all the examples in the *.md file like this:
example-preview -> This CSS class will add white background color and padding as per upstream.
Please check the code for more details. Please let me know if you have suggestions.

<div class="example-preview">
<example code pf element render>
</div>

bennypowers:
@anujsi please rebase, as #2412 is merged

bennypowers (approved):
Good Stuff, @anujsi. LGTM.

PR #2485, opened 2023-04-25
fix(tabs): fix focus on keyboard navigation for safari

### LINKED ISSUE #2484: [bug] pf-tabs tab focus/active in safari
There is an issue with safari when using keyboard navigation to move focus between tabs.  The `active` attribute gets set correctly however when roving-tabindex-controller (RTI) sets focus on the activated tab, the `document.activeElement` is not updated with that new tab and focus styles then do not render only the active attribute style.

Background:
BaseTabs attempts to update the #activeItem for RTI on the #onTabExpand handler.  Maybe this causes a race condition on returning the correct #activeItem in Safari?  It is might be possible then that this issue could be exposed in any browser on slower machines/mobile devices as well.

zeroedin:
### What I did

1. Added a check in the tabExpand event handler to check the event.tab object ensuring it is not the same as the tabindex activeItem before updating.  
2. Override `focus()` on tab applying focus to the shadowroot `button`.

Closes #2484 

### Reviewer Notes

Changes here need to be tested upstream in RHDS.  You will need to copy  or `npm link` the compiled files into RHDS to test.

### Testing instructions

1. Ensure that all arrow keys, page up, page down, and click events continue work in all browsers under different network conditions and load using the [deploy preview](https://deploy-preview-2485--patternfly-elements.netlify.app/components/tabs/demo/)
3. Check to make sure focus styles are applied correctly.

### DIFF
--- a/.changeset/perfect-dragons-double.md
+++ b/.changeset/perfect-dragons-double.md
@@ -2,6 +2,6 @@
 "@patternfly/elements": patch
 ---
 
-`BaseTabs`: 
+`BaseTab`: 
  - fixed Safari focus issue on keyboard navigation
  - fixed Safari focus issue on mouse click
--- a/.changeset/perfect-dragons-double.md
+++ b/.changeset/perfect-dragons-double.md
@@ -3,5 +3,5 @@
 ---
 
 `BaseTabs`: 
- - fixes Safari focus issue on keyboard navigation
- - fixes Safari focus issue on tab mouse click
+ - fixed Safari focus issue on keyboard navigation
+ - fixed Safari focus issue on mouse click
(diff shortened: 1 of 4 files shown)


zeroedin:
After a couple different tries this still seems to be an issue upstream.  At this point I'm not sure its a race condition but instead maybe a browser quirk with how safari deals with tabindex and setting of document.activeElement after focus is applied to a tab.  However that theory does not explain why focus seemingly gets applied to  a `pf-tab` and not for `rh-tab` when there is no fundamental difference between the two at code level.  They both inherit from BaseTabs/BaseTab and neither have specific code that should interfere with focus setting.

nikkimk (approved):
lgtm

heyMP (approved):
LGTM 👍

PR #2493, opened 2023-05-08
fix(accordion): adding rti for keyboard navigation, removed internal navigation

brianferry:
## What I did

1.  Remove the accordion specific keyboard navigation
2. Add the RTI (Roving Tab Index) controller to accordion after the accordion is initialized
3. This will fix the arrow up / arrow down issues related to https://github.com/RedHat-UX/red-hat-design-system/issues/971

## Testing Instructions

1.[ Go to DP demo page](https://deploy-preview-2493--patternfly-elements.netlify.app/components/accordion/demo/)
2. Tab down to the accordion with nested accordions
3. Validate that the instructions located at https://github.com/RedHat-UX/red-hat-design-system/issues/971 work here 

Copied below for clarity

```
Steps to reproduce

- Run accordion demo locally
- Find nested example
- Expand first accordion item
- Tab inside to the child accordion
- Hit the down arrow (assuming it should go to the next item in the child accordion)
- Focus jumps to the parent accordion header

```


bennypowers (changes requested):

elements/pf-accordion/BaseAccordion.ts
```diff
@@ -78,6 +82,12 @@ export abstract class BaseAccordion extends LitElement {
     return this.#allPanels();
   }
 
+  get activeHeader() {
```
bennypowers, line 85: should be private to avoid a minor release

bennypowers (changes requested):

bennypowers (approved):

.changeset/empty-trainers-tickle.md
```diff
@@ -0,0 +1,4 @@
+---
+"@patternfly/elements": patch
+---
+`<pf-accordion>`: fixd keyboard navigation inside of nested accordions
```
bennypowers, line 4:
```suggestion
`<pf-accordion>`: fixed keyboard navigation inside of nested accordions
```

PR #2502, opened 2023-05-23
fix(icon): use icons package

bennypowers:
## What I did

1. remove the existing icons build script
2. replace it with a script that copies over patternfly icons (mostly fontawesome 5)

## Testing Instructions

1. [DP](https://deploy-preview-2502--patternfly-elements.netlify.app/components/icon/demo/sets/)
2. copy locally into RHDS and check footer, navigation-secondary

## Notes to Reviewers

1. ~~pretty sure this is a breaking change. I think we can avoid a major release by restoring the old build script and running the new one on top of it, then any icons of the same file name can be considered 'fixes' without breaking the path they're stored at~~ breakage has been avoided


scripts/icons.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+(async function() {
+  // 1: build deprecated fa 6 icons
+  await build('fab', await import('@fortawesome/free-brands-svg-icons').then(x => x.fab));
+  await build('far', await import('@fortawesome/free-regular-svg-icons').then(x => x.far));
+  await build('fas', await import('@fortawesome/free-solid-svg-icons').then(x => x.fas));
+  // 2: overwrite with icons from `@patternfly/icons`
+  const $$ = $({ cwd: process.cwd() });
+  await $$`cp -R node_modules/@patternfly/icons/ elements/pf-icon/`;
```
heyMP, line 46:
```suggestion
  await $$`cp -R node_modules/@patternfly/icons/ elements/pf-icon/icons/`;
```

bennypowers: when i run this as-is, it works as expected, did you get fab far fas and patternfly in elements/pf-icon?

heyMP: Oh interesting.  I was thinking that we were overriding the existing directory in `pfe-icon/icons`?  I think if we go from 

```
pf-icons
  icons
    fab
    far
    patternfly
```

to 

```
pf-icons
  fab
  far
  patternfly
```

it would be a breaking change, right?

zeroedin:
Should these generated files be in the`.gitignore`

Screenshot 2023-06-06 at 10 00 15 AM

scripts/icons.ts
bennypowers: image
This is what I get when I build using the code as-is, do you have something different? I get this result on linux and macos

bennypowers:
@zeroedin I can't reproduce that - i tried busting the wireit cache, git clean, npm ci, and `npx ts-node --esm ./scripts/icons.ts`, and everything copies into `elements/pf-icon/icons`

zeroedin:
> @zeroedin I can't reproduce that - i tried busting the wireit cache, git clean, npm ci, and `npx ts-node --esm ./scripts/icons.ts`, and everything copies into `elements/pf-icon/icons`

I'm running:

```
git clean -fdx
nvm use
npm ci
npm run start
```

bennypowers:
@zeroedin I ran that and it works as expected here... i'll try cloning into /tmp

bennypowers:
I cloned into /tmp/ and ran the build, and everything came out as expected. fedora 38

scripts/icons.ts
heyMP: I actually get this 🤔:
Screenshot 2023-06-06 at 11 16 34 AM

heyMP:
> I cloned into /tmp/ and ran the build, and everything came out as expected. fedora 38

I'll try with a fresh copy

zeroedin:
> I cloned into /tmp/ and ran the build, and everything came out as expected. fedora 38

When I do the same here is my output:

https://gist.github.com/zeroedin/fe15286ee245e0ba92318ea9012b34fa

bennypowers:
ok i think this is a POSIX/cp -R issue on macos

bennypowers:
i'll refactor to use node fs apis

heyMP:
TIL: https://dev.to/ackshaey/macos-vs-linux-the-cp-command-will-trip-you-up-2p00

zeroedin:
> Try again, this should be portable to cupertinux now.

Icon location look good to me now and no extra "added" file changes after build.

zeroedin:
Ran a build and copied icons files and js over to RHDS, things look ok downstream from what I can tell as well.

zeroedin (approved):
Letters, Glyphs, Trademarks, Monograms

heyMP:
Fixed for me!  Let's go! 🚀

heyMP (approved):
Let's go 🚀

Issue #2503, opened 2023-05-24
Utilize Intl.RelativeTimeFormat for relative time


KamiQuasi:
# Component name: <pfe-timestamp>

> Name/Description of the component being updated

## Design specification updates
The [getTimeRelative function](https://github.com/patternfly/patternfly-elements/blob/6cb9318c6832a748b3e74ed74d4967a277637f62/elements/pf-timestamp/pf-timestamp.ts#L80) could leverage [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) to calculate the relative times.

## Accessibility updates

No changes

## Interaction updates

No changes

## Analytics updates

No changes

bennypowers:
Already implemented - the timestamp controller's `#getTimeRelative()` method already uses `Intl.RelativeTimeFormat` (introduced in commit a81bcb133).

PR #2504, opened 2023-05-24
refactor(timestamp): use Intl.RelativeTimeFormatUnit

KamiQuasi:
## What I did

1. Updated the relative time calculation to use [Intl.RelativeTimeFormatUnit](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat)
2. Issue: https://github.com/patternfly/patternfly-elements/issues/2503

## Testing Instructions

1. Check that the various timespan options are unaffected

## Notes to Reviewers

1. This will change the text for future times from being "XX [units] until" to "in XX [units]"
3. Past times should still remain "XX [units] ago"

### DIFF
--- a/.changeset/short-tools-lay.md
+++ b/.changeset/short-tools-lay.md
@@ -2,4 +2,4 @@
 "@patternfly/elements": minor
 ---
 
-Changed the `getTimeRelative` function to use `Intl.RelativeTimeFormat` and added a few extra demo samples
+`<pf-timestamp>`: improved performance by using browser-standard features to calculate relative time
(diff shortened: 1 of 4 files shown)


bennypowers:
Needs a changeset, and can you add some regression tests?

.changeset/short-tools-lay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@patternfly/elements": minor
+---
+
+Changed the `getTimeRelative` function to use `Intl.RelativeTimeFormat` and added a few extra demo samples
```
bennypowers, line 5:
```suggestion
`<pf-timestamp>`: improved performance by using browser-standard features to calculate relative time
```

bennypowers (approved):

PR #2505, opened 2023-05-25
feat: ssr-friendly slot-controller

bennypowers:
:point_right: DRAFT PR DO NOT MERGE :point_left: 
This approach is experimental and requires discussion before pursuing. This PR is a PoC only. However there is a lot to get excited about. for example, we could load a `/elements/pf-icon/pf-icon-ssr.js` at ssr time which, instead of setting up a bunch of client side js to fetch and render the icon, would just dump the contents of the svg into shadow root with `await readFile`, at SSR time.

Read more: https://lit.dev/docs/ssr/overview/

## TODO:
use export conditions instead of `isServer`

image
image

## What I did

1. implement lit-ssr in slot-controller via a new `has-slotted` host attr

## Testing Instructions

1. disable javascript and visit [PR Preview Card SSR demo](https://deploy-preview-2505--patternfly-elements.netlify.app/components/card/demo/ssr/) (webkit or chromium until [firefox implements](https://bugzilla.mozilla.org/show_bug.cgi?id=1712140))

## Notes to Reviewers

**Problem**: Lit-SSR can't access the children, whereas SlotController relies on knowledge about the children. 
**Proposal**: read `has-slotted="a,b,c"` attr on SlotController hosts. This has limitations: 
1. content author has to add attrs
2. we can't use `hasSlotted` in connectedCallback, only in `render` (lit may relax this limitation in the future)

### DIFF
--- a/core/pfe-core/controllers/slot-controller.ts
+++ b/core/pfe-core/controllers/slot-controller.ts
@@ -1,4 +1,5 @@
 import type { ReactiveController, ReactiveElement } from 'lit';
+import { isServer } from 'lit';
 
 import { bound } from '../decorators/bound.js';
 import { Logger } from './logger.js';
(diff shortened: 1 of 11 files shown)


bennypowers:
closing in favour of #2893

PR #2509, opened 2023-06-01
fix(tabs): vertical alignment of slotted icons

zeroedin:
## What I did

1. Fixed vertical alignment of slotted icons
2. Improved documentation: 
    -  Fixed missing  slot attribute on `<pf-icon>` 
    -  Set correct icon sizes to match PFv4.

## Testing Instructions

1.

## Notes to Reviewers

CSS tweak and docs changes fixes the following issues:

image of pf-icon missing slot attribute

`pf-icon` missing slot attribute

image of pf-icon set to small size in tabs

`pf-icons` were set to small in demos change this to md. 

The improvements to shadow CSS should ensure the slotted icons are always vertically centered with the tab text.  The icon CSS part in the BaseTab.css should have most likely never been set to display: block;  So, instead of changing that and introducing a potential breaking change, this update applies `display: flex !important` in pf-tab.css.  Also uses flex gap instead of a margin right reducing the need to set last child to a margin-right of 0.  A similar update should happen downstream in RHDS.


bennypowers (approved):
nice clean PR

Issue #2508, opened 2023-06-01
[1:1]: pf-progress


brianferry:
# Component name: pf-progress

> A progress bar informs users about the completion status of an ongoing process or task.

PR #2511, opened 2023-06-06
feat(progress): add pf-progress

brianferry:
## pf-progress
Implemented Patternfly's [Progress](https://www.patternfly.org/v4/components/progress) as web component.

## Related issues
(https://github.com/patternfly/patternfly-elements/issues/2508)

## Preview

[Link](https://deploy-preview-2511--patternfly-elements.netlify.app/components/progress/)


elements/pf-progress/pf-progress.css
```diff
@@ -0,0 +1,152 @@
+.container {
+  --pf-c-progress--GridGap: var(--pf-global--spacer--md, 1rem);
```
bennypowers, line 2: this and others are listed in the jsdoc, but if we set them here on the container, they won't be user-configurable, so they should probably be fallbacks instead, unless patternfly's implementation works like this

elements/pf-progress/docs/pf-progress.md
```diff
@@ -0,0 +1,17 @@
+{% renderOverview %}
+    <pf-progress title="Default" value="33"></pf-progress>
+{% endrenderOverview %}
+
+{% band header="Usage" %}{% endband %}
```
bennypowers, line 5: should have an example here

elements/pf-progress/pf-progress.ts
```diff
@@ -0,0 +1,196 @@
...
+    this.#updateAccessibility();
+  }
+
+  render() {
+    const { size, measureLocation, variant, value, title } = this;
+    const icon = variant && ICONS.get(variant)?.icon;
+    const singleLine = title.length === 0 ? 'singleline' : '';
+
+    return html`
+      <div part="container" class="container ${classMap({ [size]: !!size, [measureLocation]: !!measureLocation, [variant]: !!variant, [singleLine]: !!singleLine })}">
```
bennypowers, line 171: list the parts in jsdoc, or remove them

bennypowers (changes requested):
most pressing change is to put remove the private updateAccessibility method and put that stuff in the template instead

it might we good to set a private property in `willUpdate` for that purpose

bennypowers:
@nikkimk can you give this a usability review, please?

brianferry:
@bennypowers - Wanted to see what you thought about the progress element approach I've taken here, need to move the accessibility upgrades into the template still and clean up the CSS but I think accessibility wise and property wise it looks cleaner with this approach.

https://deploy-preview-2511--patternfly-elements.netlify.app/components/progress/demo/kitchen-sink/

Linked to the kitchen sink demo

elements/pf-progress/pf-progress.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  #calculatedPercentage = 0;
+
+  #logger = new Logger(this);
+
+  @property({ reflect: true, type: Number })
+    value = 0;
+
+  @property()
+    title = '';
```
zeroedin, line 135: Should `title` be a slotted element instead?  

I realize this is most likely mocking PFv4 API, but the title attribute is a [global html attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) so this feels like it is overloading that with a new implementation.

elements/pf-progress/demo/accessibility.html
```diff
@@ -0,0 +1,23 @@
...
+<pf-progress id="danger-labelled-inside" value="33" variant="danger" measure-location="inside"></pf-progress>
+
+<h2> Danger Titled </h2>
+<pf-progress title="Danger Titled" value="33" variant="danger"></pf-progress>
+
+<h2> With title </h2>
+<pf-progress title="attribute label" value="5" min="-10" max="10"></pf-progress>
+
+<h2> Fancy label title </h2>
+<label for="zoidberg">whoop whoop whoop 🦀</label>
```
bennypowers, line 22: :trophy:

elements/pf-progress/pf-progress.ts
```diff
@@ -120,83 +122,102 @@ const ICONS = new Map(Object.entries({
...
+  connectedCallback(): void {
     super.connectedCallback();
+    this.#calculatePercentage();
     this.#updateAccessibility();
   }
 
+  #updateAccessibility() {
+    const { value, title, ariaLabel, max, progressElement } = this;
+    if (progressElement) {
+      progressElement.setAttribute('aria-valuemin', '0');
```
bennypowers, line 181: this *must* be handled in the lit template, not through imperative code

elements/pf-progress/pf-progress.ts
```diff
@@ -120,83 +122,102 @@ const ICONS = new Map(Object.entries({
...
-    } else if (this.bar !== null) {
-      this.bar.setAttribute('aria-label', this.label);
+  #calculatePercentage() {
+    const { value, max, min } = this;
+    if (value === null ) {
+      this.#logger.warn(`A progress element requires a value attribute.`);
+      return;
     }
+    const percentage = (value - min) / (max - min);
+    this.#calculatedPercentage = Math.round(percentage * 100);
```
bennypowers, line 169: let's either move all this to scoped consts in the render method, or have this private method return it's value rather than mutating private state.

the logger call, since it's a side effect should move to `willUpdate`

bennypowers (changes requested):

elements/pf-progress/demo/pf-progress.html
bennypowers, line 1: we want to move to a model where demo files contain fewer element instances. we can still have demo files with multiple instances in them, but for the main demo file, please include only a single canonical example.

elements/pf-progress/demo/pf-progress.js
bennypowers, line 1: let's just inline this in the demo files, it'll be easier to read

elements/pf-progress/pf-progress.ts
```diff
@@ -0,0 +1,228 @@
...
+ * @cssprop {<length>} --pf-c-progress--m-lg__bar--Height
+ *          Height of the progress bar when the size is large.
+ *          {@default `1.5rem`}
+ *
+ */
+@customElement('pf-progress')
+export class PfProgress extends LitElement {
+  static readonly styles = [styles];
+
+  static readonly formAssociated = true;
```
bennypowers, line 125: we'll need to attachInternals, and whenever the value changes, call `setValue()` to report back the form value

elements/pf-progress/pf-progress.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  #calculatedPercentage = 0;
+
+  #logger = new Logger(this);
+
+  @property({ reflect: true, type: Number })
```
bennypowers, line 131: Need docblocks for the public attrs

elements/pf-progress/pf-progress.ts
```diff
@@ -0,0 +1,228 @@
...
+          
+          ${measureLocation !== 'none' ? html`<div class="status">
+            ${html`${!this.#showInsideStatus() ? `${this.#calculatedPercentage}%` : ''}`}
+            <pf-icon set="fas"
+                  size="md"
+                  ?hidden="${!icon}"
+                  icon="${ifDefined(icon)}"></pf-icon>
+          </div>` : ''}
+            
+          <progress tabindex="0" max="100" value="${this.#calculatedPercentage}"></progress>
```
bennypowers, line 215: @nikkimk ptal at this tabindex

elements/pf-progress/pf-progress.ts
```diff
@@ -124,70 +104,52 @@ export class PfProgress extends LitElement {
...
-  }
-
-  connectedCallback(): void {
-    super.connectedCallback();
-    this.#calculatePercentage();
-    this.#updateAccessibility();
+    if (percentage > 100) {
+      return 100;
+    }
+    return percentage;
```
bennypowers, line 146:
```suggestion
    return Math.min(percentage, 100);
```
small nit

elements/pf-progress/pf-progress.ts
```diff
@@ -124,70 +104,52 @@ export class PfProgress extends LitElement {
...
-        progressElement.setAttribute('aria-label', `${title ? title : ariaLabel}`);
-      }
-      if (max != null) {
-        progressElement.setAttribute('aria-valuemax', `${max}`);
-      }
-      if (value) {
-        progressElement.setAttribute('aria-valuenow', `${value}`);
-      }
+  willUpdate(_changedProperties: PropertyValues<this>) {
+    super.willUpdate(_changedProperties);
```
bennypowers, line 150: not needed https://lit.dev/docs/components/lifecycle/#willupdate

elements/pf-progress/pf-progress.ts
brianferry: Removed the tabindex and the form association after chatting with @hellogreg, we will add it as something we can add in later but outside of https://www.magentaa11y.com/checklist-web/progress/ it does not appear that others recommend treating progress like an input element for a screen reader.

elements/pf-progress/pf-progress.css
```diff
@@ -0,0 +1,181 @@
...
+  --pf-c-progress--m-outside__measure--FontSize: var(--pf-global--FontSize--sm, 0.875rem);
+  --pf-c-progress--m-sm__bar--Height: var(--pf-global--spacer--sm, 0.5rem);
+  --pf-c-progress--m-sm__description--FontSize: var(--pf-global--FontSize--sm, 0.875rem);
+  --pf-c-progress--m-lg__bar--Height: var(--pf-global--spacer--lg, 1.5rem);
+  display: grid;
+  align-items: end;
+  grid-gap: var(--pf-c-progress--GridGap);
+  grid-template-columns: auto auto;
+  grid-template-rows: 1fr auto;
+}
```
bennypowers, line 31: please add a newline after each rule

elements/pf-progress/pf-progress.ts
```diff
@@ -0,0 +1,189 @@
...
+    measureLocation: '' | 'outside' | 'inside' | 'none' = '';
+
+  /** Variant of the progress bar */
+  @property()
+    variant: '' | 'success' | 'danger' | 'warning' = '';
+
+  get #calculatedPercentage(): number {
+    const { value, min, max } = this;
+    const percentage = Math.round((value - min) / (max - min) * 100);
+    if (isNaN(percentage) || percentage < 0) {
```
bennypowers, line 135:
```suggestion
    if (Number.isNaN(percentage) || percentage < 0) {
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN

elements/pf-progress/pf-progress.ts
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  render() {
+    const { size, measureLocation, variant, title } = this;
+    const icon = variant && ICONS.get(variant)?.icon;
+    const singleLine = title.length === 0 ? 'singleline' : '';
+
+    return html`
+      <div
+        class="container ${classMap({ [size]: !!size, [measureLocation]: !!measureLocation, [variant]: !!variant, [singleLine]: !!singleLine })}">
```
bennypowers, line 148:
```suggestion
        class="container ${classMap({ [size]: !!size, [measureLocation]: !!measureLocation, [variant]: !!variant, singleline })}">
```

elements/pf-progress/pf-progress.ts
```diff
@@ -0,0 +1,189 @@
...
+    if (isNaN(percentage) || percentage < 0) {
+      return 0;
+    }
+    return Math.min(percentage, 100);
+  }
+
+  render() {
+    const { size, measureLocation, variant, title } = this;
+    const icon = variant && ICONS.get(variant)?.icon;
+    const singleLine = title.length === 0 ? 'singleline' : '';
```
bennypowers, line 144:
```suggestion
    const singleline = title.length === 0;
```

elements/pf-progress/pf-progress.ts
```diff
@@ -0,0 +1,189 @@
...
+ */
+@customElement('pf-progress')
+export class PfProgress extends LitElement {
+  static readonly styles = [styles];
+
+  static readonly formAssociated = true;
+
+  /** Represents the value of the progress bar */
+  @property({ reflect: true, type: Number })
+    value = 0;
```
bennypowers, line 106: since this is formAssociated, we'll need to setValue on the internals object when the value changes. Please add a test for form submission

elements/pf-progress/test/pf-progress.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    const max = Math.floor(Math.random() * 100);
+    const value = Math.floor(Math.random() * (max));
+
+    const element = await fixture<PfProgress>(html`
+      <pf-progress value="${value}" max="${max}">
+      </pf-progress>
+    `);
+    expect(element.value).to.equal(value);
+    expect(element.max).to.equal(max);
+  });
```
bennypowers, line 51:
```suggestion
  });
  
  describe('in a form', function() {
    let element: PfProgress;
    let form: HTMLFormElment;
    beforeEach(async function() {
      form = await createFixture(html`
        <form @submit=${e => e.preventDefault()}>
          <label for="progress">Progress</label>
          <pf-progress id="progress" name="progress" value="50"></pf-progress>
        </form>
      `);
      element = form.querySelector('pf-progress');
    });
    
    describe('submitting the form', function() {
      let submitEvent: SubmitEvent;
      beforeEach(async function() {
        setTimeout(() => form.requestSubmit());
        submitEvent = await oneEvent(form, 'submit');
      });
      it('sets the value on the form', function() {
        expect(form.elements.progress.value).to.equal(50);
      });
    });
  });
```

or some such. This will require some thought, since we coerce the min/value/max into a percentage.

bennypowers (changes requested):

elements/pf-progress/pf-progress.ts
```diff
@@ -141,11 +141,11 @@ export class PfProgress extends LitElement {
   render() {
     const { size, measureLocation, variant, title } = this;
     const icon = variant && ICONS.get(variant)?.icon;
-    const singleline = title.length === 0;
+    const singleLine = title.length === 0;
 
     return html`
       <div
-        class="container ${classMap({ [size]: !!size, [measureLocation]: !!measureLocation, [variant]: !!variant, [singleLine]: !!singleLine })}">
+        class="container ${classMap({ [size]: !!size, [measureLocation]: !!measureLocation, [variant]: !!variant, ['singleLine']: !!singleLine })}">
```
bennypowers, line 148:
```suggestion
        class="container ${classMap({ [size]: !!size, [measureLocation]: !!measureLocation, [variant]: !!variant, singleLine })}">
```

elements/pf-progress/pf-progress.ts
brianferry: Trying to think through whether we need it to be form associated anymore since we are not having the progress bar be accessible.  It would still have a value but be more like an image than an actual input.

bennypowers: We'd still want the host element to submit it's value with the form though. Setting the value property on the host object isn't enough, we need to call setValue on the internals object or the formdata won't read the progress value

bennypowers:
Some research we did earlier today: progress elements don't submit with the form https://codepen.io/bennyp/pen/abQxWeQ

elements/pf-progress/pf-progress.ts
brianferry: Discussed today during office hours, the native `progress` element itself does not report its value to the form element.  Further, it appears that forms only submit a control that is outlined in the spec https://www.w3.org/TR/html401/interact/forms.html#form-controls which can be a input, text area, menu, and/or `hidden` | `generic` options set through `input` | `object` respectively. As such, we will not include form association for progress element as this time.

elements/pf-progress/pf-progress.ts
```diff
@@ -129,6 +128,12 @@ export class PfProgress extends LitElement {
   @property()
     variant: '' | 'success' | 'danger' | 'warning' = '';
 
+  #mo = new MutationObserver(() => this.#onMutation());
```
bennypowers, line 131: We don't need a mutation observer for this we can use `@property` and `willUpdate`

elements/pf-progress/pf-progress.ts
```diff
@@ -6,6 +6,7 @@ import { ifDefined } from 'lit/directives/if-defined.js';
 import { styleMap } from 'lit/directives/style-map.js';
 
 import styles from './pf-progress.css';
+import { InternalsController } from '@patternfly/pfe-core/controllers/internals-controller.js';
```
bennypowers, line 9: This thing is pretty experimental, might be better to roll our own here, for now

bennypowers (changes requested):
Thanks again Brian for your research on form participation for this one. We're almost there

bennypowers (approved):
LGTM

@brianferry @eyevana please validate the changes to the test files

brianferry:
Tooltip and progress look good to me 👍🏼

elements/pf-popover/pf-popover.ts
```diff
@@ -360,7 +358,10 @@ export class PfPopover extends LitElement {
       <div id="container"
            style="${styleMap(styles)}"
            class="${classMap({ [anchor]: !!anchor, [alignment]: !!alignment })}">
-        <slot id="trigger" @keydown=${this.onKeydown} @click=${this.toggle}></slot>
+        <slot id="trigger"
+              @slotchange="${this.#triggerChanged}"
```
eyevana, line 362: @bennypowers `triggerChanged` is responsible for adding and removing event listeners to the element matching the id passed through the optional `trigger` property. Why do we need to call in the `slotchange` event handler for this slot?

elements/pf-popover/pf-popover.ts
```diff
@@ -393,6 +394,23 @@ export class PfPopover extends LitElement {
...
+    const root = this.getRootNode() as Document | ShadowRoot;
+    return !this.trigger ? null : root.getElementById(this.trigger);
+  }
+
+
+  #triggerChanged() {
+    const oldReferenceTrigger = this.#referenceTrigger;
+    this.#referenceTrigger = this.#getReferenceTrigger();
+    if (oldReferenceTrigger !== this.#referenceTrigger) {
+      oldReferenceTrigger?.removeEventListener('click', this.show);
```
eyevana, line 407: @bennypowers this is my bad...should we use `toggle` here instead of `show`?

elements/pf-popover/pf-popover.ts
bennypowers: we want to use either the slotted trigger or the assigned one. and also, if a node gets removed from the slot or added back in, we need to hook up the listeners.

bennypowers: we wouldn't want them to stay on the slot though if there's an assigned trigger :thinking: 

we might be able to handle that logic in the handler though

elements/pf-popover/pf-popover.ts
bennypowers: yeah maybe? I'll take your word for it :grin:

elements/pf-popover/test/pf-popover.spec.ts
```diff
@@ -126,38 +158,71 @@ describe('<pf-popover>', function() {
...
+        describe('clicking the first button', function() {
+          beforeEach(async function() {
+            await element.updateComplete;
+            await clickElementCenter(btn1);
+            await element.updateComplete;
+          });
+          it('remains closed', expectClose);
+        });
+        describe('clicking the sibling button', function() {
+          beforeEach(async function() {
```
eyevana, line 209: This test is failing after I made the switch from `show` to `toggle` on the event handlers we attached to the trigger. The sequence of events looks correct to me (btw, very cool how you wrote these tests!). If I change this hook to `before` instead of `beforeEach`, all tests pass 🤔 but I'm not sure why!?

PR #2512, opened 2023-06-08
fix(tools): dev server aliased element redirects

zeroedin:
## What I did

1. Corrects aliased elements redirects in dev-server

## Testing Instructions

1. Dev Server locally add the configuration from the `.pfe.config.json` below:

## Notes to Reviewers

Given a `.pfe.conf.json`  with the following contents:

```
{
  "renderTitleInOverview": false,
  "aliases": {
    "pf-modal": "Marvelous Modals",
    "pf-tabs": "Terrific Tabs",
    "pf-clipboard-copy": "Making Copies",
    "pf-tile": "Tantalizing Tiles"
  }
}
```
A request  to pf-modals demo.css would produce a URL in the dev-server as: 

http://localhost:8000/components/magnificent-modal/demo/demo.css

Using deslugify we take the `magnificient-modal` slug and check for an alias and return it, If there is no alias we simply return the original path requested ensuring it has the correct tag prefix.

http://localhost:8000/components/magnificent-modal/demo/demo.css
Redirects to => 
http://localhost:8000/components/pf-modal/demo/demo.css

### DIFF
--- a/.changeset/fuzzy-walls-walk.md
+++ b/.changeset/fuzzy-walls-walk.md
@@ -1,0 +1,5 @@
+---
+"@patternfly/pfe-tools": patch
+---
+
+`dev-server`: Corrects aliased elements redirects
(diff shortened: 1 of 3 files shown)


nikkimk (approved):
lobsters get terribly mean

bennypowers (approved):

eyevana (approved):

Issue #2513, opened 2023-06-10
[Bug] pf-popover outsideClick event trigger for all popovers present on the page.


anujsingla:
[Bug] pf-popover outsideClick event trigger for all popovers present on the page.

https://patternflyelements.org/components/popover/

PR #2514, opened 2023-06-10
fix(popover): add outsideClick event on popover show

anujsingla:
fix(popover): add outsideClick event on popover show

We were adding outsideClick event in the connectedCallback lifecycle method. It adds for all popovers present on the page. It was not removed when we closed the popover.
Now I have changed the logic. We will add an outsideClick event when we show the popover and remove it on the hide popover.

fixes - https://github.com/patternfly/patternfly-elements/issues/2513


bennypowers (changes requested):
This needs a changeset, otherwise looks good

bennypowers (approved):

PR #2516, opened 2023-06-13
fix(tools): use markdown headings on docspages

bennypowers:
Allows downstream to use markdown-it plugins to manipulate output e.g. by adding hash links to headers

Unblocks https://github.com/RedHat-UX/red-hat-design-system/pull/1051

## What I did

1. in DocsPage templates, print headings as markdown instead of as HTML

## Testing Instructions

1. compare DP to https://patternflyelements.org/, https://patternflyelements.org/get-started/, and https://patternflyelements.org/components/avatar/

### DIFF
--- a/.changeset/fuzzy-walls-walk.md
+++ b/.changeset/fuzzy-walls-walk.md
@@ -1,0 +1,5 @@
+---
+"@patternfly/pfe-tools": patch
+---
+
+`dev-server`: Corrects aliased elements redirects
--- a/tools/pfe-tools/config.ts
+++ b/tools/pfe-tools/config.ts
@@ -86,7 +86,11 @@ function getSlugsMap(rootDir: string) {
   return slugsConfigMap.get(rootDir)!;
 }
 
+/**
+ * Returns the prefixed custom element name for a given slug
+ */
 export function deslugify(slug: string, rootDir = process.cwd()): string {
   const { slugs, config } = getSlugsMap(rootDir);
-  return slugs.get(slug) ?? `${config.tagPrefix}-${slug}`;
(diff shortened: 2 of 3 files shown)


tools/pfe-tools/11ty/DocsPage.ts
```diff
@@ -85,6 +85,8 @@ export class DocsPage implements DocsPageRenderer {
     this.templates.addFilter('log', DocsPage.#log);
     this.templates.addFilter('type', DocsPage.#type);
     this.templates.addFilter('innerMD', DocsPage.#innerMD);
+    this.templates.addFilter('mdHeading', (header, length = 2) =>
+      DocsPage.#innerMD(`${Array.from({ length }, () => '#').join('')} ${header}`));
```
heyMP, line 89: Super minor suggestion, not important.

```suggestion
    this.templates.addFilter('mdHeading', (header, level = 2) =>
      DocsPage.#innerMD(`${Array.from({ level }, () => '#').join('')} ${header}`));
```

heyMP:
Deploy Preview isn't working.  Looks like there is an error associated with floating-ui.

tools/pfe-tools/11ty/DocsPage.ts
bennypowers: length is a keyword here. Array.from can construct an array from an object with a length property

bennypowers:
The error hasto do jspm's cdn provider cc @guybedford

tools/pfe-tools/11ty/DocsPage.ts
zeroedin: 👀 🤯 very cool also TIL.

PR #2523, opened 2023-06-23
feat(tabs): add detached panels

zeroedin:
## THIS IS EXPERIMENTAL

Testing ideas coming from conversation here: https://github.com/RedHat-UX/red-hat-design-system/pull/995 

## What I did

1. Added `detached` property to BaseTabs, allows panels to be detached from the tab set.

## Testing Instructions

1.  View deploy preview  [tabs detached demo](https://deploy-preview-2523--patternfly-elements.netlify.app/components/tabs/demo/detached/)

## Notes to Reviewers

1.  Test for accessibility gotchas, tab order etc.  I'm thinking this could cause issue with expected tab order and could be abused.


bennypowers:
@zeroedin this is more what I had in mind - tabs that live outside the pf-tabs container. WDYT?

bennypowers:
can we delete this branch?

do we want to remove the tab slot from pf-tabs while we're at it?

PR #2525, opened 2023-06-27
fix(core): typescript import from floating-ui library

bennypowers:
## What I did

1. fix the import from floating ui


heyMP (approved):
Fix it 🛠️!

PR #2533, opened 2023-07-05
fix: generate d.ts for react wrappers

bennypowers:
## What I did

1. generate .d.ts for react wrappers
2. bump versions in package lock so that `npm i @patternfly/elements` gets the new hotness

### DIFF
--- a/tools/pfe-tools/react/generate-wrappers.ts
+++ b/tools/pfe-tools/react/generate-wrappers.ts
@@ -1,5 +1,5 @@
 import type * as CEM from 'custom-elements-manifest';
-import javascript from 'dedent';
+import dedent from 'dedent';
 import { dirname, join, relative } from 'node:path';
 import { fileURLToPath } from 'node:url';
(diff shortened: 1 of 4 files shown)


tools/pfe-tools/react/generate-wrappers.ts
```diff
@@ -40,26 +40,39 @@ async function writeReactWrapper(
   if (!tagName) {
     throw new NonCriticalError(`declaration does not have a tag name: ${decl.name}`);
   } else {
+    const javascript = dedent;
+    const typescript = dedent;
```
brianferry, line 44: Do lines 43 / 44 do the same thing or are they different?

bennypowers: they're just aliases for the sake of IDE syntax highlighting

brianferry (approved):
Looks good, ran build and checked files, seeing the .js and .d.ts files

Issue #2537, opened 2023-07-06
[bug] pf-accordion collapses last panel when internal checkbox is checked


brianferry:
## Description of the issue

Noticed this while converting the product trials preview panel from the original `pfe-accordion` to `pf-accordion`.  When you have a checkbox in the panel it will close the first or next panel in the accordion.  This is because when we do a check for the collapse function we do not cancel -1 indexes which indicate that it comes from a non-pf-accordion element. 

### Impacted component(s)

- 

### Steps to reproduce

1. Go to [codepen](https://codepen.io/OhHeyBrian/pen/vYQeLoX)
2. Click on the checkbox in the accordion panel
3. See that it closes the panel unexpectedly 

### Expected behavior

1. Clicking an internal element that is not the accordion element should not result in the panel being collapsed.

bennypowers:
Closed by #2538

PR #2538, opened 2023-07-06
fix(accordion): panel checkbox collapsing panel on check

brianferry:
#2537 

## What I did

1.  Added an index check to the `collapse` function so that if the element that called the `change` event is not an `accordion-*` element it will not cause the panel to close.

### DIFF
--- a/.changeset/neat-colts-jam.md
+++ b/.changeset/neat-colts-jam.md
@@ -2,4 +2,4 @@
 "@patternfly/elements": patch
 ---
 
-`pf-accordion`: fixed issue related to panel collapsing when an internal element is clicked.  Removing deprecated code now handled by RovingTabIndex controller.
+`<pf-accordion>`: fixed bug which would collapse a panel if it contained a checkbox which got toggled.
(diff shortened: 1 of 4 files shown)


bennypowers:
Can we fix this with an instanceof check on the change event?

brianferry:
> Can we fix this with an instanceof check on the change event?

Good point, that would be simpler.  Went ahead and added the check & removed some of the code that was deprecated from using RTI for accordion.

bennypowers:
@brianferry take a look at [f47a340](https://github.com/patternfly/patternfly-elements/pull/2538/commits/f47a3403fa2fcc45643fdca7d6de6629dcbe75eb): i had in mind to check to see if the event was an instance of our change event.

I added a test for pf-switch and moved the setup code for the tests into the template and beforeEach blocks.

bennypowers (approved):

.changeset/neat-colts-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@patternfly/elements": patch
+---
+
+`pf-accordion`: fixed issue related to panel collapsing when an internal element is clicked.  Removing deprecated code now handled by RovingTabIndex controller.
```
bennypowers, line 5:
```suggestion
`<pf-accordion>`: fixed bug which would collapse a panel if it contained a checkbox which got toggled.
```

Issue #2496, opened 2023-05-08
roving-tabindex-controller listeners should only handle focusable item events.


nikkimk:
## Description of the issue

`roving-tabindex-controller` currently listens for keydown events on the items' common ancestor container, which works in cases where the items share a common ancestor, like for tabs or menus, but for accordions, the listener is also capturing panel key events.

### Impacted component(s)

- `pf-accordion`

### Steps to reproduce

1. Go to the [Fixed Panel demo](https://deploy-preview-2493--patternfly-elements.netlify.app/components/accordion/demo/)
2. Expand **Item three**.
3. Tab to **Focus on me!** 
4. Use arrow keys. 
5. The focus changes to  another **Fixed panel** heading. 

### Expected behavior

The listener should ignore arrow keys from inside a panel.

## Related issues

- [#2493](https://github.com/patternfly/patternfly-elements/pull/2493)

bennypowers:
does #2493 address this?

PR #2555, opened 2023-07-13
bug: accordion and roving tabindex controller

nikkimk:
## What I did

1. Determined that #2496 was an issue with `BaseAccordion`'s use of `roving-tabindex-controller` rather than the controller itself.
2. Modified `BaseAccordion`'s `get #activeHeader` so that if no active header is found, the active header is returned as `undefined`.
3. Fixed issues with testing. (When focus is in the accordion panel and tab is pressed, focus should go to the body.)

## Testing Instructions

1. Go to the [Fixed Panel demo in DP](https://deploy-preview-2555--patternfly-elements.netlify.app/components/accordion/demo/)
2. Expand Item three.
4. Tab to Focus on me!
5. Use arrow keys.
6. The focus **_should not_** change.

### DIFF
--- a/.changeset/late-actors-give.md
+++ b/.changeset/late-actors-give.md
@@ -1,0 +1,5 @@
+---
+"@patternfly/elements": patch
+---
+
+`pf-accordion`: fixed issue with focus event triggering incorrect tabindex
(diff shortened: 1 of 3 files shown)


zeroedin (approved):
lgtm

PR #2558, opened 2023-07-18
fix(card): make slotted styles !important

bennypowers:
## What I did

1. make slotted styles in card `!important`

## Testing Instructions

1. try with this page:

```html
<style>
:is(h1, h2, h3, h4, h5, h6) {
  font-family: var(--rh-font-family-heading);
  margin-block: var(--rh-space-xl);
  line-height: 1;
}
</style>
<pf-card>
  <h3 slot=header>Haha broke your styles!</h3>
</pf-card>
```

## Notes to Reviewers

1. `!important` isn't great but these are design system tokens so I feel it's justified. You could argue that the margin values should not be important, but I think in this case they're specified so it's ok.

### DIFF
--- a/.changeset/pf-card-heading-specificity.md
+++ b/.changeset/pf-card-heading-specificity.md
@@ -1,0 +1,4 @@
+---
+"@patternfly/elements": patch
+---
+`<pf-card>`: ensure that slotted content's layout always follows design specs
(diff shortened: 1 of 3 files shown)


brianferry (approved):
Looks good to me, I think for other elements we might want to do the same for slotted element styles.

PR #2559, opened 2023-07-18
feat: move converter to pfe core

eyevana:
## What I did

1. Moved the boolean string converter to pfe-core so that `rh-timestamp` can reference as well

## Testing Instructions

1.

## Notes to Reviewers

1.


bennypowers:
let's give a think about whether it's better to just reimplement that converter in rhds, at least until we have (or plan to have) more uses for it

eyevana:
Okie dokie. I'll close this one then.. Thank you.

PR #2563, opened 2023-07-20
feat(core): add listbox support for roving-tabindex-controller

nikkimk:
## What I did

1. Added directional support to `roving-tabindex-controller`.
2. Added a `listbox-controller`.
3. Added listbox components in order to test the support and demo how we can create an accessible listbox component: `pf-listbox`, `pf-listbox-group`, and `pf-listbox-option`.

## Testing Instructions

1. See [default demo](https://deploy-preview-2563--patternfly-elements.netlify.app/components/listbox/demo/)
2. See [orientation demo](https://deploy-preview-2563--patternfly-elements.netlify.app/components/listbox/demo/orientation/)
4. See [multiple selections demo](https://deploy-preview-2563--patternfly-elements.netlify.app/components/listbox/demo/multiple-selections/)
5. See [grouped options demo](https://deploy-preview-2563--patternfly-elements.netlify.app/components/listbox/demo/grouped-options/)
6. See [filter modes demo](https://deploy-preview-2563--patternfly-elements.netlify.app/components/listbox/demo/filter-modes/) (combobox support)

## Notes to Reviewers

**_This is in draft with a temporary listbox for testing._** We may be able to either use it as a reference or adapt it for a future listbox component.

Components with `role=listbox` can be used as standalone elements, as part of a combobox. See the following references:
1. [Toggle and Items Accessibility](https://docs.google.com/document/d/1lvAUBaBNiSiHDkWh3jXHDGOGnfmdOdIkWlxrBHrtgLc/edit)
7. [Web component benchmarking](https://docs.google.com/spreadsheets/d/1_mUVGwYBOuH1T5_vtc0us-CLnLUspizdEy5HXwVICvA/edit#gid=2066730968)
8. [WAI ARIA Listbox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/)
9. [WAI ARIA Example Listbox Grouped](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-grouped/)
10. [WAI ARIA Combobox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)


elements/pf-listbox/demo/pf-listbox.js
```diff
@@ -0,0 +1,29 @@
+import '@patternfly/elements/pf-listbox/pf-listbox.js';
+
+window.addEventListener('load', () => {
```
bennypowers, line 3: modules don't need to wait on `load`, since they're always deferred

elements/package.json
bennypowers, line 1: I like these elements, but I think we can't include them in `@patternfly/elements`, since they're not implemented upstream.

elements/pf-listbox/pf-listbox-group.ts
```diff
@@ -0,0 +1,45 @@
...
+      <slot @slotchange="${this.#onSlotchange}"></slot>
+    `;
+  }
+
+  get options() {
+    return this._options;
+  }
+
+  #onSlotchange() {
+    this.dispatchEvent(new Event('slotchange', { bubbles: true }));
```
bennypowers, line 37: not needed, slotchange already bubbles

elements/pf-listbox/pf-listbox.ts
```diff
@@ -0,0 +1,362 @@
...
+
+  render() {
+    return html`
+      <slot 
+        class="${this.isHorizontal ? 'horizontal' : 'vertical'}"
+        @slotchange="${this.#onSlotchange}" 
+        @optionfocus="${this.#onOptionFocus}"
+        @click="${this.#onOptionClick}"
+        @keydown="${this.#onOptionKeydown}"
+        @keyup="${this.#onOptionKeyup}">
```
bennypowers, line 140: in the absence of `<pf-listbox>`, we could share code here by snipping an [element-position directive](https://lit.dev/docs/templates/custom-directives/#parts) on some hypothetical ListboxController

```ts
html`
  <slot class="..." ${this.#listbox.listbox()}></slot>

bennypowers (changes requested):
I'm excited about this work. We'll have to get creative about how we ship reusable code here, as we want to limit ourselves in `@patternfly/elements` to existing components. We have more flexibility in `@patternfly/pfe-core`, so I think we should explore what the Lit templating and reactivity APIs have to offer, since most of the code here is or could be made agnostic to specific DOM structure and styles

elements/pf-listbox/pf-listbox.ts
nikkimk: Moved function to `listbox-controller`. Since we so have a `simple-list` (and several other components that use it) upstream, I renamed `pf-listbox` to `simple-list` in case we want to keep it.

elements/package.json
nikkimk: Since we so have a [`simple-list`](https://www.patternfly.org/components/simple-list#) upstream, I renamed `pf-listbox` to `simple-list` in case we want to keep it. There are also the following complex upstream elements that include listboxes:
- [Dual-list selector](https://www.patternfly.org/components/dual-list-selector)
- [Menu, select](https://www.patternfly.org/components/menus/select/)
- [Menu, custom menus](https://www.patternfly.org/components/menus/custom-menus/)

nikkimk:
Moved to #2570

PR #2572, opened 2023-08-02
feat(text-input): add pf-text-input

bennypowers:
## What I did

1. add `<pf-text-input>`

## Testing Instructions

1. check out the dp with emphasis on screen readers

## Notes to Reviewers

1. The approach here for FACE and aria labelling will need to be validated as we go


bennypowers:
@hellogreg @nikkimk this is very early days, but i'd like to hear some initial feedback on the approach taken here https://deploy-preview-2572--patternfly-elements.netlify.app/components/text-input/demo/kitchen-sink/

bennypowers:
@hellogreg the only instance that's used correctly with labels on that page for that commit is the first one, so yes your intuition was correct

For the left truncated version, i'm trying to thread the needle here...
the react version replaces the value with the string "...whatever the rest", so when you copy it, you get the ellipsis

What I've done here is a css hack with `direction:rtl`, which doesn't really work properly :/, but you do get the whole value when copying

bennypowers:
I think this is feature-complete now, the above caveats notwithstanding.
still needs docs

elements/pf-text-input/pf-text-input.ts
```diff
@@ -0,0 +1,261 @@
...
+
+  /** Trim text on left */
+  @property({ type: Boolean, reflect: true, attribute: 'left-truncated' }) leftTruncated = false;
+
+  /** Value to indicate if the input is modified to show that validation state.
+   * If set to success, input will be modified to indicate valid state.
+   * If set to warning,  input will be modified to indicate warning state.
+   * Invalid inputs will display an error state
+   */
+  @property({ reflect: true }) validated?: 'success' | 'warning';
```
zeroedin, line 163: PFv4 looks to have `'success' | 'warning' | 'error' | 'default'` is there a reason not to have the `error` option here?

elements/pf-text-input/pf-text-input.ts
```diff
@@ -0,0 +1,261 @@
...
+  }
+
+  override willUpdate() {
+    /** A best-attempt based on observed behaviour in FireFox 115 on fedora 38 */
+    this.#derivedLabel =
+      this.accessibleLabel ||
+      this.#internals.ariaLabel ||
+      Array.from(this.#internals.labels as NodeListOf<HTMLElement>)
+        .reduce((acc, label) =>
+          `${acc}${getLabelText(label)}`, '');
```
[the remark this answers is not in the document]
bennypowers: In this case I just tried to emulate Firefox' behavior (as shown in the accessibility panel) when multiple labels are applied to an input.

This function is a best-effort, and I presume there will be cases we just won't be able to handle

elements/pf-text-input/pf-text-input.ts
zeroedin: I was thinking if set declaratively it could be set by a server template form rerender?  Maybe I'm not thinking through that completely.  Does this always assume client side validation?

bennypowers: it could be :shrug: i'd be willing to defer that until the request comes in

zeroedin (approved):

PR #2573, opened 2023-08-07
docs(button): examples of onclick

bennypowers:
## What I did

1.

## Testing Instructions

1.

## Notes to Reviewers

1.

### DIFF
--- a/docs/components/components.md
+++ b/docs/components/components.md
@@ -15,6 +15,27 @@
 }
 ---
 
+<script type="module">
+  const LS_KEY = 'html-lit-react-snippets-index';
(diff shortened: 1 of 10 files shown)


docs/components/components.md
```diff
@@ -15,6 +15,27 @@
 }
 ---
 
+<script type="module">
```
brianferry, line 18: Does this code need to be available on all component pages or could it be a script pages import if they're using it?

bennypowers: We could break it out but then contributors would have to remember

brianferry (approved):

zeroedin (approved):

PR #2576, opened 2023-08-09
feat(table): expansion slots

bennypowers:
experimental approach to expandable rows...
still broken


bennypowers:
I feel pretty good about this now so let's merge it and hope for the best :)

PR #2582, opened 2023-08-21
fix(progress): fixing css selectors for variants

brianferry:
## What I did

1.  Updated `success`, `warning`, and `danger` variant css to fix variant background color.


zeroedin (approved):
LGTM

PR #2586, opened 2023-08-21
feat(card)!: remove BaseCard

brianferry:
## What I did

1.  Moved out the render and css from BaseCard into `pf-card`.  
2. Removed `BaseCard` class and styling.


.changeset/sharp-spiders-float.md
```diff
@@ -0,0 +1,7 @@
+---
+"@patternfly/elements": minor
+---
+
+`BaseCard`: Deprecating BaseCard
+
+`pf-card`: Removing dependency on BaseCard.  Moved styles and logic into `pf-card` element.
```
bennypowers, line 7:
```suggestion
`<pf-card>`: Removes `BaseCard` base class. If your project extends `BaseCard`, we recommend extending `LitElement` instead and re-implementing card's properties. Alternately, extend from `PfCard`.
```

bennypowers (changes requested):
Some elaboration on the changeset. Preferably there would be a before/after as well, but that might be a bit much for this case

.changeset/sharp-spiders-float.md
bennypowers, line 1: we can't release this until the next major. cc @hellogreg

bennypowers (commented):
if we merge this, to main, we'll enter release season for pfe 3

bennypowers (approved):

PR #2588, opened 2023-08-22
feat(background-image): pf-background-image element

zeroedin:
## What I did

1. Added `<pf-background-image>` element

### TODO

- [x] ~~Determine if container based examples need to exist. Background image [which states](http://v4-archive.patternfly.org/v4/components/background-image) they can be part of an "area of a page":  Summary: A background image allows you to place an image in the background of your page or area of a page.~~ 
Answer: Discussion with the PatternFly team resulted in the following "The current implementation doesn’t support “Area of page” and "sounds like we should update the summary of that component as it currently is described on the website. I’ll open an issue" https://github.com/patternfly/patternfly-org/issues/3729
- [x] Improve docs / add screenshot image
- [x] Add changeset
- [x] Add tests

## Testing Instructions

1.  View [Deploy Preview](https://deploy-preview-2588--patternfly-elements.netlify.app/components/background-image/)

## Notes to Reviewers

1.


elements/pf-background-image/pf-background-image.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  @queryAssignedElements({ slot: 'filter', selector: 'svg' }) private _svg?: SVGElement[];
+
+  #svg?: SVGElement;
+
+  #updated = false;
+
+  render() {
+    return html`
+      <div id="container" part="container" style="${this.#cssVars()}">
```
bennypowers, line 66: better to do this as a styleMap, and you can keep everything in the render method IMO

elements/pf-background-image/pf-background-image.ts
```diff
@@ -0,0 +1,122 @@
...
+                      <feFuncG type="table" tableValues="0.086274509803922 0.43921568627451"></feFuncG>
+                      <feFuncB type="table" tableValues="0.086274509803922 0.43921568627451"></feFuncB>
+                      <feFuncA type="table" tableValues="0 1"></feFuncA>
+                    </feComponentTransfer>
+                  </filter>
+                </svg>
+              `}
+          </slot>
+        `}
+        <div class="content" part="content">
```
bennypowers, line 85: do we need the wrapper div? why not `<slot id="content" part="content">` and set `display: block`?

zeroedin: Mostly likely could do that, trying to remember what I ran into to add the wrap in the first place.  Might have been the question of needing the content slot in the first place.  I don't think PF implementation shows that idea in code IIRC / part of the first todo of determining if it even needs to exist.  But yes, wouldn't see why this suggested change wouldn't work.

elements/pf-background-image/pf-background-image.ts
zeroedin: 💯 didn't even think about that

PR #2589, opened 2023-09-05
feat(back-to-top): pf-back-to-top element

zeroedin:
## What I did

1.  Added `<pf-back-to-top>` element

### Todo 
- [x] ~~Determine accessibility improvements~~ A href attribute and using a `<a>` in shadowdom is warranted as expected.  Can still use button version but there is caveat of owning accessibility in implementation, such as setting focus on element scrolled to.
- [x] ~~Figure out improvements for how a user hooks up element to "returning to top".  PFv4 doesn't make assumptions on how the button is hooked up.  We may opt for a `href` attribute and use a link instead of a button when present, and definitely have to add a event when button is clicked otherwise for implementors to hook into.~~ See above, `href` implementation
- [x] Improve documentation
- [x] Add tests
- [x] Add changeset

## Testing Instructions

1. View [Deploy Preview](https://deploy-preview-2589--patternfly-elements.netlify.app/components/back-to-top/)

## Notes to Reviewers

1.


elements/pf-back-to-top/pf-back-to-top.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    this.#scrollSpy = !!this.scrollableSelector;
+
+    if (this.scrollableSelector?.trim() === '') {
+      this.#logger.error(`scrollable-selector attribute cannot be empty`);
+      return;
+    }
+
+    if (this.#scrollSpy && !!this.scrollableSelector) {
+      const scrollableElement = this.#rootNode.querySelector(this.scrollableSelector) as HTMLElement;
```
bennypowers, line 88: shouldn't need the type assertion since we have the `if`

bennypowers: if you need to, change the predicate to `if (!(scrollableElement instanceof HTMLElement))`

elements/pf-back-to-top/pf-back-to-top.ts
```diff
@@ -0,0 +1,154 @@
...
+      if (!scrollableElement) {
+        this.#logger.error(`Unable to find element with selector ${this.scrollableSelector}`);
+        return;
+      }
+      this.#scrollElement = scrollableElement;
+    } else {
+      this.#scrollElement = window;
+    }
+
+    this.#scrollElement.addEventListener('scroll', this.#toggleVisibility, { passive: true });
```
bennypowers, line 98: this needs to go in a reaction effect for `scrollable-selector`, and make sure it's also called on connect if lit framework doesn't do so already

elements/pf-back-to-top/pf-back-to-top.ts
```diff
@@ -0,0 +1,154 @@
...
+    this.#toggleVisibility();
+  }
+
+  override disconnectedCallback(): void {
+    super.disconnectedCallback?.();
+    this.#scrollElement?.removeEventListener('scroll', this.#toggleVisibility);
```
bennypowers, line 104: make sure to also do this at the top of the reaction callback from the previous comment

elements/pf-back-to-top/pf-back-to-top.ts
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  render() {
+    const classes = { 'visually-hidden': !this.#visible };
```
bennypowers, line 108: consider calling this `visuallyHiddenClass`

elements/pf-back-to-top/test/pf-back-to-top.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+  let element: PfBackToTop;
+  let link: HTMLAnchorElement;
+
+  beforeEach(async function() {
+    await setViewport({ width: 320, height: 640 });
+    window.scrollTo({ top: 0, behavior: 'instant' });
+    await nextFrame();
+  });
+
+  describe('default a tag implementation', function() {
```
bennypowers, line 19:
```suggestion
  describe('with `href` attribute', function() {
```

elements/pf-back-to-top/test/pf-back-to-top.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    beforeEach(async function() {
+      await fixture<HTMLElement>(html`<div style="height: 2000px;"></div>`);
+      element = await createFixture<PfBackToTop>(html`<pf-back-to-top href="#top">Back to Top</pf-back-to-top>`);
+      await allUpdates(element);
+      link = element.shadowRoot!.querySelector('a')!;
+    });
+
+    it('imperatively instantiates', function() {
```
bennypowers, line 27: lift to outer scope, no href here

elements/pf-back-to-top/test/pf-back-to-top.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+      expect(document.createElement('pf-back-to-top')).to.be.an.instanceof(PfBackToTop);
+    });
+
+    it('should upgrade', async function() {
```
bennypowers, line 31: same

elements/pf-back-to-top/test/pf-back-to-top.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+      const klass = customElements.get('pf-back-to-top');
+      expect(element)
+        .to.be.an.instanceOf(klass)
+        .and
+        .to.be.an.instanceOf(PfBackToTop);
+    });
+
+    it('should be visible after 400px scroll', async function() {
```
bennypowers, line 39: describe/beforeEach to setup state, it/expect to examine state. use nested describes.

elements/pf-back-to-top/docs/pf-back-to-top.md
```diff
@@ -0,0 +1,95 @@
...
+
+  ### Scrollable Selector
+  {% htmlexample %}
+    <div id="scrollable-selector-example">
+      <div class="overfill">
+        <div class="scroll-indicator">
+          <pf-icon icon="arrow-down"></pf-icon> Scroll down to end of cyan box, 400px (default).
+        </div>
+      </div>
+      <pf-back-to-top scrollable-selector="#scrollable-selector-example" href="#main">Back to Top</pf-back-to-top>
```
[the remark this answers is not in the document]
zeroedin: The `href` target is `#main` which is the `<main>` element they should scroll the page to the top of that element.  The `scrollable-selector` is a baseline selector to measure when the back to top button would become visible via intersection observer and set scroll distance.  Are you expecting these demos to scroll just the overflowed element or something different?   

I personally do agree it is a bit confusing, so maybe these demos don't really benefit the user inline like this and should be linked to a specific full page demo instead.

The scroll on this page is made this even more confusing given the layout of the page itself flows under the header.  I believe this is flaw with PFE's docs layout.

elements/pf-back-to-top/pf-back-to-top.ts
```diff
@@ -0,0 +1,169 @@
...
+  override updated(changed: PropertyValues<this>): void {
+    super.updated(changed);
+    if (changed.has('scrollableSelector')) {
+      if (changed.get('scrollableSelector') === undefined) {
+        this.#scrollElement?.removeEventListener('scroll', this.#toggleVisibility);
+      } else {
+        this.#setScrollListener();
+      }
+    }
+  }
```
bennypowers, line 96:
```suggestion
  override disconnectedCallback(): void {
    super.disconnectedCallback?.();
    this.#removeScrollListener();
  }

  override willUpdate(changed: PropertyValues<this>): void {
    if (changed.has('scrollableSelector')) {
      this.#addScrollListener();
    }
  }
```

This should be enough, but check the tests

elements/pf-back-to-top/pf-back-to-top.ts
```diff
@@ -0,0 +1,169 @@
...
+        return;
+      }
+      this.#scrollElement = scrollableElement;
+    } else {
+      this.#scrollElement = window;
+    }
+
+    this.#scrollElement.addEventListener('scroll', this.#toggleVisibility, { passive: true });
+    this.#toggleVisibility();
+  }
```
bennypowers, line 148:
```suggestion
  #removeScrollListener() {
    this.#scrollElement?.removeEventListener('scroll', this.#toggleVisibility);
  }
  
  #addScrollListener() {
    this.#removeScrollListener();
    this.#scrollSpy = !!this.scrollableSelector;

    if (this.scrollableSelector?.trim() === '') {
      this.#logger.error(`scrollable-selector attribute cannot be empty`);
      return;
    }

    if (this.#scrollSpy && !!this.scrollableSelector) {
      const scrollableElement = this.#rootNode.querySelector(this.scrollableSelector);
      if (!scrollableElement) {
        this.#logger.error(`unable to find element with selector ${this.scrollableSelector}`);
        return;
      }
      this.#scrollElement = scrollableElement;
    } else {
      this.#scrollElement = window;
    }

    this.#scrollElement.addEventListener('scroll', this.#toggleVisibility, { passive: true });
    this.#toggleVisibility();
  }
```

elements/pf-back-to-top/pf-back-to-top.ts
```diff
@@ -0,0 +1,169 @@
...
+  render() {
+    const visuallyHiddenClass = { 'visually-hidden': !this.#visible };
+
+    // ensure href has a hash
+    if (this.href && !this.href.includes('#')) {
+      this.href = `#${this.href}`;
+      this.#logger.warn(`missing hash in href fragment link`);
+    }
+
+    return this.href ? html`
```
bennypowers, line 107: nit, i think it's acceptable to use an `if` here instead of a tern, WDYT?

elements/pf-back-to-top/test/pf-back-to-top.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+      link = element.shadowRoot!.querySelector('a')!;
+    });
+
+    it('should be visible after 400px scroll', async function() {
+      expect(link).to.have.class('visually-hidden');
+      window.scrollTo({ top: 401, behavior: 'instant' });
+      await allUpdates(element);
+      await nextFrame();
+      expect(link).to.not.have.class('visually-hidden');
+    });
```
bennypowers, line 47:
```suggestion
    it('should be hidden', async function() {
      const snapshot = await a11ySnapshot();
      expect(snapshot).to.deep.equal(/* expected */);
    });
    
    describe('after 400px scroll', function() {
      beforeEach(async function() {
        window.scrollTo({ top: 401, behavior: 'instant' });
        await allUpdates(element);
        await nextFrame();
      });
          
      it('should be visible', async function() {
        const snapshot = await a11ySnapshot();
        expect(snapshot).to.deep.equal(/* expected */);
      });
    });

```

etc, in other places. don't assert on shadow dom, instead assert on the ax tree. `it(title, async function() {});` is a smell, the `async` should not be necessary. we can make an exception for asserting on the ax tree because that's an async function, but in generally we should seek to find async functions mostly in beforeEach blocks only.

elements/pf-back-to-top/test/pf-back-to-top.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  describe('pf-button implementation', function() {
+    beforeEach(async function() {
+      await fixture<HTMLElement>(html`<div style="height: 2000px;"></div>`);
+      element = await createFixture<PfBackToTop>(html`<pf-back-to-top>Back to Top</pf-back-to-top>`);
+      await allUpdates(element);
+    });
+
+    it('should contain a pf-button if an href is not set', async function() {
+      expect(element.shadowRoot!.querySelector('pf-button')).to.exist;
```
bennypowers, line 76: assert on ax tree, not on shadow root

bennypowers (changes requested):
last few changes

elements/pf-back-to-top/test/pf-back-to-top.spec.ts
zeroedin: I'm not modifying the ally tree here though, so not sure what to assert on there.  Its not `hidden` its `visually-hidden` off screen.  This leaves the link available to screen readers at all times since they don't "scroll".  Maybe need to reevaluate that from an accessibility standpoint /cc @nikkimk

bennypowers: what happens when you focus it?

zeroedin: Point taken, large refactor of the component done along with tests, PTAL.

elements/README.md
bennypowers, line 1: please revert

bennypowers (changes requested):

elements/pf-back-to-top/docs/pf-back-to-top.md
```diff
@@ -0,0 +1,106 @@
...
+{% renderOverview %}
+  <pf-back-to-top href="#main">Back to top</pf-back-to-top>
+{% endrenderOverview %}
+
+{% band header="Usage" %}
+
+  ### Default
+  {% htmlexample %}<pf-back-to-top href="#main">Back to top</pf-back-to-top>{% endhtmlexample %}
+
+  ### Label attribute
```
nikkimk, line 53: Add some text explaining what the difference between the label and the slot are.

elements/pf-back-to-top/docs/pf-back-to-top.md
```diff
@@ -0,0 +1,106 @@
...
+  {% htmlexample %}<pf-back-to-top href="#main" label="Back to top"></pf-back-to-top>{% endhtmlexample %}
+
+
+  ### No text 
+  {% htmlexample %}<pf-back-to-top href="#main"></pf-back-to-top>{% endhtmlexample %}
```
nikkimk, line 58: More text here perhaps.

elements/pf-back-to-top/docs/pf-back-to-top.md
```diff
@@ -0,0 +1,106 @@
...
+        <div class="scroll-indicator">
+          <pf-icon icon="arrow-down"></pf-icon> Scroll down to end of cyan box, 400px (default).
+        </div>
+      </div>
+      <pf-back-to-top href="#main" scrollable-selector="#scrollable-selector-example">Back to top</pf-back-to-top>
+    </div>
+  {% endhtmlexample %}
+
+  ### Scroll Distance
+  {% htmlexample %}
```
nikkimk, line 75: Explain the difference between scroll and scroll distance.

elements/pf-back-to-top/docs/pf-back-to-top.md
```diff
@@ -0,0 +1,106 @@
...
+    </div>
+  {% endhtmlexample %}
+
+  </div>
+
+  ### Content guidelines
+  - The button should consistently be placed on the lower right side of the page.
+  - Only use 1 back to top component per page, avoid using multiple buttons in different sections.
+  - Keep the button stationary. Once the button reveals itself, it should not move on the page.
+  - Include proper spacing around the button, see CSS variables under React and HTML tabs.
```
nikkimk, line 92: Was `Include proper spacing around the button, see CSS variables under React and HTML tabs.` was pulled from upstream?

elements/pf-back-to-top/docs/pf-back-to-top.md
```diff
@@ -0,0 +1,106 @@
...
+
```
nikkimk, line 93: `  - Button must not have an ancestor with \`position: relative\`.`

elements/pf-back-to-top/docs/pf-back-to-top.md
nikkimk, line 75: Explain the difference between scroll and scroll distance.

elements/pf-back-to-top/docs/pf-back-to-top.md
nikkimk, line 58: Might be worth explaining how it is labelled when there is no text.

elements/pf-back-to-top/docs/pf-back-to-top.md
```diff
@@ -0,0 +1,106 @@
...
+  <pf-back-to-top href="#main">Back to top</pf-back-to-top>
+{% endrenderOverview %}
+
+{% band header="Usage" %}
+
+  ### Default
+  {% htmlexample %}<pf-back-to-top href="#main">Back to top</pf-back-to-top>{% endhtmlexample %}
+
+  ### Label attribute
+  {% htmlexample %}<pf-back-to-top href="#main" label="Back to top"></pf-back-to-top>{% endhtmlexample %}
```
nikkimk, line 54: Add some text explaining what the difference between the label and the slot are.

elements/pf-back-to-top/test/pf-back-to-top.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+
+      await allUpdates(element);
+    });
+
+    it('should be hidden on init', async function() {
+      const { children } = snapshot;
+      expect(children).to.be.undefined;
+    });
+
+    it('should be visible after scrolling 401px', async function() {
```
nikkimk, line 59: Wearing my @bennypowers hat, use beforeEach for async:

```
 describe('when scrolling 401px', function() {
      beforeEach(async function() {
          window.scrollTo({ top: 401, behavior: 'instant' });
          await nextFrame();
          await allUpdates(element);
          snapshot = await a11ySnapshot();
      });
     it('should be visible', async function() {
          expect(snapshot.children?.map(takeProps(['name', 'role']))).to.deep.equal([{ role: 'link', name: 'Back to top' }]);
    });
});
    ```

File: elements/pf-back-to-top/test/pf-back-to-top.spec.ts [hunk 22]
```diff
@@ -0,0 +1,189 @@
...
+        await allUpdates(element);
+        snapshot = await a11ySnapshot();
+      });
+
+      it('should be hidden', async function() {
+        const { children } = snapshot;
+        expect(children).to.be.undefined;
+      });
+
+      it('should be visible after scrolling 1001px', async function() {
```
nikkimk, line 93: See previous comment.

elements/pf-back-to-top/test/pf-back-to-top.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+          <pf-back-to-top href="#top" scrollable-selector="#top">Back to top</pf-back-to-top>
+        </div>
+      `);
+      element = container.querySelector('pf-back-to-top')!;
+      snapshot = await a11ySnapshot();
+
+      await allUpdates(element);
+    });
+
+    it('should be hidden on init', async function() {
```
nikkimk, line 122: remove `async` here

elements/pf-back-to-top/test/pf-back-to-top.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+      const { children } = snapshot;
+      expect(children).to.be.undefined;
+    });
+
+    it('should be visible after scrolling 401px', async function() {
```
nikkimk, line 127: see comments above

elements/pf-back-to-top/test/pf-back-to-top.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+          <div id="top">
+            <div style="height: 2000px;"></div>
+            <pf-back-to-top href="#top"></pf-back-to-top>
+          </div>
+        `);
+        element = container.querySelector('pf-back-to-top')!;
+        await allUpdates(element);
+      });
+
+      it('should have a label of "Back to top"', async function() {
```
nikkimk, line 156: See above

elements/pf-back-to-top/test/pf-back-to-top.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+          <div id="top">
+            <div style="height: 2000px;"></div>
+            <pf-back-to-top></pf-back-to-top>
+          </div>
+        `);
+        element = container.querySelector('pf-back-to-top')!;
+        await allUpdates(element);
+      });
+
+      it('should have a label of "Back to top"', async function() {
```
nikkimk, line 180: See above

elements/pf-back-to-top/test/pf-back-to-top.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+        window.scrollTo({ top: 401, behavior: 'instant' });
+        await nextFrame();
+        await allUpdates(element);
+        snapshot = await a11ySnapshot();
+        expect(snapshot.children?.map(takeProps(['name', 'role']))).to.deep.equal([{ role: 'button', name: 'Back to top' }]);
+      });
+    });
+  });
+});
```
nikkimk, line 189: Do we want to include some tests for the following?
- whether it is keyboard navigable
- whether there is an accessible label when there is nothing slotted and label property is blank

nikkimk (dismissed):

elements/pf-back-to-top/docs/pf-back-to-top.md
zeroedin: The more and more I think about I think this is where we deviate from PF's API and just kill the `label` attribute.  The only functional difference here would be the slot allowing additional `<html>` in the vs an attribute being text only in the label.

bennypowers: We also have `accessible-label` elsewhere

elements/pf-back-to-top/docs/pf-back-to-top.md
zeroedin: Yes this whole list was pulled from PF docs, actually forgot I copied it over.  Needs to be reworded for our use.

elements/pf-back-to-top/docs/pf-back-to-top.md
zeroedin: Refactored `label` completely that removes this ambiguity. PTAL

elements/pf-back-to-top/test/pf-back-to-top.spec.ts
```diff
@@ -56,6 +56,10 @@ describe('<pf-back-to-top>', function() {
       expect(children).to.be.undefined;
     });
 
+    it('should not be accessible', function() {
+      expect(element).to.not.be.accessible;
```
bennypowers, line 60: This won't do anything. You need to call and await accessible
```suggestion
      await expect(element).to.not.be.accessible();
```

I think snapshot would be better suited for this assertion

bennypowers (changes requested):

bennypowers (approved):

bennypowers:
Merging ahead of 2.5.0. @nikkimk @hellogreg please open new issues for anything that comes up

PR #2592, opened 2023-09-19
docs: make installation section reusable 

brianferry:
## What I did

1.  Added name variable to the `install.njk` file so that we can use this render in RHDS and fix https://github.com/RedHat-UX/red-hat-design-system/issues/1218


brianferry:
Closing as we're going to fix the issue in RHDS with https://github.com/RedHat-UX/red-hat-design-system/pull/1236

Issue #2536, opened 2023-07-06
pf-date-picker


Arathy-s:
# Component name: <pf-date-picker>

> A date picker enables users to select or input a date. It should allow users to select or input a date by day, month and year.

![Screenshot 2023-07-06 at 12 51 08 PM](https://github.com/patternfly/patternfly-elements/assets/49025521/b737beb8-a07a-4d1a-bbeb-9d1e2f2e4588)

## Design specification

The design specifications should match with https://www.patternfly.org/v4/components/date-picker/design-guidelines.

## Accessibility

The component should implement all the accessibility features as https://www.patternfly.org/v4/components/date-picker.

PR #2599, opened 2023-09-27
Feature/pf date picker

Arathy-s:
## What I did

Created date-picker component to resolve the issue https://github.com/patternfly/patternfly-elements/issues/2536.


elements/pf-date-picker/demo/demo.css
bennypowers, line 1: we should prefer to inline all the css and js for the demos directly into the html files

elements/pf-date-picker/demo/pf-date-picker.html
bennypowers, line 1: we should prefer to split large demo files up into individual html files, i.e. `pf-date-picker.html` for "basic", `date-format.html` for "date format`, etc.

elements/pf-date-picker/demo/pf-date-picker.html
```diff
@@ -0,0 +1,75 @@
...
+    <pre><code>&lt;pf-date-picker&gt;&lt;/pf-date-picker&gt;</code></pre>
+    <pf-date-picker></pf-date-picker>
+    <p>The basic date picker will use the user's locale to determine the date format. The default minimum valid date 
+      is <span>January 1, 1900</span>, and the default maximum valid date is <span>December 31, 9999</span>.</p>
+  </section>
+
+  <section>
+    <h2>Date Format</h2>
+    <pre><code>&lt;pf-date-picker dateFormatInput="YYYY-DD-MM"&gt;&lt;/pf-date-picker&gt;</code></pre>
+    <pf-date-picker dateFormatInput="YYYY-DD-MM"></pf-date-picker>
```
bennypowers, line 16: we should prefer dash-case attributes for camelCase DOM properties, i.e. `date-format-input`, since HTML attributes are case-insensitive.

elements/pf-date-picker/demo/pf-date-picker.html
```diff
@@ -0,0 +1,75 @@
...
+    </pf-date-picker>
+    <p>The minimum and maximum valid dates can be passed to the date picker to restrict the range of dates that can be selected</p>
+    <p>In the above given example, the minimum valid date 
+      is <span>January 2, 2023</span>, and the maximum valid date is <span>January 20, 2023</span></p>
+  </section>
+
+  <section>
+    <h2>Disabled</h2>
+    <pre><code>&lt;pf-date-picker isDisabled="true"&gt;&lt;/pf-date-picker&gt;</code></pre>
+    <pf-date-picker isDisabled="true"></pf-date-picker>
```
bennypowers, line 71: We typically remove the `is` prefix for boolean attributes, see `<pf-button disabled>` for example

elements/package.json
```diff
@@ -33,6 +33,13 @@
     "./pf-clipboard-copy/pf-clipboard-copy.js": "./pf-clipboard-copy/pf-clipboard-copy.js",
     "./pf-code-block/BaseCodeBlock.js": "./pf-code-block/BaseCodeBlock.js",
     "./pf-code-block/pf-code-block.js": "./pf-code-block/pf-code-block.js",
+    "./pf-date-picker/date-picker-helper.js": "./pf-date-picker/date-picker-helper.js",
+    "./pf-date-picker/pf-calendar.js": "./pf-date-picker/pf-calendar.js",
+    "./pf-date-picker/pf-date-picker.js": "./pf-date-picker/pf-date-picker.js",
+    "./pf-date-picker/pf-month-select.js": "./pf-date-picker/pf-month-select.js",
+    "./pf-date-picker/pf-next-button.js": "./pf-date-picker/pf-next-button.js",
+    "./pf-date-picker/pf-previous-button.js": "./pf-date-picker/pf-previous-button.js",
+    "./pf-date-picker/pf-year-input.js": "./pf-date-picker/pf-year-input.js",
```
bennypowers, line 42: since upstream only provides the `DatePicker` react component, we should also try to limit the number of public components we ship here. We may make an exception if, for accessibility reasons, we need to put these elements in the same root. However, in the case of date picker, We can probably allow ourselves the luxury of doing everything in the same shadow root. We'll need to investigate that carefully

elements/pf-date-picker/README.md
bennypowers, line 1: this should be updated with prose from v4-archive.patternfly.org

elements/pf-date-picker/date-picker-helper.ts
bennypowers, line 1: Note to future self and other reviewers: see if / how many of these functions can be replaced with Intl built-ins.

elements/pf-date-picker/docs/pf-date-picker.md
bennypowers, line 1: This should be updated with some prose and examples

elements/pf-date-picker/pf-calendar.ts
```diff
@@ -0,0 +1,269 @@
+import { LitElement, html } from 'lit';
+import { customElement } from 'lit/decorators/custom-element.js';
+import { property } from 'lit/decorators/property.js';
+import { ref, createRef } from 'lit/directives/ref.js';
+import styles from './pf-calendar.css';
+import { getFormattedDate, getMonthNamesFromLocale } from './date-picker-helper.js';
+import { ComposedEvent } from '@patternfly/pfe-core/core.js';
```
bennypowers, line 7: prefer to extend `Event` with `bubbles: true` and optionally `cancellable: true`, rather than using `ComposedEvent`

elements/pf-date-picker/pf-calendar.ts
```diff
@@ -0,0 +1,269 @@
...
+
+/**
+ * Calendar
+ * @slot - Place element content here
+ */
+@customElement('pf-calendar')
+export class PfCalendar extends LitElement {
+  static readonly styles = [styles];
+
+  private currentDate: Date = new Date();
```
bennypowers, line 53: prefer ecmascript private fields where possible
```suggestion
  #currentDate = new Date();
```

also note that in this case you can do without the type annotation, TS will figure it out from the initializer on the RHS

elements/pf-date-picker/pf-calendar.ts
```diff
@@ -0,0 +1,269 @@
...
+  private weekdays: number[] = [0, 1, 2, 3, 4, 5, 6]; // S, M, T, W, T, F, S
+  private weeks: number[] = [0, 1, 2, 3, 4, 5]; // 1 previous month week, 4 current month weeks, 1 next month week
+
+  // Input properties from the parent
+  @property() currentYear: number = this.currentDate.getFullYear();
```
bennypowers, line 58: I noticed this isn't a public prop on pfv4 DatePicker or CalendarFormat, so do we need it to be a public `@property` here? or can it be a private field?

if it does need to be private, see notes on the demo file regarding attribute names
```suggestion
  @property({ type: Number, attribute: 'current-year' }) currentYear: number = this.currentDate.getFullYear();
```

elements/pf-date-picker/pf-calendar.ts
```diff
@@ -0,0 +1,269 @@
...
+  @property() currentMonth: number = this.currentDate.getMonth();
+  @property() currentWeek = 0;
+  @property() selectedDay: number = this.currentDate.getDate();
+  @property() focusRef!: HTMLButtonElement;
```
bennypowers, line 62: we shouldn't need this to be a public `@property`

elements/pf-date-picker/pf-calendar.ts
```diff
@@ -0,0 +1,269 @@
...
+ * Calendar
+ * @slot - Place element content here
+ */
+@customElement('pf-calendar')
+export class PfCalendar extends LitElement {
+  static readonly styles = [styles];
+
+  private currentDate: Date = new Date();
+  private weekdays: number[] = [0, 1, 2, 3, 4, 5, 6]; // S, M, T, W, T, F, S
+  private weeks: number[] = [0, 1, 2, 3, 4, 5]; // 1 previous month week, 4 current month weeks, 1 next month week
```
bennypowers, line 55: if these are constants, we probably don't need to store them on the class.

bennypowers (changes requested):
some initial review to get the ball rolling

elements/pf-date-picker/demo/demo.css
Arathy-s: Added inline css and js for the html demo files.

elements/package.json
Arathy-s: Done. 
Removed the child components exports list.

elements/pf-date-picker/pf-calendar.ts
Arathy-s: Added constants to the helper file.

elements/pf-date-picker/pf-calendar.ts
```diff
@@ -0,0 +1,269 @@
...
+  @property() minDate!: Date;
+  @property() maxDate!: Date;
+  @property() translationLanguageCode!: string;
+  @property() monthNames: string[] = getMonthNamesFromLocale(this.translationLanguageCode);
+  focusDateRef = createRef<HTMLButtonElement>(); // Reference to the button that needs to be focused
+
+  constructor() {
+    super();
+  }
+
```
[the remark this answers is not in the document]
Arathy-s: Removed the constructor.

elements/pf-date-picker/demo/pf-date-picker.html
Arathy-s: Changed the DOM properties format from cameCase to dash-case.

elements/pf-date-picker/demo/pf-date-picker.html
Arathy-s: Removed the `is` prefix form the disabled property.

elements/pf-date-picker/pf-next-button.ts
bennypowers, line 1: These shouldn't be separate elements, but rather pf-button in the parent template

elements/pf-date-picker/pf-date-picker.ts
```diff
@@ -0,0 +1,599 @@
...
+                      <div class="date-picker-table-col date-picker-table-month-year">
+                        <pf-month-select 
+                          .currentMonthName=${this.currentMonthSelection}
+                          .currentMonthIndex=${this.currentMonthIndex}
+                          @currentMonth=${this.#getCurrentMonth}
+                          .translationLanguageCode=${this.translationLanguageCode}
+                          @monthExpandState=${this.#getMonthExpandState}
+                          .isMonthExpanded=${this.monthExpand}>
+                        </pf-month-select>
+                        <pf-year-input
```
zeroedin, line 211: Similar to the above, I would think we'd want to use `<pf-text-input type="number">` instead of creating a new component here.  However I do see that pf-text-input doesn't support `min` and `max` attrs which are needed here, @bennypowers since PF4 doesn't appear to extend these attributes on a number type, would we just want to implement a standard `<input type="number">`?

elements/pf-date-picker/pf-date-picker.ts
```diff
@@ -0,0 +1,599 @@
...
+                    <div class="date-picker-table-row date-picker-table-main-header">
+                      <div class="date-picker-table-col">
+                        <pf-previous-button 
+                          .currentYear=${this.currentYear} 
+                          .currentMonth=${this.currentMonthIndex} 
+                          @previousMonthAndYear=${this.#getPreviousMonthAndYear}>
+                        </pf-previous-button>
+                      </div>
+                      <div class="date-picker-table-col date-picker-table-month-year">
+                        <pf-month-select
```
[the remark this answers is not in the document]
Arathy-s: Hey @zeroedin 
Thank you so much for reviewing.

I developed the <pf-month-select> component for the date-picker because <pf-dropdown> was not yet available. I will go ahead and change it now to <pf-dropdown> as it will be a much simpler implementation.

elements/pf-date-picker/README.md
Arathy-s: Updated README file

elements/pf-date-picker/pf-calendar.ts
Arathy-s: Changed `ComposedEvent` to extend `Event` with `bubbles: true `and optionally `cancellable: true`.

elements/pf-date-picker/pf-calendar.ts
Arathy-s: Added ES6 private fields

Issue #2603, opened 2023-10-10
[core] Add default slot getter to SlotController


brianferry:
## Description of the requested feature

SlotController needs to be able to get the default slot and check if it has content.  We also want to be able to add functionality such as .isEmpty to the slot controller to check for during templating.

PR #2604, opened 2023-10-10
feat(core): adding isEmpty, hasSlotted w/ default slot

brianferry:
Linked Issue: https://github.com/patternfly/patternfly-elements/issues/2603

## What I did

1.  Added an `isEmpty` function that can be called either empty (`default` slot) or with a slot name.  This will return if the slot **does not** have content.
2. `hasSlotted` now returns default slot if not slot is provided.


core/pfe-core/controllers/slot-controller.ts
```diff
@@ -41,15 +41,15 @@ function isObjectConfigSpread(config: ([SlotsConfig] | (string | null)[])): conf
  * for the default slot, look for direct children not assigned to a slot
  */
 const isSlot =
-  <T extends Element = Element>(n: string | typeof SlotController.anonymous) =>
+  <T extends Element = Element>(n: string | typeof SlotController.default) =>
```
bennypowers, line 44: this is a breaking change, so I suggest something like

```js
static default = new Symbol('default slot');
static anonymous = this.default;
```

core/pfe-core/controllers/slot-controller.ts
```diff
@@ -142,13 +141,31 @@ export class SlotController implements ReactiveController {
...
+   * @returns {Boolean}
+   */
+
+  isEmpty(slotName?: string): boolean {
+    if (!slotName) {
+      return !this.#nodes.get(SlotController.default)?.hasContent ?? true;
+    } else {
+      return !this.#nodes.get(slotName)?.hasContent ?? true;
+    }
+  }
```
bennypowers, line 166:
```suggestion
  /**
   * Returns a boolean statement of whether or not any of those slots exists in the light DOM.
   *
   * @param names The slot names to check.
   * @example this.hasSlotted('header');
   */
  hasSlotted(...names: (string | null | undefined)[]): boolean {
    const { anonymous } = SlotController;
    const slotNames = Array.from(names, x => x == null ? anonymous : x);
    if (!slotNames.length) {
      slotNames.push(anonymous);
    }
    return slotNames.some(x => this.#nodes.get(x)?.hasContent ?? false);
  }

  /**
   * Whether or not all the requested slots are empty.
   *
   * @param  slots The slot name.  If no value is provided, it returns the default slot.
   * @example this.isEmpty('header', 'footer');
   * @example this.isEmpty();
   * @returns {Boolean}
   */
  isEmpty(...names: (string | null | undefined)[]): boolean {
    return !this.hasSlotted(...names);
  }

```

core/pfe-core/controllers/slot-controller.ts
```diff
@@ -41,15 +41,16 @@ function isObjectConfigSpread(config: ([SlotsConfig] | (string | null)[])): conf
...
+  <T extends Element = Element>(n: string | typeof SlotController.default) =>
     (child: Element): child is T =>
-        n === SlotController.anonymous ? !child.hasAttribute('slot')
+        n === SlotController.default ? !child.hasAttribute('slot')
       : child.getAttribute('slot') === n;
 
 export class SlotController implements ReactiveController {
-  public static anonymous = Symbol('anonymous slot');
+  public static default = Symbol('default slot');
+  public static anonymous = this.default;
```
bennypowers, line 51:
```suggestion
  /** @deprecated use `default` */
  public static anonymous = this.default;
```

.changeset/cold-cars-relate.md
```diff
@@ -0,0 +1,8 @@
+---
+"@patternfly/pfe-core": minor
+---
+
+`SlotController`: 
+
+- Add `isEmpty` method to check if a slot is empty.  If no slot name is provided it will check the default slot. (#2603)
+- `hasSlotted` method now returns default slot if no slot name is provided. (#2603)
```
bennypowers, line 8:
```suggestion
`SlotController`: Add `isEmpty` method to check if a slot is empty.  If no slot name is provided it will check the default slot. (#2603)
`SlotController`: `hasSlotted` method now returns default slot if no slot name is provided. (#2603)
```

bennypowers (approved):

PR #2606, opened 2023-10-17
feat(text-input): validation attrs

bennypowers:
## What I did

1. forward `pattern` to input
2. add `helper-text`
3. add `error-text`
4. add `validate-on`
5. add `--pf-c-form-control__error-text--m-status--Color` (NOT PRESENT in react version)
6. fix form state restore.
7. 
## Testing Instructions

1. check the demo
2. fill out a field and refresh - state should be restored.

## Notes to Reviewers

1. see http://v4-archive.patternfly.org/v4/components/text-input/design-guidelines#errors-and-validation
2. please critique my naming / api surface


elements/pf-text-input/pf-text-input.ts
```diff
@@ -213,48 +228,78 @@ export class PfTextInput extends LitElement {
...
              ?required="${this.required}"
              aria-label="${this.#derivedLabel}"
-             placeholder="${ifDefined(this.placeholder)}"
              type="${ifDefined(this.type)}"
-             .value="${this.value}"
              style="${ifDefined(this.customIconUrl && styleMap({
                backgroundImage: `url('${this.customIconUrl}')`,
                backgroundSize: this.customIconDimensions,
              }))}">
+        <span id="helper-text" ?hidden="${!this.helperText ?? valid}">${this.helperText}</span>
```
eyevana, line 248: If you specify helper text and error text, should only the error text show if the field is invalid? They both display at the moment. I couldn't find an example that had both in the PF4 docs either.

eyevana (approved):
Minor changes. Otherwise, looks good! :)

elements/pf-text-input/pf-text-input.ts
bennypowers: Maybe they should both show? Sure the helper applies in all cases. 
Let's merge and fix before release it it's a problem

PR #2607, opened 2023-10-18
fix(core): logger.debug, info

bennypowers:
## What I did

1. moves debug logs from SlotController to debug

Technically breaking but NBD IMO


brianferry (approved):
Lets Generate The Messages

PR #2630, opened 2023-10-19
fix(avatar)!: remove BaseAvatar

### LINKED ISSUE #2349: BaseAvatar properties should be abstract

### LINKED ISSUE #2575: Abstract Tabs functionality from BaseTabs to a Controller
Abstract tabs functionality out of BaseTabs to a Controller.
```[tasklist]
### Tasks
- [x] determine if we can reasonably replace the `aria-disabled` facility from pf-tab with a custom `disabled` attr
```
Update: After discussion between @nikkimk  and I we determined that we still need the `aria-disabled`

### LINKED ISSUE #2617: Remove BaseAvatar

bennypowers:
Closes #2617
Closes #2349
Closes #2414 

## What I did

1. remove BaseAvatar

## Testing Instructions

1. compare dp to prod demos

## Notes to Reviewers

1. Don't merge until the non-breaking items in the review queue are merged and published as 2.5.0


elements/pf-avatar/pf-avatar.css
```diff
@@ -1,3 +1,17 @@
+:host {
+  display: contents;
```
zeroedin, line 2: setting `display: contents` causes the describe`without src attr` test to fail.  This is because an element set to display contents will always return 0 for offsetWidth or clientWidth as it doesn't produce a box. 

> These elements don't produce a specific box by themselves. They are replaced by their pseudo-box and their child boxes. Please note that the CSS Display Level 3 spec defines how the contents value should affect "unusual elements" — elements that aren't rendered purely by CSS box concepts such as replaced elements. See [Appendix B: Effects of display: contents on Unusual Elements](https://drafts.csswg.org/css-display/#unbox) for more details.

https://developer.mozilla.org/en-US/docs/Web/CSS/display#contents

Example: https://codepen.io/zeroedin/pen/ExrwgMR

```Suggestion
  display: inline-block;
```

bennypowers: yeah more and more i'm thinking that we should avoid `:host { display: contents; }`, and maybe calculate a private css var in resize observer instead

PR #2632, opened 2023-10-19
fix(label)!: remove BaseLabel

### LINKED ISSUE #2620: Remove BaseLabel
Affects `<rh-tag>`

bennypowers:
Closes #2620

## What I did

1. remove baselabel

## Testing Instructions

1. compare dp to prod demos

## Notes to Reviewers

1. see https://github.com/RedHat-UX/red-hat-design-system/pull/1279
2. this pr causes label to remove itself on close if the event isn't cancelled. is that desired?
3. don't merge until non-breaking changes are released in 2.5.0


zeroedin:
> 2\. this pr causes label to remove itself on close if the event isn't cancelled. is that desired?

I take it you are making reference to whether or not the close button should do anything internally? 

Patternfly does not provide remove functionality directly and needs wired up, removal of the element `.remove()` is most likely preferred as it is already triggering the `LabelCloseEvent` that implementation can use to do whatever it is they desire on the click.

zeroedin:
> > 2. this pr causes label to remove itself on close if the event isn't cancelled. is that desired?
> 
> I take it you are making reference to whether or not the close button should do anything internally?
> 
> Patternfly does not provide remove functionality directly and needs wired up, removal of the element `.remove()` is most likely preferred as it is already triggering the `LabelCloseEvent` that implementation can use to do whatever it is they desire on the click.

And I now see the change in the changeset that instead provides an example of how to cancel the event, reverted commit to change that behavior.  My bad.

zeroedin (approved):
LGTM

PR #2633, opened 2023-10-19
fix(code-block)!: remove BaseCodeBlock

### LINKED ISSUE #2613: Remove BaseCodeBlock

bennypowers:
Closes #2613

## What I did

1. remove basecodeblock

## Testing Instructions

1. dp vs prod demos

## Notes to Reviewers

1. @eyevana check cp-elements
2. don't merge until all pending non-major prs are merged **and** released as 2.5.0


eyevana:
@bennypowers just checked and cpelements isn't extending any of the base classes from pfe. You're good to remove.

zeroedin (commented):
Issue exists in `main` as well, but slotted clipboard copy adds a text input field which displays an additional border next to the button, creates a dark border that doesn't exist in PFv4.   
Also strange to tab into it.  Assuming this field should be set hidden, which is what happens when clipboard-copy is set to inline.  However that changes the button styles

image

When set inline:

image

PR #2641, opened 2023-10-24
docs: update react integration

anujsingla:
## What I did

1.

## Testing Instructions

1.

## Notes to Reviewers

1.


docs/framework-integration/react.md
```diff
@@ -10,356 +10,287 @@ tags:
...
+- Github repo
+- Why does React need special treatment?
+- Recap
+
+Each section will show you exactly what you need to do with code snippets.
 {% endband %}
 
 {% band header="Initial setup" %}
-  Import React and ReactDOM at the top of the `index.js` file in the `/src/` directory.
+Create a new React project with [Vite](https://vitejs.dev/guide/#scaffolding-your-first-vite-project) by running this command
```
zeroedin, line 31: I'm not sure we'd want to focus on a single bundle tool here.  If anything maybe we should have this section contain a `pf-tabs` and give instructions for the vanilla and other popular tools.

bennypowers: Agreed, we should start from import

We will have to document various gotchas and workarounds with all the wonderful, faultless, and totally rad opinionated and forward-breaking tooling that's grown up around react, not unlike a festering pustule...

But we'll have to do that in a separate page

anujsingla: @zeroedin @bennypowers 
Regarding multiple tools -> 
I think we can add one note to the document something like this:
Note: We have provided an example using the Vite tool. You can also try using other tools, such as Webpack.
What do you think?

Regarding various examples of components:
We can provide framework examples for each component, similar to the popover component page: https://patternflyelements.org/components/popover/
This allows users to see how a component can be used in different frameworks.

bennypowers (approved):
Thank you @anujsingla! It took awhile but we finally got this merged!

PR #2655, opened 2023-12-21
fix!: v3 deprecations

bennypowers:
## What I did

1. remove deprecatedCustomEvent
2. remove framework test wrapper fixtures
3. remove icons from elements package
4. remove render functions from DocsPage
5. fix the tokens in accordion

## Notes to Reviewers

In reviewing this PR, please take a look at the wds config in https://github.com/RedHat-UX/red-hat-design-system/pull/1420/files/7f8e1bc5c1054e55083a0d50582d0bf0a6d56ff4..bfa209ba48e697ad4c78df20d285ff55518a2bbe, as a sample "downstream" usage of these changes


tools/pfe-tools/dev-server/index.html
```diff
@@ -36,7 +36,6 @@
     import 'element-internals-polyfill';
     import '@patternfly/elements/pf-button/pf-button.js';
     import '@patternfly/elements/pf-card/pf-card.js';
```
zeroedin, line 38: These imports cause issues for the demo's when the component uses them as a dependency or the demo also tries to import them:

```
Uncaught DOMException: Failed to execute 'define' on 'CustomElementRegistry': the name "pf-button" has already been used with this registry
    at http://localhost:8001/node_modules/@lit/reactive-element/development/decorators/custom-element.js:27:24
    at __decorateClass (http://localhost:8001/elements/pf-button/pf-button.ts:8:57)
    at http://localhost:8001/elements/pf-button/pf-button.ts:68:12
```

Should we also load all the elements in a import map and not rely on web test server to rewrite the import paths?  Assuming that I understand the full issue.

elements/pf-progress-stepper/pf-progress-step.ts
```diff
@@ -13,9 +13,9 @@ import { InternalsController } from '@patternfly/pfe-core/controllers/internals-
 import style from './pf-progress-step.css';
 
 const ICONS = new Map(Object.entries({
-  success: { icon: 'circle-check' },
-  danger: { icon: 'circle-exclamation' },
-  warning: { icon: 'triangle-exclamation' },
+  success: { icon: 'check-circle' },
+  danger: { icon: 'exclamation-circle' },
+  warning: { icon: 'exclamation-triangle' },
```
zeroedin, line 18: Because we are moving to the 'non-combined' set of icons which is purely font awesome v5, then these icons names no longer will work.

elements/pf-label/demo/patternfly.html
zeroedin, line 1: Because we are moving to the 'non-combined' set of icons which is purely font awesome v5, then these icons names no longer will work.

elements/pf-label/demo/pf-label.html
zeroedin, line 1: same as previous

elements/pf-table/demo/expandable-rows-compound.html
zeroedin, line 1: there was no direct 1:1 icon here in fa v5.   `code-branch` was the closest to the idea i could find.

elements/pf-button/demo/pf-button.html
zeroedin, line 1: same, icons names changed fa6 -> fa5

tools/pfe-tools/dev-server/index.html
bennypowers: We should either use import maps or node resolution but not both

bennypowers: Note to self: let's see how far we can get by just using the `nodemodules` jspm provider in the dev server, or equivalent

bennypowers: Maybe this would be helpful

https://modern-web.dev/docs/dev-server/plugins/import-maps/#multiple-import-maps

bennypowers: We might generate the map with the jspm generator outside of the config function, inside the user's dev server config module

bennypowers: PTAL @zeroedin

bennypowers:
I apparently fixed the import map generator plugin. I then ran roughshod over the demo files because the import map plugin requires all top-level modules be inlined.

We'll have to test this with rhds (and maybe cp too, cc @eyevana)

bennypowers:
Ok this appears to work in rhds with the following dev server config


```js
// @ts-check
import { pfeDevServerConfig } from '@patternfly/pfe-tools/dev-server/config.js';
import { glob } from 'glob';

export const litcssOptions = {
  include: (/** @type{string[]}*/(/** @type{unknown}*/([
    /elements\/rh-[\w-]+\/[\w-]+\.css$/,
    /lib\/.*\.css$/,
  ]))),
  exclude: /lightdom/,
};

const libImports =
  await glob('./lib/**/*.js', { ignore: ['**/test/**'] })
    .then(files => files.map(spec => [`@rhds/elements/${spec}`, `./${spec}`]))
    .then(Object.fromEntries);

const elementImports =
  await glob('./elements/**/*.js', { ignore: ['**/test/**'] })
    .then(files => files.map(spec => [`@rhds/elements/${spec.replace('elements/', '')}`, `./${spec}`]))
    .then(Object.fromEntries);


export default pfeDevServerConfig({
  litcssOptions,
  tsconfig: 'tsconfig.json',
  importMapOptions: {
    providers: {
      '@patternfly/icons': 'nodemodules',
      '@patternfly/elements': 'nodemodules',
      '@patternfly/pfe-tools': 'nodemodules',
      '@patternfly/pfe-core': 'nodemodules',
    },
    inputMap: {
      imports: {
        ...libImports,
        ...elementImports,
      },
    },
  },
  middleware: [
    /** redirect requests for lightdom css to /elements */
    function(ctx, next) {
      const match = ctx.path.match(/^\/components\/(?<slug>[-\w]+)\/(?<path>.*)\.css$/);
      if (match) {
        const { slug, path } = /** @type{{ slug: string; path: string }} */ (match.groups);
        ctx.redirect(`/elements/rh-${slug}/${path}.css`);
      }
      return next();
    },
    /** redirect requests for /assets/* css to /docs/assets/prism.css */
    function(ctx, next) {
      if (ctx.path.startsWith('/assets/')) {
        ctx.redirect(`/docs${ctx.path}`);
      }
      return next();
    }
  ]
});

```

elements/pf-accordion/acc-lightdom.css
```diff
@@ -0,0 +1 @@
+/* hi */
```
zeroedin, line 1: `/* hello what are you doing here? */`

Does this file need removed?

zeroedin (approved):
So good.  🔥 

Late Guests Timidly Mingle

elements/pf-accordion/acc-lightdom.css
bennypowers: Thanks for catching that! this was left over from local testing

PR #2661, opened 2024-01-18
fix: remove color theme token, replace with PFv4 tokens

### LINKED ISSUE #2660: [bug] incorrect PFv4 token usage
bennyp  from chat: 
there's also this --pf-theme--color--ui-base token in some of the files which i couldn't find on upstream

this is based off the listbox PR: [menu-dropdown-listbox](https://github.com/patternfly/patternfly-elements/pull/2570)

zeroedin:
## What I did

1.  Remove color theme token, replace with PFv4 tokens in a best guess.  

Closes #2660

## Testing Instructions

1.

## Notes to Reviewers

1.  If I could find one that seemed to fit the use case I swapped it otherwise used the global version.   Not ideal but don't think our use cases always match perfect.  PTAL, correct my mistakes were plausible.

### DIFF
--- a/elements/pf-background-image/demo/sibling-content.html
+++ b/elements/pf-background-image/demo/sibling-content.html
@@ -1,4 +1,3 @@
-<script type="module" src="pf-background-image.js"></script>
 
 <pf-background-image
   src="/components/background-image/demo/pfbg.jpg"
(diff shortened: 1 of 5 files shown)


bennypowers (approved):
tyvm

PR #2671, opened 2024-01-19
fix(dropdown): trigger button should never be disabled

nikkimk:
## What I did

1. Removed disabled from toggle button.

## Testing Instructions

1. In Dropdown docs -> disabled, use a keyboard to navigate to toggle the dropdown. 
2. Verify the menuitems can be read, by a screenreader but are disabled.

## Notes to Reviewers

1. [WAI-ARIA recommends](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols) elements of a larger composite widget remain focusable.

### DIFF
--- a/elements/pf-dropdown/pf-dropdown.ts
+++ b/elements/pf-dropdown/pf-dropdown.ts
@@ -87,6 +87,9 @@ export class PfDropdown extends LitElement {
 
   render() {
     const { expanded, anchor, alignment } = this.#toggle ?? {};
+    /**
+     * Note: For accessibility compliance, trigger button should never be disabled.
+     */
     const { disabled } = this;
     return html`
     <div class="${classMap({ expanded, [anchor ?? '']: !!anchor, [alignment ?? '']: !!alignment })}"
@@ -96,8 +99,7 @@ export class PfDropdown extends LitElement {
             name="trigger"
             @slotchange="${this.#setTriggerElement}">
         <pf-button id="default-button"
-                   variant="control"
-                   ?disabled="${disabled}">
+                   variant="control">
           Dropdown
           <svg viewBox="0 0 320 512" fill="currentColor" aria-hidden="true" width="1em" height="1em">
             <path d="M31.3 192h257.3c17.8 0 26.7 21.5 14.1 34.1L174.1 354.8c-7.8 7.8-20.5 7.8-28.3 0L17.2 226.1C4.6 213.5 13.5 192 31.3 192z"></path>
@@ -107,7 +109,7 @@ export class PfDropdown extends LitElement {
       <pf-dropdown-menu id="menu"
                         part="menu"
                         class="${classMap({ show: expanded })}"
-                        ?disabled="${this.disabled}"
+                        ?disabled="${disabled}"
                         @keydown="${this.onKeydown}"
                         @click="${this.#handleSelect}">
         <slot></slot>


bennypowers:
http://v4-archive.patternfly.org/v4/components/dropdown#disabled-toggles demonstrates a disabled trigger button, as such we must align with upstream behaviours. if this is an accessibility issue upstream, we should file there

bennypowers (changes requested):
please add unit tests

zeroedin:
If a keyboard user is able to open the menu, then a mouse click should do the same.  BOTH shouldn't be able to select a sub item.

nikkimk:
Blocked by #2716

zeroedin (approved):
(oops approved wrong PR)

nikkimk:
> If a keyboard user is able to open the menu, then a mouse click should do the same. BOTH shouldn't be able to select a sub item.

@hellogreg @zeroedin and I decided that disabled menu should open for both keyboard and mouse but that menu items can't be clicked

nikkimk:
@bennypowers ptal

bennypowers (approved):

hellogreg:
FYI: looks like the disabled button and menu items are not giving an indication to VoiceOver/Safari or JAWS/Chrome that they are disabled:

disabled button not announcing dimmed in VoiceOver
disabled menu item not announcing dimmed in VoiceOver

The menu items _do_ announce "unavailable" to NVDA/Firefox.

bennypowers:
@nikkimk @hellogreg does [d8eaf8d](https://github.com/patternfly/patternfly-elements/pull/2671/commits/d8eaf8d18c304902476e4fac67d71a3dd35738e4) help?

elements/pf-dropdown/pf-dropdown-group.ts
```diff
@@ -15,17 +16,60 @@ export class PfDropdownGroup extends LitElement {
...
+   */
+  @property({ type: Boolean, reflect: true }) disabled = false;
+
   /**
    * The label for the group of dropdown items.
    */
   @property({ reflect: true }) label?: string;
 
+
+  #internals = InternalsController.of(this, { });
```
bennypowers, line 30:
```suggestion
  #internals = InternalsController.of(this);
```
If we're not availing ourselves of the options, we [shouldn't need to supply](https://github.com/patternfly/patternfly-elements/blob/1b8f5aebfe280570df1a863f11afe1d4c61c10da/core/pfe-core/controllers/internals-controller.ts#L66) an empty object.

elements/pf-dropdown/pf-dropdown-group.ts
```diff
@@ -15,17 +16,60 @@ export class PfDropdownGroup extends LitElement {
...
+   * Disable the dropdown
+   */
+  @property({ type: Boolean, reflect: true }) disabled = false;
+
   /**
    * The label for the group of dropdown items.
    */
   @property({ reflect: true }) label?: string;
 
+
```
bennypowers, line 29:
```suggestion
```
extra whitespace

elements/pf-dropdown/pf-dropdown-group.ts
```diff
@@ -15,17 +16,60 @@ export class PfDropdownGroup extends LitElement {
...
   }
+
+  /**
+   * handles focusing on an option:
+   * updates roving tabindex and active descendant
+   */
+  #onMenuitemFocusin(event: FocusEvent) {
+    if (this.disabled) {
+      event.preventDefault();
+      event.stopPropagation();
```
bennypowers, line 58: do we need to stop propagation? is preventDefault enough?

I think we should also leave a "why comment" here and below, to prevent confusion in the future, since the reader may not immediately grok that disabled dropdowns are still permitted to open

elements/pf-dropdown/pf-dropdown.ts
```diff
@@ -65,7 +62,7 @@ export class PfDropdown extends LitElement {
   static override readonly shadowRootOptions: ShadowRootInit = { ...LitElement.shadowRootOptions, delegatesFocus: true };
 
   /**
-   * Disable the dropdown trigger element
+   * Disable the dropdown
```
bennypowers, line 65:
```suggestion
   * When disabled, the dropdown can still be toggled open and closed via keyboard, but menu items cannot be activated.
```

The [upstream description for `isDisabled` prop](http://v4-archive.patternfly.org/v4/components/dropdown/#dropdowntoggle) is not particularly relevant to us, so I decided to wing it. How does this read to you?

elements/pf-dropdown/pf-dropdown.ts
```diff
@@ -99,29 +96,27 @@ export class PfDropdown extends LitElement {
...
+            class="${classMap({ show: expanded })}"
+            ?hidden="${!this.expanded}"
+            @focusout="${this.#onMenuFocusout}"
+            @keydown="${this.#onMenuKeydown}"
+            @click="${this.#onSelect}">
+        <pf-dropdown-menu id="menu" part="menu" ?disabled="${disabled}">
+          <slot></slot>
+        </pf-dropdown-menu>
+      </slot>
+    </div>`;
```
bennypowers, line 119:
```suggestion
      <div class="${classMap({ expanded,
                               [anchor ?? '']: !!anchor,
                               [alignment ?? '']: !!alignment })}"
           style="${styleMap(styles)}"
           @slotchange="${this.#onSlotchange}">
        <slot name="trigger"
              @keydown="${this.#onButtonKeydown}"
              @click="${() => this.toggle()}">
          <pf-button id="default-trigger"
                     variant="control"
                     icon="caret-down"
                     icon-set="fas">Dropdown</pf-button>
        </slot>
        <slot name="menu"
              ?hidden="${!this.expanded}"
              @focusout="${this.#onMenuFocusout}"
              @keydown="${this.#onMenuKeydown}"
              @click="${this.#onSelect}">
          <pf-dropdown-menu id="menu" part="menu" ?disabled="${disabled}">
            <slot></slot>
          </pf-dropdown-menu>
        </slot>
      </div>
    `;
```

- revert whitespace changes to match formatting in other files
- prevent wide line in classmap 
- remove the `show` class from the menu slot, as it's not needed in the element's css anymore

elements/pf-dropdown/pf-dropdown.css
```diff
@@ -71,20 +71,34 @@ pf-dropdown-menu,
 }
 
 :host([disabled]) :is(pf-dropdown-menu, ::slotted(pf-dropdown-menu)) {
```
bennypowers, line 73: should we use a shadow class on the container for these? WDYT?

elements/pf-dropdown/pf-dropdown-group.ts
```diff
@@ -15,17 +16,60 @@ export class PfDropdownGroup extends LitElement {
 
   static override readonly shadowRootOptions: ShadowRootInit = { ...LitElement.shadowRootOptions, delegatesFocus: true };
 
+  /**
+   * Disable the dropdown
+   */
+  @property({ type: Boolean, reflect: true }) disabled = false;
+
```
bennypowers, line 23: this doesn't appear on v4's group, so i wonder if we should use a context for this instead? or perhaps just a query selector?

bennypowers (changes requested):
Changes look really nice. A few points, mostly nits, but give a think about the disabled attr on group.

elements/pf-dropdown/context.ts
```diff
@@ -0,0 +1,7 @@
+import { createContextWithRoot } from '@patternfly/pfe-core/functions/context.js';
+
+export interface PfDropdownMenuContext {
+  disabled: boolean;
+}
+
+export const context = createContextWithRoot<boolean>(Symbol('pf-dropdown-menu-context'));
```
bennypowers, line 7:
```suggestion
export interface PfDropdownContext {
  disabled: boolean;
}

export const context = createContextWithRoot<PfDropdownContext>(Symbol('pf-dropdown-menu-context'));
```

elements/pf-dropdown/pf-dropdown-item.ts
```diff
@@ -81,23 +84,33 @@ export class PfDropdownItem extends LitElement {
...
+  @consume({ context, subscribe: true })
+  @property({ attribute: false })
+  private ctx?: PfDropdownMenuContext;
+
   #internals = InternalsController.of(this, { role: 'none' });
 
   /** @internal */
   @query('#item') menuItem!: HTMLElement;
 
+  get isDisabled() {
```
bennypowers, line 96:
```suggestion
  get disabled() {
```

bennypowers (approved):

elements/pf-dropdown/pf-dropdown-item.css
```diff
@@ -36,7 +36,8 @@
   display: none !important;
 }
 
-:host([disabled]) {
+:host([disabled]), 
+.disabled {
   pointer-events: none;
```
bennypowers, line 41: [Screencast from 2024-03-28 09-18-25.webm](https://github.com/patternfly/patternfly-elements/assets/1466420/b75dd09b-b2a1-49a4-bc12-3480af14c6d6)
That pointer-events: none can be problematic if there are interactive controls deeper than the disabled item

bennypowers (changes requested):

hellogreg (approved):
Passed the usual browser and screen reader tests this afternoon.

Lighted Grid, Tron Motorbike.

bennypowers (changes requested):
Oof GitHub left all these comments pending until now. 😐 Sorry

elements/pf-dropdown/pf-dropdown-menu.ts
```diff
@@ -26,6 +28,9 @@ export class PfDropdownMenu extends LitElement {
    */
   @property({ type: Boolean, reflect: true }) disabled = false;
 
+  @provide({ context })
```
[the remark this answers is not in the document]
bennypowers: resolved by 6caa845

nikkimk:
@bennypowers in the DP's disabled example, the second menu item has a pointer cursor.

hellogreg:
A little weirdness here with VoiceOver. The first time a menu item receives focus, VO announces, "[text], menu item." But on subsequent focus, it just reads the text.

Action announced as menu item

Action announced, but not menu item


This may not be a showstopper. Other menu button examples (e.g., WAI's [Navigation Menu Button](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-links/)) don't announce "menu item" with the initial text _at all_. And VoiceOver does ultimately indicate it is a menu item (as it does in the WAI example):

VoiceOver does indicate the option is a menu item


But the inconsistency might be worth investigating. If we're okay with launching it with this, I can open another issue afterward.

hellogreg:
Windows (Chrome/JAWS and Firefox/NVDA) and mobile (Android and iOS) working as expected.

PR #2672, opened 2024-01-21
fix(core): decouple listbox controller from DOM

bennypowers:
## What I did

1. removed the dispatchEvent calls from listbox controller


nikkimk (changes requested):
This is not working with a keyboard.

nikkimk:
> This is not working with a keyboard.

@bennypowers where are we at with this one?

bennypowers:
I think the goals of this pr were accomplished elsewhere. closing

PR #2674, opened 2024-01-22
docs: fix band shortcode

### LINKED ISSUE #2667: [docs] remove duplicate chip demos from docs page
## Description of the issue

The chip demo section appears twice on the [chip docs page](https://patternflyelements.org/components/chip/).

The page is currently structured like this, so that it starts over at "Default" after "Chip group with removable categories": 

- **Default** (first occurrence)
- Chip variants
    - Read-only
    - Overflow chip
    - Max-width
- Chip group
    - Chip group with categories
    - Chip group with removable categories
- **Default** (second occurrence)
- _...and so on..._
 
### Expected behavior
The demos should only appear once. Delete the duplicate group of demos.

bennypowers:
Closes #2667

## What I did

1. the band shortcode mistakenly repeated the content in place of the header. I fixed that
2. also removed .njk templates which are no longer in use


hellogreg (approved):
Leto, God-emperor, Takes Melange.

Nice work, Benny!

PR #2675, opened 2024-01-22
fix(chip): add toolbar role

### LINKED ISSUE #2666: chip: chip group should have toolbar role
## Description of the issue

Chip groups should have a `toolbar` role to communicate to assistive tech users that they have entered a group of controls

### Impacted component(s)

- `pf-chip`
- `pf-chip-group`

### Steps to reproduce

1. Go to [chip documentation](https://patternflyelements.org/components/chip/).
2. Enable screen reader (I initially used Mac VoiceOver with Safari).
3. Navigate to a chip group using the `tab` key.

### Expected behavior

The user receives some indication that they have entered a group of controls and should now use arrow keys to navigate within the chip group.

bennypowers:
Closes #2666

## What I did

1. added toolbar role to shadow container for chip group
2. some nitpicky formatting


hellogreg:
Running some tests in the preferred desktop browser screen reader combos...

**Windows Chrome and JAWS:** All good

**Mac Safari and VoiceOver:** All good

**Windows Firefox and NVDA:** Looks like there's a keyboard navigation issue in Windows and Mac Firefox, even before trying with a screen reader. This issue is also present in the current live version of `chip`, so I can make a separate issue for it this afternoon. (Basically, if you tab to a chip group and then try using the arrows, you can't navigate between the regular chips in that group. However, if you hit the arrow key enough times, you _will_ navigate to the "x more / see less" chip.)

That said, everything regarding chip that is keyboard navigable in FF looks to be working well with NVDA.

So, I guess what I'm saying is this issue looks good to me, but I can't really test FIrefox fully, because keyboard nav isn't totally working.

hellogreg:
Added the Firefox issue: #2682

PR #2676, opened 2024-01-22
fix(chip): layouts, max width tooltip

### LINKED ISSUE #2663: chip: tooltip for a long chip is not positioned correctly
### Impacted component(s)

- `pf-chip`

### Steps to reproduce

1. Go to [Chip Docs--> max-width](https://patternflyelements.org/components/chip/#max-width)
2. Hover on the really long tooltip.

### Expected behavior

Tooltip should pop up close to the chip text.

bennypowers:
Closes #2663 

## What I did

1. restore chip layout to align with pfv4. please don't revert this!
2. fix some css in plain icon button for the sake of above
3. fix an annoying dev server misconfiguration bug


elements/pf-chip/pf-chip.ts
```diff
@@ -55,24 +55,23 @@ export class PfChip extends LitElement {
...
-          <svg aria-hidden="true" fill="currentColor" viewBox="0 0 352 512">
-            <path d="M242.72 256l100.07-100.07c12.28-12.28 12.28-32.19 0-44.48l-22.24-22.24c-12.28-12.28-32.19-12.28-44.48 0L176 189.28 75.93 89.21c-12.28-12.28-32.19-12.28-44.48 0L9.21 111.45c-12.28 12.28-12.28 32.19 0 44.48L109.28 256 9.21 356.07c-12.28 12.28-12.28 32.19 0 44.48l22.24 22.24c12.28 12.28 32.2 12.28 44.48 0L176 322.72l100.07 100.07c12.28 12.28 32.2 12.28 44.48 0l22.24-22.24c12.28-12.28 12.28-32.19 0-44.48L242.72 256z"></path>
-          </svg>
-        </button>
+        <pf-button id="close-button"
+                   plain
+                   icon="close"
+                   icon-set="patternfly"
+                   label="${this.accessibleCloseLabel}"
+                   aria-describedby="chip-text"
```
nikkimk, line 72: This is an accessibility issue that @hellogreg and I reverted back to button because that was the way to solve cross-root aria issues. We need the `aria-describedby` to be a working IDRef to the chip text in order to be compatible with all assistive tech. This was the one issue that you and I had discussed last week that if we reverted all the other cross-root aria stuff would not be deal breakers.

elements/pf-chip/test/pf-chip.spec.ts
```diff
@@ -45,19 +45,21 @@ describe('<pf-chip>', async function() {
...
     describe('pressing Tab', function() {
       beforeEach(tab);
-      it('should focus', function() {
-        expect(activeElement(element)?.getAttribute('aria-label'))
-          .to.equal(element.accessibleCloseLabel);
+      it('should focus', async function() {
+        const snapshot = await a11ySnapshot();
+        expect(snapshot.children?.at(0)?.name).to.equal(element.accessibleCloseLabel);
+        expect(snapshot.children?.at(0)?.focused).to.be.true;
       });
```
nikkimk, line 64: We need a test for accessible description as well.

elements/pf-chip/pf-chip.css
```diff
@@ -1,106 +1,108 @@
...
+  bottom: 0;
+  left: 0;
+  content: "";
+  border: var(--pf-c-chip--before--BorderWidth) solid var(--pf-c-chip--before--BorderColor);
+  border-radius: var(--pf-c-chip--before--BorderRadius);
+}
+
+span {
+  overflow: hidden;
   text-overflow: ellipsis;
```
nikkimk, line 65: This is not working in DP docs.

nikkimk (changes requested):

elements/pf-chip/pf-chip.ts
bennypowers: 24e5ca26 should hopefully fix this

elements/pf-chip/test/pf-chip.spec.ts
bennypowers: I'm not sure we can test for this, the snapshot doesn't surface the aria description on the button's container
here's how it looks in ff's ax tree inspector:
image

elements/pf-chip/pf-chip.css
bennypowers: this should be resolved by 49f3b022

elements/pf-chip/test/pf-chip.spec.ts
nikkimk: I added @hellogreg as a reviewer as well to see if this resolves the Win/FF and Win/Edge testing issues he's had with chip.

bennypowers: ok, if not, since it appears we won't be able to apply the description across roots, that means we can't use pf-button until browser vendors ship a solution for x-root aria. This creates two maintenance problems:
1. duplicating pf-button here
2. what to do when browser support improves

for (1), i'd like to suggest we give some thought to creating some pf-button abstractions that can sit in a shadow template, maybe an element-position directive that applies styles based on attributes? barring that, to see whether or not we could just apply the shadow stylesheet from pf-button here
for (2), I think we should leave a comment in the template with a note to check back in q4, depending on how the browser vendors get on

WDYT?

hellogreg: So, the good news is that chips in this deploy preview are navigable via keyboard in all browsers.

The not-so-good news is what's announced when screen readers are enabled.

Typically (e.g., in the live version of chip), the chip's contents get announced:

Screencap of live chip with keyboard focus

However, in the current version of this deploy preview, screen readers just say "close, button":

Screencap of deploy preview chip with focus

nikkimk: @hellogreg what's the situation with a single chip? I'm wondering if the issue is the `aria-describedby` not being identified.

hellogreg: Yep, you're right, @nikkimk . Looks like the description isn't being identified for any chips.

bennypowers: this should be ready for another review

nikkimk (approved):
livid grocer throws melons

hellogreg:
Checked the chip groups with the major desktop browser/screen reader combos, and looks like both arrow nav and descriptions are working. In other words, Luigi grabbed twenty mushrooms.

PR #2686, opened 2024-01-24
fix(tools): dev server load ts sources

bennypowers:
## What I did

1. make sure dev server loads typescript sources
2. make sure dev server body fills screen height


nikkimk (approved):
Learned geologists test minerals

PR #2687, opened 2024-01-24
fix(tools): load ts sources from core

bennypowers:
## What I did

1. more whoopsies from the dev server

### DIFF
--- a/tools/pfe-tools/dev-server/plugins/pfe-dev-server.ts
+++ b/tools/pfe-tools/dev-server/plugins/pfe-dev-server.ts
@@ -123,6 +123,24 @@ function getRouter(options: PfeDevServerInternalConfig) {
       } else {
         return next();
       }
+    })
+
+    .get(`/${elementsDir}/:element/:splatPath*.(css|js)`, async function(ctx, next) {
+      ctx.response.etag = performance.now().toString();
+      return next();
+    })
+
+    // Redirect `core/pfe-core/controllers/thingy.js` to `core/pfe-core/controllers/thingy.ts` for requests not previously handled
+    .get(`/core/pfe-core/:splatPath*.js`, async ctx => {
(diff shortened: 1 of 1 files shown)


tools/pfe-tools/dev-server/plugins/pfe-dev-server.ts
```diff
@@ -130,6 +130,11 @@ function getRouter(options: PfeDevServerInternalConfig) {
       return next();
     })
 
+    // Redirect `core/pfe-core/controllers/thingy.js` to `core/pfe-core/controllers/thingy.ts` for requests not previously handled
```
zeroedin, line 133: is there anything we should be worried about downstream here?  eg `lib/elements` in RHDS?

zeroedin: follow up, should we allow for optional routes?

bennypowers: downstream can do their own middleware i think

zeroedin: You are correct and we are already doing as such:

```
export default pfeDevServerConfig({
  litcssOptions,
  tsconfig: 'tsconfig.json',
  middleware: [
    /** redirect requests for lightdom css to /elements */
    function(ctx, next) {
      const match = ctx.path.match(/^\/components\/(?<slug>[-\w]+)\/(?<path>.*)\.css$/);
      if (match) {
        const { slug, path } = /** @type{{ slug: string; path: string }} */ (match.groups);
        ctx.redirect(`/elements/rh-${slug}/${path}.css`);
      }
      return next();
    },
    /** redirect requests for /assets/* css to /docs/assets/prism.css */
    function(ctx, next) {
      if (ctx.path.startsWith('/assets/')) {
        ctx.redirect(`/docs${ctx.path}`);
      }
      return next();
    }
  ]
});
```

PR #2689, opened 2024-01-31
feat(tools): css import caching transformer

bennypowers:
## What I did

1. check all source files before transforming their css imports:
   - WHEN we find an imported css file which has more than one importee,
   - THEN don't let those importees inline the css file contents, instead emit a `.css.js` module

## Testing Instructions

1. build
2. check pf-button.js
3. check pf-chip.js


nikkimk (approved):
lofty goal totally met

PR #2699, opened 2024-03-03
fix(tabs)!: tabscontroller refactor

### LINKED ISSUE #2695: TabsController should not make assumptions about host's DOM
In `TabsController`'s `#rebuild()` method:

```ts
const tabSlot = this.#element.shadowRoot?.querySelector<HTMLSlotElement>('slot[name=tab]');
const panelSlot = this.#element.shadowRoot?.querySelector<HTMLSlotElement>('slot:not([name])');
```

In order that tabs controller be applicable to any tabs implementation, these should be replaced with getter functions in the options.

We should also consider removing the dependency on expand / disabled events and the `active` property in a similar manner

bennypowers:
Closes #2695 
## What I did

Total separation between RTIC and tabs state. Disabled tabs are still able to be focused via RTIC, just that they can't change the Tab's active tab state.

Remove most of tabscontroller's logic in favour of context, rtic.

- Rename TabsController > TabsAriaController. It still has some helper functions. WDYT, @zeroedin?
- PfTabs(host) - context for active index, cascading stuff like variant, box etc
- PfTab(host) - updates own template accd to context. to avoid breaking css selectors, maybe reflect those context props for now. at least investigate it before removing the attrs on the host.

## Testing Instructions

1. units, usability testing


elements/pf-tabs/pf-tab.ts
```diff
@@ -69,84 +72,102 @@ import styles from './pf-tab.css';
...
-    this.active = true;
+    return this.dispatchEvent(new TabExpandEvent(this));
   }
 
-  private _activeChanged(oldVal: boolean, newVal: boolean) {
-    if (oldVal !== newVal && newVal === true) {
-      this.dispatchEvent(new TabExpandEvent(this));
+  private _activeChanged(old: boolean) {
+    if (this.active && !old) {
+      this.#activate();
```
zeroedin, line 165: I believe we need to set the `#internals.ariaSelected` state here as well, but only if its not disabled.

```Suggestion
  private _activeChanged(old: boolean) {
    if (this.active && !old) {
      this.#activate();
    }
    if (this.active && !this.disabled) {
      this.#internals.ariaSelected = 'true';
    } else {
      this.#internals.ariaSelected = 'false';
    }
```

zeroedin:
Should we now say that if `<pf-tab aria-disabled="true">` is set that this has the same function as `<pf-tab disabled>` and update the component accordingly?  /cc @nikkimk

elements/pf-tabs/pf-tab.ts
bennypowers: @nikkimk I'm eager to hear your input on this

bennypowers:
The idea is that users should not set aria attrs on any tab elements. They should only and always use `disabled` when they need that feature

core/pfe-core/controllers/tabs-aria-controller.ts
```diff
@@ -0,0 +1,123 @@
...
+      this.#element = host;
+    } else {
+      const element = options.getHTMLElement?.();
+      if (!element) {
+        throw new Error('TabsController must be instantiated with an HTMLElement or a `getHTMLElement()` option');
+      }
+      this.#element = element;
+    }
+    (this.#host = host).addController(this);
+    this.#element.addEventListener('slotchange', this.#onSlotchange);
```
zeroedin, line 63: Can you explain how this option of `getHTMLElement` would be different in implementation from `host` given we are attaching a `slotchange` event to the element?  IIRC `slotchange` would only work given the element is a custom element with a shadowroot so wouldn't that also just be a host?  Is this a way to instantiate this in a parent element outside of that target?

```
<outer-parent> (has tabs-aria-controller setup with getHTMLElement targetting inner-tabs)
  <inner-tabs>
     <im-a-tab></im-a-tab>
     <im-a-panel></im-a-panel>
     ...
  <inner-tabs>
</outer-parent>
```

bennypowers: Here's an example from haunted, a react-a-like ce framework

https://github.com/matthewp/haunted/pull/239/files#diff-c5df8cacedd70d19bd53b47ac1cf3de080ba8b014f4f8c9ba73f7b70fecdb305

Similarly for full-menlo react: https://lit.dev/docs/frameworks/react/#usecontroller
And https://github.com/lit/lit/blob/main/packages/react/src/use-controller.ts

TL;DR: reactive controller hosts don't have to be custom elements, they just have to be POJOs which implement a particular interface. If we can avoid requiring our controllers to be hosted by elements, our controllers will be of use in more places

bennypowers: So why slot change?

Maybe the controller host object (which owns the update lifecycle) isn't the same as the rendering custom element

Maybe they're are Very Important Reasons for someone to reimplement our ces (like... It's embedded in their dashboard rendering framework), and they use their own unbranded ces, which aren't reactive elements, but maybe we could create host objects which play puppeteer with the real ces.

elements/pf-tabs/pf-tab.ts
zeroedin: Another thing to note on this same code chunk is looks like we are no longer logging a warning about attempted setting of `disabled` tabs to active.  Oversight or is there reasoning for removal?

image

elements/pf-tabs/context.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { createContextWithRoot } from '@patternfly/pfe-core/functions/context.js';
+
+export interface PfTabsContext {
+  activeTab: PfTab | undefined;
+  box: 'light' | 'dark' | null; fill: boolean;
+  vertical: boolean;
+  manual: boolean;
+  borderBottom: 'true' | 'false';
+}
```
zeroedin, line 11: missing line return

```Suggestion
export interface PfTabsContext {
  activeTab: PfTab | undefined;
  box: 'light' | 'dark' | null; 
  fill: boolean;
  vertical: boolean;
  manual: boolean;
  borderBottom: 'true' | 'false';
}
```

core/pfe-core/controllers/tabs-aria-controller.ts
```diff
@@ -0,0 +1,123 @@
+import type { ReactiveController, ReactiveControllerHost } from 'lit';
+
+import { Logger } from '@patternfly/pfe-core/controllers/logger.js';
+
+export interface TabsControllerOptions<Tab, Panel> {
```
zeroedin, line 5: nit:  might consider renaming options to `TabsAriaControllerOptions` for consistency

elements/pf-tabs/pf-tab.ts
bennypowers: 49f7ec92

core/pfe-core/controllers/tabs-aria-controller.ts
bennypowers: 2515c32f

elements/pf-tabs/pf-tab.ts
bennypowers: c8b39e60 cc @nikkimk

zeroedin (approved):
Given resolution of last remaining open comment 

Lifted Granite Tall Mountains

core/pfe-core/controllers/tabs-aria-controller.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  hostConnected() {
+    this.#mo.observe(this.#element, { attributes: false, childList: true, subtree: false });
+    this.#onSlotchange();
+  }
+
+  hostUpdated() {
+    for (const [tab, panel] of this.#tabPanelMap) {
+      if (!panel.hasAttribute('aria-labelledby')) {
```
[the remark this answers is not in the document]
bennypowers: Users shouldnt set aria on their own

PR #2701, opened 2024-03-05
fix(select): hr styles

### LINKED ISSUE #2700: `<pf-select>`: hr dividers should be styled
pf-select with slotted hr should look like http://v4-archive.patternfly.org/v4/components/select/

bennypowers:
Closes #2700 
## What I did

1. implement styles for pf select hr separators

## Testing Instructions

1. compare to http://v4-archive.patternfly.org/v4/components/select/


zeroedin:
> > `pf-select-group` has a border bottom creating the illusion of a double border when the `hr` appears after it
> 
> Can we just consider that an antipattern, then?

So if I understand correctly, `<hr>` is slotted with `<pf-option>` but when an `<pf-option-group>` is used you don't need to add the `<hr>` manually.

I feel like the use case is different enough to have a different expectation on the implementation.  `pf-option-group` is an explicit pattern and grouping, where as the `<hr>` between `<pf-option>` is an additive visual reference of a perceived separation, but technically all the options are still the same "group".

nikkimk (commented):
Luxurious garden topiary marvel

bennypowers:
@nikkimk don't forget the gh approval, plznthx

PR #2703, opened 2024-03-06
fix(dropdown): slotted trigger

bennypowers:
## What I did

1. added a slot for trigger buttons (already documented)
2. moved event listeners and aria relationships to a div which now wraps that slot

## Testing Instructions

1. SR usability, ax tree. check the `custom-trigger` demo

## Notes to reviewers
This is required in order to replace the docs version `<select>` with pf-dropdown


bennypowers:
@hellogreg ptal at https://deploy-preview-2703--patternfly-elements.netlify.app/components/dropdown/demo/custom-trigger/

nikkimk (approved):
lost gold treasure map

hellogreg:
Looks like the button and list may be getting misaligned upon expansion in Chrome and Safari (Firefox is okay): 
dropdown alignment issue

hellogreg:
With VoiceOver/Safari on Mac, the button is not currently announcing an expanded or collapsed state upon focus.

bennypowers:
@hellogreg PTAL, thanks

hellogreg:
Looks like the `aria-expanded` attribute isn't being applied to the button--and thus the state is not getting announced by screen readers: 
accessibility tree showing no aria expanded state on button

hellogreg:
WIth the Windows Firefox/NVDA browser and screen reader combo, the screen reader stays in browse/read mode when focus is on the button. The user _can_ manually activate focus mode to use the button, but it should auto-detect.

This isn't an issue with any other browser/reader combo. (Typically, this issue crops up for more than one of them at a time, so this is a little unusual.)

bennypowers:
i think we'll need to remove the native `<button>` from pf-button's shadow root, and make the pf-button host a role: button with element internals.

nikkimk:
> i think we'll need to remove the native `<button>` from pf-button's shadow root, and make the pf-button host a role: button with element internals. this could affect chip, select, and others.

Yep

bennypowers:
blocked by https://github.com/patternfly/patternfly-elements/issues/2706

elements/pf-dropdown/test/pf-dropdown.spec.ts
bennypowers, line 1: don't merge before #2707

bennypowers:
@hellogreg PTAL

bennypowers:
> ### Visibility issue
Please see #2710 

> ### Mac Safari/VoiceOver, Win Firefox/NVDA, Win Chrome/JAWS, Android Chome/Talkback
> 
> Dropdown works well with these combos. I do have one question. Do you want/need to have a default designated focus destination when a menu item is activated and the dropdown closes? After pressing `esc`, focus returns to the button, as I'd expect. But when activating a menu item, visible focus disappears. (It seems like it's then being set back to the beginning of the page, maybe.)

@nikkimk wdyt?

bennypowers:
I'm going to merge this now, and if @nikkimk you want to adjust the default focus behaviour that @hellogreg  mentioned above, we can follow up later or do a patch release

PR #2709, opened 2024-03-19
fix(chip): fixed `<pf-chip-group>` keyboard accessibility

nikkimk:
## What I did

1. Ensured that roving tabindex is updated when `<pf-chip-group>` buttons are added.

## Testing Instructions

1. In DP, go to [Chip group with removable categories](https://deploy-preview-2709--patternfly-elements.netlify.app/components/chip/#chip-group-with-removable-categories)
2. Use <kbd>Tab</kbd> to set focus on first chip in a chip group.
3. Use right arrow key <kbd>&rarr;</kbd> to navigate to **"1 more"** overflow button.
4. Use right arrow key <kbd>&rarr;</kbd> to navigate to close button.


bennypowers (approved):
Loosey goosey, tighty moosey

PR #2710, opened 2024-03-20
fix: unminify css when bundling

bennypowers:
## What I did

1. unminify the shadow css when bundling

## Testing Instructions

1. on the [DP chip group demo page](https://deploy-preview-2710--patternfly-elements.netlify.app/components/chip/demo/chip-group-with-visible-category-name/), chip close buttons should be vertically centered, as normal. If this PR is broken, they will appear below the vertical center line. 

Screenshot 2024-03-20 at 8 00 43


zeroedin (approved):
lunar gravity, tugging moons

PR #2711, opened 2024-03-20
fix(button): use :active pseudo

bennypowers:
## What I did

1. remove the `active` shadow class and use the host pseudo :active instead

## Testing Instructions

1. add these tokens to the dp demo and test clicking and dragging the button.
2. when clicking the button, active state should persist as long as the mouse is down
3. when mousing down, then dragging out of the button: active state should persist as long as the mouse is down, even if it leaves the browser viewport.

### DIFF
--- a/elements/pf-button/pf-button.css
+++ b/elements/pf-button/pf-button.css
@@ -77,21 +77,6 @@ pf-icon,
   position: relative;
 }
 
-.active,
-:host(:active) {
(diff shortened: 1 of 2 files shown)


zeroedin (changes requested):
`<pf-button variant="link">` do not appear to have `:active` or `:hover` styles that are shown in PFv4: https://v4-archive.patternfly.org/v4/components/button

zeroedin (approved):
lonely goats travel mountains

Issue #2422, opened 2023-03-13
[docs]: shortcut icon


bennypowers:
on the heels of #2396, we need to add an svg icon (in lieu of favicon) for patternflyelements.org

gautamkrishnar:
I am interested in working on this

PR #2722, opened 2024-03-24
docs: updated shortcut icon

gautamkrishnar:
## What I did

1. Updated shortcut icon to new logo

## Testing Instructions

1. Check favicon when preview is generated

## Notes to Reviewers
n/a

fixes: #2422


tools/pfe-tools/config.ts
```diff
@@ -36,7 +36,7 @@ export interface PfeConfig {
 
 const SITE_DEFAULTS: Required<SiteOptions> = {
   description: 'PatternFly Elements: A set of community-created web components based on PatternFly design.',
-  favicon: '/docs/images/logo/pfe-icon-blue.svg',
+  favicon: '/docs/images/favicon.ico',
```
bennypowers, line 39: instead of the .ico, can we use an svg here?

gautamkrishnar: @bennypowers do you mean using `pfe-logo-icon-white.svg` ? It is a pure white variant of the logo.

bennypowers: WDYT about this one? https://github.com/patternfly/patternfly-elements/blob/main/brand/logo/svg/pf-logo-small.svg

PR #2729, opened 2024-03-27
fix(core)!: deprecate cascades decorator

bennypowers:
Lit context provides us with an easier to manage and more robust alternative to the imperative cascades system, let's migrate to that and remove this in PFE 4

## What I did

1. mark `@cascades` and `CascadeController` as deprecated

## Notes to reviewers:
RHDS uses `@cascades` in one file:

```
❯ rg cascades
elements/rh-tabs/rh-tabs.ts
8:import { cascades, deprecation } from '@patternfly/pfe-core/decorators.js';
140:  @cascades('rh-tab')
150:  @cascades('rh-tab', 'rh-tab-panel')
156:  @cascades('rh-tab', 'rh-tab-panel')
```

which we will remove when we update to TabsController, following PFE 3 release


bennypowers:
@Arathy-s does cp elements use this?

zeroedin (approved):
Large granite tower mountains (aka the cascades mountains 🤣 )

PR #2740, opened 2024-04-02
fix(select): placeholder button text

bennypowers:
## What I did

1. ensure placeholder value is used for button label unless there is a selected item

## Testing Instructions

1. add placeholder attr
2. see that it is used as button text
3. select an item
4. see that the item's value is used as button text

## Notes to Reviewers

1. is the use of the placeholder as a disabled item correct?

### DIFF
--- a/elements/pf-select/pf-select.ts
+++ b/elements/pf-select/pf-select.ts
@@ -17,6 +17,7 @@ import { FloatingDOMController, type Placement } from '@patternfly/pfe-core/cont
 import { PfOption } from './pf-option.js';
 
 import styles from './pf-select.css';
+import { SlotController } from '@patternfly/pfe-core/controllers/slot-controller.js';
 
 export interface PfSelectUserOptions {
   id: string;
@@ -50,7 +51,10 @@ export class PfSelectChangeEvent extends Event {
 export class PfSelect extends LitElement {
   static readonly styles = [styles];
 
-  static override readonly shadowRootOptions: ShadowRootInit = { ...LitElement.shadowRootOptions, delegatesFocus: true };
+  static override readonly shadowRootOptions = {
+    ...LitElement.shadowRootOptions,
+    delegatesFocus: true,
+  };
 
   static readonly formAssociated = true;
 
@@ -59,6 +63,8 @@ export class PfSelect extends LitElement {
   #float = new FloatingDOMController(this, {
     content: () => this.shadowRoot?.getElementById('listbox-container') ?? null,
   });
+
+  #slots = new SlotController(this, null, 'placeholder');
 
   #listbox?: ListboxController<PfOption>; /* | ListboxActiveDescendantController */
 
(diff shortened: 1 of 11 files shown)


nikkimk (changes requested):
@bennypowers Neither placeholders nor selected items should be used as labels. And placeholder's IDREF could be used in as an `aria-describedby`, but a label serves a different purpose than a placeholder.

bennypowers:
@nikkimk @adamjohnson is this better?
Now, the button will have text content from either the selected item or the placeholder, be described by the placeholder, and be labelled by the input (in the case of typeahead) or directly with the computed label text (based on FACE labels)

elements/pf-select/pf-select.ts
```diff
@@ -163,11 +163,29 @@ export class PfSelect extends LitElement {
     return this.variant === 'checkbox' && !this.checkboxSelectionBadgeHidden;
   }
 
-  /**
-   * array of text content from listbox's array of selected options
-   */
-  get #valueTextArray() {
-    return this.#listbox?.selectedOptions.map(option => option.optionText || '') ?? [];
+  get #buttonLabel() {
```
nikkimk, line 166: Button label **_should not_** be the placeholder or a selected value. Use `accessibleLabel` or have a way to associate the select with a `<label>` element.

nikkimk (changes requested):

hellogreg:
I think shouldn't use values from the options to label the entire select list--especially considering those values can change.

For example, if we have a select menu for choosing a system's status, its label should always be something like "Current system status," and not "Running". Here's a screencap from our docs:

Select list with the value Running selected

For an example of what we could do, the [WAI Select-Only Combobox Example](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/) uses `aria-labelledby` to point to the text just preceding the select list to label it. We could do something like this, even if that text is placed offscreen.

adamjohnson (changes requested):

bennypowers:
@nikkimk 's notes from chat

1. what text must appear in the combobutton before the user opens the select menu?
   If an option is pre-sepected, that option, if not but there is placeholder, the placeholder, otherwise, the first option. Note: What appears in the box is NOT a label.
2. what text must be announced when sr focuses on the combobox?
   First a label and then the text in the box as the selected value. Note that text is different from a label. The text should not be labelled as the combobox's label.
3. how many focusable items must appear in the select list - is there a fourth unselectable item?
   If there is a placeholder, it is an unselectable item that is only read when the select has not value. The placeholder text can also serve as aria-describedby for the combobox.
4. when the users selects an item, what effect must that have on the answers to questions 1 and 2?
   What appears in the comboxbox is the selected item. The screenreader should read a label (which is distinct from an option/placeholder) followed by the current value as the text.
5. pf-select is a FACE with no role. I can <label> the host from outside but not the shadow dom controls, or use the InternalsController#computedLabelText helper to copy the host FACE's label text onto a shadow control. If i did that, which control should it be copied to?
   If there is a label, the aria-label should be copied to the element with role="combobox". If there is not or if there is an accessibleLabel property set, that's what would be the combobox's aria-label

nikkimk (approved):
Lonely ghost terminates matchmaker

hellogreg:
Only thing I noticed with screen readers is that the placeholder text is _always_ read when the button receives focus, even when another value in the list is currently selected. (If the placeholder is currently selected, it is read twice.)

PFE demo screencap

For example, when the above button receives focus after I've selected "Orange" (which is not the placeholder value), NVDA announces, `Color, combobox, Orange, collapsed, opens list, Select a value` on focus:

NVDA screencap

So you get the label, role, selected value, expanded state, button action, and placeholder value.

WAI combobox screencap

By comparison, the above [WAI Combobox](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/) announces, `Favorite Fruit, combobox, Blueberry, collapsed, opens list`. In other words, everything but the placeholder value from our demo.

Other screen readers act similarly. VoiceOver on the Mac actually does something especially odd. It trims the last letter of the placeholder (e.g., "Select a valu" without the "e" at the end).

bennypowers:
I think we can add an aria hidden on the placeholder text whenever there's a selected value

bennypowers:
@hellogreg 73014286 addresses this, you can test in at https://patternflyelements.org/elements/select/demo/

hellogreg:
@Thanks, @bennypowers . The current version has this odd flickering ([video link](https://drive.google.com/file/d/1Ru7JU-6RX0D-aIHLLPqEQlhuLUPFfYUz/view?usp=sharing)) when triggering the button via either mouse or keyboard. You also then can't collapse the list, except by choosing an option.

PR #2742, opened 2024-04-08
chore: eslint flat config

bennypowers:
## What I did

1. export eslint flat config from eslint packages
2. update repo to use flat config

## Testing Instructions

1. `npm run lint`

## Notes to Reviewers
I recommend using the "hide whitespace" option in the github diff view


elements/pf-chip/test/pf-chip-group.spec.ts
```diff
@@ -192,7 +192,7 @@ describe('<pf-chip-group>', async function() {
     });
 
     describe('keyboard navigating with arrow keys to third chip and pressing enter', function() {
-      beforeEach(() => element.focus());
```
bennypowers, line 195: @nikkimk this was the only change I was able to ID that might have practical significance. tests fail on this branch with `.focus` on the chip group, and in my investigation, i found that this assertion failed because the `activeElement` remained `<body>` the whole time.

this fix, using playwright to simulate a `Tab` key press, rather than calling `.focus()` fixes the test.

I suspect, but can't yet confirm, that this is because of a bump in playwright version in this PR.

In any event, please confirm that this change doesn't represent breakage for users.

PR #2762, opened 2024-05-19
feat!: ssr

bennypowers:
## What I did

1. remove DOM side effects from core main module
2. add ssr test helpers
3. add ssr test demo files
4. write some ssr tests

## Testing Instructions

```
npx playwright test -g ssr --update-snapshots --max-failures=1000
npx playwright show-report test-results
```

## Notes to Reviewers

1. ~~this should maybe be refactored to put everything in PfeDemoPage~~
2. maybe we should containerize this to get uniform results
3. The test results show exactly where components fail to SSR
4. This PR **does not need** to get to 100% ssr success for every element, it's enough for now to just understand clearly where we fail


bennypowers:
I think we've gone as far with this as we can. The remaining failures are blocked by https://github.com/lit/lit/issues/3301

bennypowers:
- [ ] remove `unresolved` from wds template

bennypowers (commented):
Missing changesets for new feature in `@patternfly/pfe-tools/ssr`

PR #2781, opened 2024-06-20
feat(tools): knobs

bennypowers:
## What I did

1. add elements to tools
2. add pft-element-knobs
3. add knobs demos to dev server

## Testing Instructions

1.

## Notes to Reviewers

1.

### DIFF
--- a/elements/pf-select/pf-select.ts
+++ b/elements/pf-select/pf-select.ts
@@ -306,7 +306,7 @@ export class PfSelect extends LitElement {
       this.#expandedChanged();
     }
     if (changed.has('value')) {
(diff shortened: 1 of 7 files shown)


zeroedin:
@bennypowers this is great!  Thinking about picking up `pf-drawer` to support this.  My thinking is maybe the knobs could be out of the way on initial load then toggleable via a drawer.  Thoughts?

bennypowers:
There can be quite a lot of content, more than a side drawer


See text-input for example

Maybe a bottom panel

zeroedin:
> Maybe a bottom panel

Yeah I also noted the resizable drawer too which might be nice on certain viewport sizes.

PR #2782, opened 2024-06-21
fix(create-element): inline demo subresources

bennypowers:
## What I did

1. update the generator's demo templates, removing the `demo.css` and `${elementName}.js` subresources, and inlining them to the HTML instead.


Arathy-s (approved):
The changes look good to me.

PR #2808, opened 2024-07-09
feat(core): internals form disabled update

bennypowers:
## What I did

1. make `InternalsController` call `requestUpdate` in `formDisabledCallback`
2. remove manual calls from elements

## Testing Instructions

1. unit tests, form control demos

## Notes to Reviewers

1. keep an eye out for update loops / extraneous update calls


bennypowers:
I think there's a browser or spec issue here that needs fleshing out

PR #2809, opened 2024-07-11
feat(select): typeahead

bennypowers:
## What I did

1. add `ActivedescendantController`
2. add `typeahead` variant to select

## Testing Instructions

1. this pr requires thorough screen reader testing, but not until it's out of draft


hellogreg:
On the demo in the latest versions of all Mac browsers and Windows Chrome/Edge (but not Win Firefox, Linux Firefox, iOS Safari, or Android Chrome), the browser window flickers upon moving focus to the combobox and toggling it expanded. Looks like vertical and horizontal scrollbars are rapidly appearing/disappearing. This occurs with or without screen readers enabled.

Combobox window flickering

hellogreg:
Here's the good, the bad, and the weird about keyboard nav with Chrome (even without a screen reader on). Quite a bit is different from the WAI example--which I would try to mimic as much as possible.

Chrome keyboard nav notes:
- Focus on input/button is visible.
- List opens upon space.
- List does not open upon enter.
- List opens upon typing _any_ character, whter it begins a list item or not.
   - Visible focus outline remains on input/button.
   - Highlighted state appears on "Alabama," no matter what is typed.
Focus remains on the input/button when the list opens. Pressing tab moves focus into the list (to the "Select a state" option).
- However, focus does not need to move into the list in order to highlight options. 
  - Arrows highlight options when focus is still on the combobox.
  - In fact, arrows _don't_ work when focus is in the list.
- Selecting any option and pressing enter closes the list but does not put selected text in the input. Same with the escape key.
  - If nothing has been typed, "Select a state" is still in the input, but is no longer a dimmed placeholder. It's the actual value.
  - If a value has been typed, it remains the input value when the list closes.
- If the closed list is re-opened, the most-recently highlighted value is still highlighted.

hellogreg:
Safari is similar, if a bit more forgiving in one way. Highlighting an option in the list with Safari does select it--so it is in the input upon closing the list.

hellogreg:
Mac Firefox is like Chrome, best I can tell.

I'd tackle the keyboard issues before even going into screen reader testing.

hellogreg:
Oh, in case this is helpful, here's where the WAI example is the same or differs:

- Focus on input/button is visible. (same as our DP)
- List does not open upon space. (different)
- List does not open upon enter. (same)
- Full list opens upon pressing down arrow. (same)
- When typing characters, the list only opens when you type a letter than an option begins with. (different)
- List is then navigated via arrows. (same)
- Focus indicator moves with arrows. (different)
- When a list item receives focus, it is now selected and displays in the input upon pressing enter or escape. (different)
- Pressing tab moves to the next focusable item after the combobox--not into the list. (different)

kelsS:
@bennypowers @hellogreg based on what I have read and with all these different issues happening with different screen readers and browsers. I am not sure that cloning the nodes is the best path. The a11y expert who wrote that book I mentioned strongly suggested having all focusable elements either in the light dom or in the shadow dom. I was wondering if that approach is possible to try? It could solve a lot of these headaches with screen readers and browsers. also can send ya'll the relevant pages from the book to read yourself.

bennypowers:
That's what the node cloning accomplishes

For supporting browsers, no cloning is necessary, since the aria relationship is set up in javascript by way of the (unfortunately capitalized) `ariaActiveDescendantElement` IDL property.

 This is the case in safari, where inspecting the accessibility properties of the combobox node (in shadow) shows that it has "Shared focus" with the AT-focused item node. I'm on mobile now so please forgive me for not attaching a screenshot.

There is a separate ,unrelated bug in safari whereby activedescendant doesn't work with voiceover *at all* regardless, that bug would still affect us even if we never used shadom dom for anything. We've documented our workaround and it's tradeoffs in this pr, and we've discussed a possible more sophisticated workaround involving Alert role, which comes with its own issues.

Now, for unsupporting browsers, when we clone in the nodes, we end up with a situation where all the nodes are in the same root - the shadow root. Its true that we retain the original nodes in place (in light dom), but they are slotted into a container which is `hidden`, so they don't affect A.T. The reason we retain those nodes is to make it easier to integrate pf-select in projects of various kinds.  

There are gotchas here for sure, but i would like to think we've done a job here which is good enough to cover the majority of cases and users while looking forward to better browser support as early as next year.

kelsS:
> g browsers, when we clone in the nodes, we end up with a situation where all the nodes are in the same root - the shadow root. Its true that we retain the original nodes in place (in light dom), but they are slotted into a container which is `hidden`, so they don't affect A.T. The reason we retain those nodes is to make it easier to integrate pf-select in projects of various kinds.
> 
> There are gotchas here for sure, but i would like to think we've done a job here which is goo

@bennypowers thank you for that wonderful explanation! I think I am on the same page now. I am looking forward to getting to use this controller. Thank you so much and to everyone else who has been working hard on this. These are such complicated issues with strange gotchas. I am glad to know that they are being accounted for as best as possible within the constraints we are working in. 🎉

kelsS:
@bennypowers @hellogreg what is this component supposed to do? When I look at the [demo](https://deploy-preview-2809--patternfly-elements.netlify.app/components/select/demo/typeahead/) and try to interact with it nothing happens :/ 

Here are my OS and Browser specs:
MacBook Pro Intel (2019)
OS - Sonoma 14.5

ChromeVersion 127.0.6533.100 (Official Build) (x86_64)

adamjohnson:
I am seeing the same browser console errors Kelsey is seeing in the DP. 

I am also having trouble getting the project to `build`/`start`. When I run `npm run build`, it says `a11y-snapshot.js` is not listed within the file list of the project. Here's [the output](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-both/).

@kelsS Yes, this is supposed to function like an [Editable Combobox with Autocomplete](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-both/).

bennypowers:
The technical term for the errors is "oopsies"

My bad, I'll hopefully have these sorted by am EST tmrw

bennypowers:
OK, it turns out that when I pushed my changes at EoD yesterday, I had forgotten to stage the file `pf-select.ts`, which lead to all the errors you saw @kelsS and @adamjohnson. I'm sorry about that, they should be resolved now: https://deploy-preview-2809--patternfly-elements.netlify.app/components/select/demo/typeahead/

I noticed another issue: I put the demo elements in a container with padding and gave them a max width. Prior to this, the select element filled the entire width of the container, extending all the way to the right edge of the screen. under that condition, when i clicked on the typeahead combobox, the following happened:
- the listbox displayed
- the browser scrollbars (mac ff) rapidly flickered
- the browser viewport rapidly flickered
I suspect this is more related to `@floating-ui/dom`, so i'd like to ignore that bug and tackle it in a separate PR

hellogreg:
If it's helpful, I started a [Google Sheets doc](https://docs.google.com/spreadsheets/d/1-8u-_ta9pXyTSd-eH-pcK4OFF8M7ihL5wvz-SIkV5eQ/edit?usp=sharing) comparing what happens upon various keypresses in this DP, the WAI demo, `<select>`, and `<datalist>`.

I think the biggest issue of concern with this specific demo is the auto-closing of the list immediately upon pressing space to open it. (I'm guessing that's because no option in the list begins with a space.)

hellogreg:
Steven, Adam, and I created a [running `<pf-select>` doc](https://docs.google.com/document/d/1N-UddDJyDh_1B1k78S15q68evOsgX7T4gH2daOTp5IQ/edit?usp=sharing), with details on issues as we spot and document them. Seemed easiest to put something there, so we could team up and edit.

adamjohnson:
Guidance on label/placeholder combos:

**Without placeholder with label:**
Have the first <pf-option> show in the input as the placeholder (aka: no changes from current code).


**With placeholder no label:**
Add an aria-label to the input with the value from the `placeholder` attribute.

Maybe add an `accessible-label` API?


**With placeholder with label:**
Ideal case, render both as the user intended.


**Without placeholder no label:**
I would `logger` out a message saying something like "Your typeahead requires an HTML label element to be accessible. Please add a label."

If this is not practical, then... Perhaps some generic aria-label added to the input, "Select a value"?

bennypowers:
> **Without placeholder with label:** Have the first show in the input as the placeholder (aka: no changes from current code).

My fav kind of spec ;)

> **With placeholder no label:** Add an aria-label to the input with the value from the `placeholder` attribute.
> 
> Maybe add an `accessible-label` API?

We have an `accessible-label` attribute already, and most test cases use it. see pf-select.spec.ts
 
> **With placeholder with label:** Ideal case, render both as the user intended.

When a `<label>` element is associated, the user is responsible for styles, so that's fine. In that case, the toggle button, combobox input, and listbox elements all receive either `ariaLabelledByElements` or `ariaLabel` (computed from the label elements), depending on browser support, from the host.
When `accessible-label` is applied, same as above, but with the text value of the attribute  

> **Without placeholder no label:** I would `logger` out a message saying something like "Your typeahead requires an HTML label element to be accessible. Please add a label."
> 
> If this is not practical, then... Perhaps some generic aria-label added to the input, "Select a value"?

We'd like to move away from these kind of DOM-validating logs in favour of something like @paceaux's [Selector Hound](https://github.com/paceaux/selector-finder/), which can statically analyze full pages' html.

So, summing it up, it sounds like the only change needed in code is to label-by the placeholder as a last-resort fallback when there are no other labels?

zeroedin:
oops, that commit message should have been `test(core): ...` 😞

PR #2845, opened 2024-09-15
feat(core): hasSlotted SSR hints

bennypowers:
## What I did

1. adds ssr-only `ssr-hint-has-slotted="slotA,default,slotB"` attribute to elements which use SlotController

## Testing Instructions

1. view and verify results for pf-card.e2e.ts

## Notes to Reviewers

1. Not entirely sure how best to document this, but i'm mostly certain I don't want to add this to individual element docs, but rather on a centralized ssr docs page


elements/pf-card/test/pf-card.e2e.ts
```diff
@@ -22,4 +22,21 @@ test.describe(tagName, () => {
...
+      demoContent: /* html */ `
+        <pf-card ssr-hint-has-slotted="header,default,footer">
+          <h2 slot="header">Header</h2>
+          <span>Body</span>
+          <span slot="footer">Footer</span>
+        </pf-card>
+      `,
+    });
+    await fixture.updateCompleteFor('pf-card');
+    await expect(fixture.page.locator('pf-card #title')).toHaveAttribute('hidden');
```
zeroedin, line 40:
```Suggestion
    await expect(fixture.page.locator('pf-card #title')).toHaveAttribute('hidden');
    await expect(fixture.page.locator('pf-card #header')).not.toHaveAttribute('hidden');
```
Add an inverse test as well checking if the hidden attribute is not added for slots that are hinted.

zeroedin (approved):
just the one comment on the test, otherwise LGTM

PR #2855, opened 2024-09-23
fix(core): overflow controller resize loop

zeroedin:
Downstream issue:

>  ❌ <rh-tabs> > on small screen > reversed right to left language overflow actions > previousTab should be disabled
      Error: ResizeObserver loop completed with undelivered notifications. (http://localhost:8003/?wtr-session-id=E_xerQsq1EQq2yny1h-Ga:0)

> If you want to prevent these errors, the solution depends on what your intended effect is. If you actually intend to have an infinite loop, you just need to defer the resizing code in your ResizeObserver callback to after the browser repaints. You can put it into a [requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame) callback.

[Documentation Reference](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#observation_errors)

## What I did

1. Await browser repaint before triggering overflow side effects

## Testing Instructions

1.

## Notes to Reviewers

1.

### DIFF
--- a/.changeset/fluffy-bananas-remember.md
+++ b/.changeset/fluffy-bananas-remember.md
@@ -1,5 +1,5 @@
 ---
-"@patternfly/pfe-core": major
+"@patternfly/pfe-core": patch
 ---
 
 `OverflowController`: corrected sideeffects of resize event to occur after browser paint
--- a/.changeset/fluffy-bananas-remember.md
+++ b/.changeset/fluffy-bananas-remember.md
@@ -1,0 +1,5 @@
+---
+"@patternfly/pfe-core": major
+---
+
+`OverflowController`: corrected sideeffects of resize event to occur after browser paint
--- a/.changeset/fluffy-bananas-remember.md
+++ b/.changeset/fluffy-bananas-remember.md
@@ -2,4 +2,4 @@
 "@patternfly/pfe-core": patch
 ---
 
-`OverflowController`: corrected sideeffects of resize event to occur after browser paint
+`OverflowController`: prevent browser from locking up in some scenarios
--- a/core/pfe-core/controllers/overflow-controller.ts
+++ b/core/pfe-core/controllers/overflow-controller.ts
@@ -44,11 +44,8 @@ export class OverflowController implements ReactiveController {
     }
   });
 
-  #ro = new ResizeObserver((entries: ResizeObserverEntry[]) => {
-    window.requestAnimationFrame((): void => {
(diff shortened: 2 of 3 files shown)


core/pfe-core/controllers/overflow-controller.ts
```diff
@@ -44,8 +44,13 @@ export class OverflowController implements ReactiveController {
...
 
-  #ro = new ResizeObserver(() => {
-    this.#setOverflowState();
+  #ro = new ResizeObserver((entries: ResizeObserverEntry[]) => {
+    window.requestAnimationFrame((): void => {
+      if (!Array.isArray(entries) || !entries.length) {
+        return;
+      }
+      this.#setOverflowState();
+    });
```
bennypowers, line 53:
```suggestion
  #ro = new ResizeObserver(() => {
    requestAnimationFrame(() => {
      this.#setOverflowState();
    });
```

in my testing, we didn't need the checks. might as well save a few bytes

.changeset/fluffy-bananas-remember.md
```diff
@@ -0,0 +1,5 @@
+---
+"@patternfly/pfe-core": major
```
bennypowers, line 2:
```suggestion
"@patternfly/pfe-core": patch
```

.changeset/fluffy-bananas-remember.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+`OverflowController`: corrected sideeffects of resize event to occur after browser paint
```
bennypowers, line 5:
```suggestion
`OverflowController`: prevent browser from locking up in some scenarios
```

bennypowers (changes requested):

.changeset/fluffy-bananas-remember.md
zeroedin: Oops didn't mean to select major there, good catch 🙃

zeroedin:
@bennypowers adding `await aTimeout(50);` allows the test to complete.  That said I'm not convinced that just adding a timeout to the test is the correct way to handle red/green here.  Another way to do it is await 3x nextFrame.

```js
beforeEach(nextFrame);
beforeEach(nextFrame);
beforeEach(nextFrame);
beforeEach(updateComplete);
```

I can almost understand the need for nextFrame better here as we are now adding a tick given the overflow controller implementing the requestNextFrame.  Thoughts?

bennypowers (approved):

PR #2863, opened 2024-10-09
fix(core): tabs aria ssr compat

bennypowers:
## What I did

1. avoid browser globals/apis in tabs aria controller


bennypowers:
Reviewed with Steven

PR #2880, opened 2024-11-20
fix(core): slotController isEmpty respects text nodes

bennypowers:
## What I did

1. make slotcontroller's hasContent/isEmpty consider text nodes

## Testing Instructions

1. patch contents into RHDS, and check that cards with text nodes in default slot but no wrapping element still display their contents

## Notes to Reviewers

1. see https://stackblitz.com/edit/stackblitz-starters-u2qbnw?file=index.html


adamjohnson (approved):
Letters get thoroughly matched 👍

PR #2882, opened 2024-11-27
2566: Create pf-radio component

Arathy-s:
## What I did
Developed a custom `pf-radio` component utilizing the standard `<input type="radio" />` element for enhanced functionality and styling.
Screenshot 2024-11-27 at 1 19 45 PM

## Testing Instructions

**1. Mouse Interaction**

- [ ] **Single Click**: Clicking a radio button should select it.
- [ ]  **Subsequent Click**: Clicking an already selected radio button should not deselect it.
- [ ] **Group Selection**: When clicking a different radio button within the same group, the previously selected radio button should be deselected, and the new radio button should be selected.

**2. Keyboard Navigation**

- [ ] **Tab Key**(on page navigation): Pressing the Tab key should focus on the first radio button in the group. If a radio button is already selected, that selected button should be focused instead of the first one.
- [ ] **Shift + Tab**(on page navigation): Pressing Shift + Tab should focus on the last radio button in the group. If a radio button is already selected, that selected button should be focused instead of the last one.
- [ ] **Radio Selection**: A radio button can be selected using the **Space**, or **arrow keys**.
- [ ] **Arrow Keys**:
- **Arrow Down or Right**: Pressing the Down Arrow or Right Arrow should move the focus in a clockwise direction within the radio group.
- **Arrow Up or Left**: Pressing the Up Arrow or Left Arrow should move the focus in a counterclockwise direction within the radio group.

## Notes to Reviewers

Please note that the CSS variables, code comments, test cases, and updated documentation will be added asap.

### DIFF
--- a/elements/package.json
+++ b/elements/package.json
@@ -36,6 +36,7 @@
     "./pf-jump-links/pf-jump-links-list.js": "./pf-jump-links/pf-jump-links-list.js",
     "./pf-jump-links/pf-jump-links.js": "./pf-jump-links/pf-jump-links.js",
     "./pf-label/pf-label.js": "./pf-label/pf-label.js",
+    "./pf-radio/pf-radio.js": "./pf-radio/pf-radio.js",
     "./pf-select/pf-select.js": "./pf-select/pf-select.js",
     "./pf-select/pf-listbox.js": "./pf-select/pf-listbox.js",
     "./pf-select/pf-option-group.js": "./pf-select/pf-option-group.js",
--- a/elements/pf-radio/README.md
+++ b/elements/pf-radio/README.md
@@ -1,0 +1,11 @@
+# Radio
+Add a description of the component here.
+
+## Usage
+Describe how best to use this web component along with best practices.
+
+```html
+<pf-radio>
+
+</pf-radio>
+```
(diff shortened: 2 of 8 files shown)


hellogreg (commented):
So far, this is passing the tests you listed. However, I noticed that all the radio buttons are reading as "1 of 1" in screenreaders, and not "1 of 5."

radio buttons listing as 1 of 1

adamjohnson (changes requested):
## Tabbing Behavior

Looking at the "Multiple Groups" demo in Safari 18.1.1, there's currently some [very strange tabbing behavior](https://github.com/user-attachments/assets/3923fb97-ab06-4745-ba0f-c3fe6ea0d162).

This can be fixed by giving unique values to the `name` attributes for each radio group. For example:

```
<p><code>&lt;pf-radio&gt;</code> group with same name inside different parent</p>
<div class="radio-group-container">
  <div class="radio-group">
    <p>Score: </p>
    <pf-radio value="a" name="score2" label="A"></pf-radio>
    <pf-radio value="b" name="score2" label="B"></pf-radio>
    <pf-radio value="c" name="score2" label="C"></pf-radio>
  </div>
...
</div>
```

Note the unique `score2` value for `name`. The `name` attributes in the other radio groups should be made unique as well.

## Fieldset and legend

Instead of using `<div class="radio-group">` and `<p>` tags for labels, use `<fieldset>` and `<legend>` for optimum accessibility. [WebAIM fieldset / legend reference](https://webaim.org/techniques/forms/controls#radio).

## Screen Reader Testing

When this component is close to being "done", it's probably a good idea to have @hellogreg test it out with various screen readers to ensure it's accessible everywhere.

Issue #2813, opened 2024-07-18
pf-table: Accessibility issue


Arathy-s:
## Description of the issue

Multiple accessibility issues are being logged on performing an accessibility audit on the pf-table component.

### Impacted component(s)

- pf-table

### Steps to reproduce

Run Lighthouse check for the page https://patternflyelements.org/components/table/ 

### Expected behavior

There should not be any accessibility issues logged.

### Screenshots

![Screenshot 2024-07-18 at 4 57 41 PM](https://github.com/user-attachments/assets/b241f5e5-6278-40fd-94d4-ad180a3ef2ed)
![Screenshot 2024-07-18 at 4 58 21 PM](https://github.com/user-attachments/assets/28afcb37-fff3-49c7-83ba-33ec1c6468cb)

## Related issues

- #100

PR #2885, opened 2024-12-06
2813: Fix table accessibility issue

atharvkulkarni2002:
## What I did

- Fixed issues mentioned in the [issue]( https://github.com/patternfly/patternfly-elements/issues/2813).
- Improved accessibility  score of pf-table (Table, Expandable Rows Compound, Sortable) from **91** to **100** 

**Before**:
Screenshot from 2024-12-06 12-11-29
Screenshot from 2024-12-06 12-11-53
Screenshot from 2024-12-06 12-12-17

**After**:
Screenshot from 2024-12-06 12-08-09
Screenshot from 2024-12-06 12-09-16
Screenshot from 2024-12-06 12-09-40

---------------------------------------------------------------------------------

- Improved accessibility score of pf-table (Expandable Rows) improved from **81** to **90**.

**Before**:
Screenshot from 2024-12-06 12-12-55

**After**:
Screenshot from 2024-12-06 12-10-17

### DIFF
--- a/elements/pf-table/pf-tr.ts
+++ b/elements/pf-table/pf-tr.ts
@@ -110,6 +110,8 @@ export class PfTr extends LitElement {
           <pf-button id="toggle-button"
                      aria-expanded=${String(this.expanded) as 'true' | 'false'}
                      plain
+                     role="button"
+                     aria-label="Expand Button"
                      @click=${this.#onClick}>
             <pf-icon id="toggle-icon"
                      icon="angle-right"
--- a/elements/pf-table/pf-th.ts
+++ b/elements/pf-table/pf-th.ts
@@ -51,7 +51,7 @@ export class PfTh extends LitElement {
     const closestThead = this.closest('pf-thead');
     const closestTable = this.closest('pf-table');
     const isChildOfThead = !!closestThead && !!closestTable?.contains(closestThead);
-    const role = isChildOfThead ? 'colheader' : 'rowheader';
+    const role = isChildOfThead ? 'columnheader' : 'rowheader';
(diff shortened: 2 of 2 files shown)


elements/pf-table/pf-tr.ts
```diff
@@ -110,6 +110,8 @@ export class PfTr extends LitElement {
           <pf-button id="toggle-button"
                      aria-expanded=${String(this.expanded) as 'true' | 'false'}
                      plain
+                     role="button"
```
zeroedin, line 113: We shouldn't need to set role here as this gets applied at element internals level if the variant is to any other variant (or lack of variant) other then value of `link` which sets to `none`.

```Suggestion
```

elements/pf-table/pf-tr.ts
```diff
@@ -110,6 +110,8 @@ export class PfTr extends LitElement {
...
+                     aria-label="Expand Button"
```
zeroedin, line 114: `pf-button` has an API for adding the accessible label to the shadow dom button.  We'd us `label` here instead of using the `aria-label` on the host.
```Suggestion
label="Expand Button"
```

bennypowers:
can the button's `aria-controls` reference be to the row on https://github.com/patternfly/patternfly-elements/blob/1e2f9a198402602a2f9bccd387be80957444705f/elements/pf-table/pf-tr.ts#L125?

elements/pf-table/pf-tr.ts
atharvkulkarni2002: Thank you for your suggestion.
 I have removed role="button" and pushed the changes.

PR #2891, opened 2025-01-13
fix(core): ssr events

bennypowers:
The ## What I did

1. update lit packages related to ssr and context (i.e. events in ssr)
2. ~write a server-side `SlotController` which adds hint attributes to all elements that use `SlotController`.~ deferred to #2893 

## Testing Instructions

1. view SSR test output, there should be no errors related to `addEventListener`

## Notes to Reviewers

1. see https://github.com/lit/lit.dev/pull/1390 and https://github.com/lit/lit/pull/4755
2. there are some element refactors here which account for the fact that we're opting in to calling `connectedCallback` and `hostConnected` - i suspect some of these should be refactored in order to make that more clear from the code
3. netlify builds are failing, probably because there are too many files. that will have to be dealt with somehow


core/pfe-core/ssr-shims.ts
```diff
@@ -52,7 +28,27 @@ globalThis.getComputedStyle ??= function() {
       return '';
     },
   };
-}
+};
 
-;
+// @ts-expect-error: opt in to event support in ssr
+globalThis.litSsrCallConnectedCallback = true;
```
bennypowers, line 34: should we really be making this decision for our users? or should e.g. rhds' ssr regime choose whether or not to enable this?

core/pfe-core/controllers/slot-controller.ts
```diff
@@ -55,60 +55,108 @@ export class SlotController implements ReactiveController {
...
 
   #deprecations: Record<string, string> = {};
 
-  constructor(public host: ReactiveElement, ...config: ([SlotsConfig] | (string | null)[])) {
-    this.#logger = new Logger(this.host);
+  #mo = new MutationObserver(this.#initSlotMap.bind(this));
 
-    if (isObjectConfigSpread(config)) {
+  constructor(public host: ReactiveElement, ...args: SlotControllerArgs) {
+    const singleton = SlotController.singletons.get(host);
```
bennypowers, line 73: This is probably a breaking change to core.

bennypowers (commented):
This PR should be pared down to concentrate on lit, context-api, and table changes

slot controller changes should be deferred to another PR

core/pfe-core/decorators/slots.ts
```diff
@@ -0,0 +1,27 @@
+import type { ReactiveElement } from 'lit';
+
+import { SlotController, type SlotControllerArgs } from '../controllers/slot-controller.js';
+
+/**
+ * Enable ssr hints for element
+ * @param args a spread of slot names, or a config object.
+ * @see SlotController constructor args
+ */
+export function slots<T extends typeof ReactiveElement>(...args: SlotControllerArgs) {
```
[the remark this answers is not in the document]
bennypowers: Managed to remove this

core/pfe-core/controllers/slot-controller.ts
bennypowers: Managed to avoid this

elements/pf-table/pf-table.ts
zeroedin, line 1: Maybe move these changes to a new PR since they are less concerned with the SSR changes and are a fix for `pf-table` explicitly, although the context change allows it to happen in SSR.

bennypowers: Because of the connected callback call, SSR on table would throw at the closest call

So this change corrects table behaviour, true, but it also addresses a "regression"*  brought on by connectedCallback

core/pfe-core/ssr-shims.ts
bennypowers: Making this decision for users means that yes they would have to refactor their elements, any time ssr-shims.js is loaded. i.e. if they use the `renderGlobal` function from `@patternfly/pfe-tools/ssr/global.js`. adding event listeners would not be a problem, because `EventTarget` is supported in node, but DOM access would throw. see element changes in this PR.

`isServer` is a boolean exported from a module in the lit package with import and node conditions, so it exports `true` when the node condition obtains and `false` otherwise. I don't see how wrapping that in a method would help.

bennypowers: This is a major change for core, and a patch for elements.

adamjohnson (approved):
Listeners grow through maintenance 👏

zeroedin (approved):

zeroedin:
@bennypowers I'll let you squash and merge here so you can look at the change I added due to updating the branch and changes made to scrollspycontroller in the meantime.

bennypowers:
Thanks! I reverted 701f33fc7 here so we can deal with it in #2900

PR #2893, opened 2025-01-30
feat(core): slot controller ssr hint attributes

bennypowers:
Follow up to #2891 

## What I did

1. Add a server-side only slotcontroller
2. add `ssr-hint-has-slotted` and `ssr-hint-has-default-slot` attributes

## Testing Instructions

1. tbd

## Notes to Reviewers

1.


heyMP:
It's such a bummer that Lit-SSR doesn't support reading children. I get it but it's unfortunate. I think it would be a great feature if you could disable "streaming mode" on Lit-SSR and you'd have access to the whole DOM content.

I think this fix is really creative and I think it's a good solution. I do worry though about making it mandatory for implementors who are trying to pre-render. I can see more instances of these `ssr-hint-*` attributes in the future to get around the Lit-SSR issue.

I also wonder if we could reduce the need for slot-controller by utilizing the CSS :has() selector.  For instance, if you're relying on hiding and showing certain pieces of the shadowroot based on what slots are in the lightdom, I think that could be a better use case for CSS :has() over slot-controller.  It would remove the requirement for `ssr-hint-*` attributes for that specific component.  Just thinking out loud. 😀

zhawkins:
Very creative for sure. That said, I share Potters concerns about tackling it with attributes.

Can you share an example or two of

> elements which use SlotController to render the shadowroot differently depending on which slots have content

Curious to know what that looks like and if a CSS solution like Potter touched on might be viable. 🤔

bennypowers:
Light DOM CSS solves a different problem than this PR. 

Light DOM CSS is *only* for styling deeply nested children, since the alternative (slapping style attributes on children via the DOM API) is truly unpalatable.

This PR is _primarily_ about ponyfilling `:has-slotted` for shadow stylesheets via shadow classes, but not exclusively.


```ts
#slots = new SlotController(this, 'belly');

render() {
  return html`
    <div id="stomach" class="${classMap({ empty: this.#slots.isEmpty('belly') })}">
      <belly-button>
      <slot name="belly"></slot>
    </div>
    <!-- head, shoulders, knees, toes -->
  `;
}
```

This case shows how using slotcontroller we could hide or otherwise style the container for the belly slot when there's no slotted belly content. How would you handle this case with light dom css? You couldn't without exposing a shadow part. 

So why not just expose a shadow part then? That would be applying a solution in the element's *design* to solve a technical problem in the element's *implementation*. that's a severe inversion of priorities.


The solution we're looking for is something like:

```css
#stomach:has(:has-slotted(*)) {
  /* ... */
}
```

but that's not quite ready yet, so we need a stopgap.

docs/components/demos.11tydata.cjs
```diff
@@ -0,0 +1,64 @@
...
+    'lit@2.0.2/directives/unsafe-svg.js',
+    'lit@2.0.2/directives/until.js',
+    'lit@2.0.2/experimental-hydrate-support.js',
+    'lit@2.0.2/experimental-hydrate.js',
+    'lit@2.0.2/html.js',
+    'lit@2.0.2/index.js',
+    'lit@2.0.2/polyfill-support.js',
+    'lit@2.0.2/static-html.js',
+  ],
+};
```
zeroedin, line 64: These are really old versions;  and it doesn't appear to output unless modifying the for loop in demos.html

```
  {%- for path in preloads -%}  <!-- random comment -->
    <link rel="modulepreload" href="https://ga.jspm.io/npm:{{ path }}">
  {%- endfor -%} 
```

added a comment to the for loop to get it to show up on :8080 otherwise it doesn't render 🙃 

likely we could just remove?

bennypowers: yeeted

zeroedin (approved):
LGTM

adamjohnson (approved):
Labels give template mapping

PR #2896, opened 2025-03-06
fix(tools): pass ignore option to generator plugin

bennypowers:
## What I did

1. pass important options to import map generator plugin


markcaron (approved):
Let's Generate Those Maps!

PR #2904, opened 2025-04-10
chore: update wireit dep for cache v2 api

zeroedin:
## What I did

1 . Updated the wireit dependency to 0.14.12

Ref: https://github.com/google/wireit/issues/1297

## Testing Instructions

1. Does the DP fail with cache issues?

## Notes to Reviewers

1.


bennypowers (approved):
lessening girth through middleware

PR #2912, opened 2025-06-17
fix(tools)!: remove cem config

bennypowers:
## What I did

1. removed custom elements manifest analyzer config
2. replace cem/a with `@pwrs/cem`

## Testing Instructions

1. see that the deploy preview has the same element api docs as patternflyelements.org


ArathyKumar (approved):
The changes look good to me.

PR #2944, opened 2025-08-20
fix(tools): correct lightdom css routing, improve trailing slash beha…

zeroedin:
## What I did

1.  Corrected lightdom(-shim).css routing in dev server
2. Improved trailing slash behavior in dev server

## Testing Instructions

1.  Open dev server, add `-lightdom.css` to a component and link in a demo.  Ensure lightdom css is properly linked.

## Notes to Reviewers

1.

### DIFF
--- a/tools/pfe-tools/dev-server/plugins/dev-server-router.ts
+++ b/tools/pfe-tools/dev-server/plugins/dev-server-router.ts
@@ -39,6 +39,62 @@ const cacheBustingMiddleware: PfeMiddleware = () => async function(ctx, next) {
 };
 
 /**
+ * Redirects both /elements and /elements/ to the root
+ * Handles both cases: /elements → / and /elements/ → /
+ */
+const elementsToRootRedirectMiddleware: PfeMiddleware = () => ctx => ctx.redirect('/');
+
+/**
+ * Ensures trailing slash for component URLs
+ * FROM: `/elements/footer`
+ *   TO: `/elements/footer/`
+ * @param config normalized PFE dev server config
+ */
+const ensureTrailingSlashMiddleware: PfeMiddleware = () => (ctx, next) => {
+  // Only add trailing slash if path doesn't already end with one
+  if (!ctx.path.endsWith('/')) {
+    return ctx.redirect(`${ctx.path}/`);
+  }
+  // If it already ends with slash, continue to next middleware
+  return next();
+};
+
+/**
+ * Handles lightdom CSS files that are missing the tag name in the path
+ * FROM: `elements/pf-jazz-hands-lightdom.css` or `elements/pf-jazz-hands-lightdom-shim.css`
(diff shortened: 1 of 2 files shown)


zeroedin:
`npm run docs` fails on `main` branch as well right now for the same error. 

>[11ty] Problem writing Eleventy templates:
>[11ty] 1. Having trouble rendering njk template ./docs/quick-start.md (via TemplateContentRenderError)
>[11ty] 2. (./docs/quick-start.md)
>[11ty]   EleventyNunjucksError: Error with Nunjucks paired shortcode `generateImportMap` (via Template render error)
>[11ty] 3. Cannot read properties of undefined (reading 'replace') (via TypeError)

tools/pfe-tools/dev-server/plugins/dev-server-router.ts
```diff
@@ -38,6 +38,62 @@ const cacheBustingMiddleware: PfeMiddleware = () => async function(ctx, next) {
...
+ */
+const elementsToRootRedirectMiddleware: PfeMiddleware = () => ctx => ctx.redirect('/');
+
+/**
+ * Ensures trailing slash for component URLs
+ * FROM: `/elements/footer`
+ *   TO: `/elements/footer/`
+ * @param config normalized PFE dev server config
+ */
+const ensureTrailingSlashMiddleware: PfeMiddleware = () => (ctx, next) => {
```
adamjohnson, line 53: This is a minor nitpick, but if:

http://localhost:8000/components/jump-links

will add the trailing `/`, would it also be worthwhile to add the trailing slash to the other demos as well? eg:

```
http://localhost:8000/components/jump-links/demo/centered-list
...becomes...
http://localhost:8000/components/jump-links/demo/centered-list/
```

Feel free to resolve if this is out of scope.

adamjohnson (approved):
Dunno if you wanna also try tackling adding a trailing slash to non-`index.html` demos. Otherwise, LGTM 👏 .

adamjohnson:
This probably needs a changeset.

bennypowers:
Sorry for the delay in review. Please consider that an indication of how ill-advised my original approach is, moreso than as a comment on your changes

- has this been tested on Windows?
- what is this coming to solve? Im still not clear on what the exact problem is that this addresses

tools/pfe-tools/dev-server/plugins/dev-server-router.ts
zeroedin: the trailing slashes themselves were kinda a nitpick of my own, so why not solve one more case.

zeroedin:
> That being said, I have questions:
> 
> * has this been tested on Windows?
> * what is this coming to solve? Im still not clear on what the exact problem is that this addresses

- Has not been tested on Windows. I'm not sure why it would be any different, but I will test.
- Solves the fact that in the dev server, the `-lightdom.css` files literally do not load downstream in RHDS given the current state of things.

bennypowers (changes requested):

PR #2952, opened 2025-11-06
feat(select): add dark mode support for <pf-select>

dali327519294:
## What I did

Fixed bug https://github.com/redhat-ux/red-hat-design-system/issues/2681

1.Added dark mode support for `<pf-select>` to ensure proper contrast and accessibility in both light and dark themes.

## Testing Instructions

1. Open the `<pf-select>` demo page.
2. Switch between light and dark color schemes (using the browser or system setting).
3. Verify that the dropdown, text, and border colors adjust correctly.

## Notes to Reviewers

1. No breaking changes.
2. Added a changeset for minor version bump.
3. Verified behavior manually in both modes.

### DIFF
--- a/elements/pf-select/pf-select.css
+++ b/elements/pf-select/pf-select.css
@@ -535,6 +535,7 @@ pf-badge {
   flex: 1 0 100%;
 }
 
+
 @media (prefers-color-scheme: dark) {
   :host {
     color: var(--pf-global--Color--light-100, #f0f0f0);


bennypowers:
@zeroedin has an approach in https://github.com/redhat-ux/red-hat-design-system/issues/2681. if we update PFE there to v4.2.0, we can hack around this in RHDS without having to introduce light/dark to PFE here.

https://github.com/RedHat-UX/red-hat-design-system/pull/2684#discussion_r2457148845

bennypowers:
we have `<rh-select>` now

PR #2960, opened 2025-12-26
fix: restore missing bundled entrypoint

bennypowers:
Avoid using this though, because of custom element double-registration

## What I did

1. re-add pfe.min.js to the tarball


bennypowers:
let's remove this in pfe 5

PR #3093, opened 2026-04-15
fix(core): prevent SlotController.getSlotted() returning empty array prior to initialization

### LINKED ISSUE #2946: [bug]: SlotController issue in `getSlotted()` not returning elements as expected
`getSlotted()` should return the assignedElements() of a slot.  It's usage downstream in `rh-subnav` was returning an empty array `[]`.  Needs more investigation.  

For now, removing the SlotController from `rh-subnav` is lighter weight for the simple usage in `rh-subnav` rather than adding the whole dependency of SlotController.

bennypowers:
## Summary

Fixed `getSlotted()` returning empty arrays when called before async
slot record initialization completes. Falls back to querying slot
elements directly when records aren't yet populated.

Closes #2946

## Test plan

- [ ] Existing slot-controller tests pass
- [ ] Verify `getSlotted()` returns correct elements in downstream consumers

### DIFF
--- a/.changeset/fix-slot-controller.md
+++ b/.changeset/fix-slot-controller.md
@@ -1,0 +1,6 @@
+---
+"@patternfly/pfe-core": patch
+---
+
+`SlotController`: fixed `getSlotted()` returning empty arrays in
+certain timing scenarios.
--- a/core/pfe-core/controllers/slot-controller.ts
+++ b/core/pfe-core/controllers/slot-controller.ts
@@ -215,11 +215,24 @@ export class SlotController implements SlotControllerPublicAPI {
    */
   public getSlotted<T extends Element = Element>(...slotNames: string[] | [null]): T[] {
     if (!slotNames.length || slotNames.length === 1 && slotNames.at(0) === null) {
-      return (this.#slotRecords.get(SlotController.default)?.elements ?? []) as T[];
+      return (this.#getAssignedElements(SlotController.default)) as T[];
     } else {
       return slotNames.flatMap(slotName =>
-        this.#slotRecords.get(slotName ?? SlotController.default)?.elements ?? []) as T[];
+        this.#getAssignedElements(slotName ?? SlotController.default)) as T[];
     }
+  }
+
+  /**
+   * Returns the assigned elements for a given slot name, falling back to
+   * querying the slot element directly if the slot record hasn't been
+   * initialized yet.
+   */
+  #getAssignedElements(slotId: string | symbol): Element[] {
+    const record = this.#slotRecords.get(slotId);
+    if (record) {
+      return record.elements;
+    }
+    return this.#getSlotElement(slotId)?.assignedElements?.() ?? [];
   }
 
   /**


zeroedin:
**Alternative approach:** 

synchronous `hostUpdated()` instead of async fallback

The current fix adds a fallback in `getSlotted()` to query slot elements directly when `#slotRecords` hasn't been populated yet. This addresses the symptom but leaves the root cause:  `hostConnected` and `#initSlotMap` are `async`, creating a timing gap where no data source exists.

The gap also affects `hasSlotted()` and `isEmpty()`, which aren't addressed here.

**Why `hostConnected` is `async`**

Both `hostConnected `and `#initSlotMap` `await host.updateComplete` just to ensure the shadow DOM exists so they can query for <slot> elements. But Lit already provides a synchronous hook that  guarantees this: `hostUpdated()`, which runs after every render when the DOM has been committed.

**Suggested change**

Make `hostConnected` synchronous (just sets up the MutationObserver), move record initialization to `hostUpdated()`, and have the MutationObserver trigger a re-render instead of calling  `#initSlotMap` directly:

```js
  hostConnected(): void {
    this.#mo.observe(this.host, { childList: true });
    this.#slotRecords.clear();
  }

  hostUpdated(): void {
    if (this.#slotRecords.size === 0) {
      for (let slotName of this.#slotNames.concat(Object.values(this.#deprecations))) {
        slotName ||= SlotController.default;
        const slot = this.#getSlotElement(slotName);
        if (slot) {
          this.#slotRecords.set(slotName, new SlotRecord(slot, slotName, this.host));
        }
      }
      if (this.#slotRecords.size > 0) {
        this.host.requestUpdate();
      }
    }
  }
```

The MutationObserver callback becomes:

```js
#mo = new MutationObserver(() => this.host.requestUpdate());
```

And `#initSlotMap` can be removed entirely.

This also requires updating the SlotControllerPublicAPI declaration to match:

```js
hostConnected?(): void;
```

**Why this works**

- First render: Records empty, `isEmpty()` returns true (matches SSR server behavior). `hostUpdated()` populates records and calls requestUpdate().
- Second render: Records exist, `isEmpty()`/`getSlotted()` read live DOM via SlotRecord getters, correct results. `hostUpdated()` sees records exist, no `requestUpdate()`, no infinite loop.
- MutationObserver: Child changes trigger `requestUpdate()` → re-render → SlotRecord getters reflect new DOM state. No record rebuild needed since they query live DOM.
- Reconnection: `hostConnected` clears records → next render → `hostUpdated()` repopulates.

**What this fixes that the current PR doesn't**

- `hasSlotted()` and `isEmpty()` have the same timing gap, they also read from #slotRecords and return incorrect results before async init completes. The `hostUpdated()` approach fixes all three  methods.
- Eliminates the `async` architecture that caused the issue rather than working around it.

**Trade-off**

The two-render pattern on initial connection is inherent `<slot>` elements are created by `render()`, so the first render can't know about them. This was already the case with the `async` approach.

We can spin this off to another issue if we feel the need

zeroedin (approved):
This looks good to me as is, the alternative comment can be broken out into its own issue and is not a blocker to this PR.  Tested artifact down stream in RHDS.

PR #3094, opened 2026-04-15
fix(switch): add static accessible label

### LINKED ISSUE #2754: [bug] `<pf-switch />` - Switch lacks a static accessible label
## Description of the issue

`<pf-switch />` lacks an accessible static label. The current label uses the state of the switch as the label.

### Impacted component(s)

- `<pf-switch />`

#### Basic Demo

1. Go to the [`<pf-switch />` "Basic" demo](https://patternflyelements.com/components/switch/).
2. When clicking the switch, the message changes from "Message when on" to "Message when off". These messages incorrectly act as the label for the switch.

#### Without Label Demo

1. Go to the ["Without Label"](https://patternflyelements.com/components/switch/#without-label) demo.
2. This demo lacks an accessible label and accessible states for its controls.

#### Other demos

Similarly, the "Checked with label" and "Disabled" follow these trends and need fixed.

### Expected behavior

All `<pf-switch />` elements should include an accessible label for the control. 

1. They should also have an `aria-describedby` attribute that targets the controls toggled state options.
2. _If_ a `<label>` isn't used, the switch needs an `accessible-label` property/attribute.

#### Proposed fixes:

```html
<label for="with-label">Switch B</label>
<pf-switch id="with-label" aria-describedby="b" checked>
  <div id="b">
    <span data-state="on">Message when on</span>
    <span data-state="off" hidden>Message when off</span>
  </div>
</pf-switch>
```

```html
<pf-switch aria-describedby="a" accessible-label="Switch A" checked>
  <div id="a">
    <span data-state="on">Message when on</span>
    <span data-state="off" hidden>Message when off</span>
  </div>
</pf-switch>
```

Related to https://github.com/RedHat-UX/red-hat-design-system/pull/1513#issuecomment-2062119853.

bennypowers:
## Summary

- Rename `label` property to `accessibleLabel` (`accessible-label` attribute)
- Wire to `ariaLabel` on ElementInternals for a proper static accessible name
- Updated "without label" demo to use `accessible-label`
- Added tests for accessible label behavior

Closes #2754

## Test plan

- [ ] Verify switch has accessible name when `accessible-label` is set
- [ ] Verify accessible name is static regardless of checked state
- [ ] Run tests: `npx wtr elements/pf-switch/test/pf-switch.spec.ts`

### DIFF
--- a/elements/pf-switch/pf-switch.ts
+++ b/elements/pf-switch/pf-switch.ts
@@ -24,8 +24,8 @@ export class PfSwitch extends LitElement {
 
   #internals = InternalsController.of(this, { role: 'switch' });
 
-  /** Accessible label text for the switch */
-  @property({ reflect: true }) label?: string;
+  /** Accessible label for the switch when there is no associated `<label>` element. */
+  @property({ reflect: true, attribute: 'accessible-label' }) accessibleLabel?: string;
 
   /** Flag to show if the switch has a check icon. */
   @property({ reflect: true, type: Boolean, attribute: 'show-check-icon' }) showCheckIcon = false;
@@ -57,6 +57,7 @@ export class PfSwitch extends LitElement {
   override willUpdate(): void {
     this.#internals.ariaChecked = String(!!this.checked);
     this.#internals.ariaDisabled = String(!!this.disabled);
+    this.#internals.ariaLabel = this.accessibleLabel || null;
   }
 
   override render(): TemplateResult<1> {
--- a/elements/pf-switch/pf-switch.ts
+++ b/elements/pf-switch/pf-switch.ts
@@ -24,7 +24,14 @@ export class PfSwitch extends LitElement {
 
   #internals = InternalsController.of(this, { role: 'switch' });
 
-  /** Accessible label for the switch when there is no associated `<label>` element. */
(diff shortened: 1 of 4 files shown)


elements/pf-switch/pf-switch.ts
```diff
@@ -24,8 +24,8 @@ export class PfSwitch extends LitElement {
 
   #internals = InternalsController.of(this, { role: 'switch' });
 
-  /** Accessible label text for the switch */
-  @property({ reflect: true }) label?: string;
+  /** Accessible label for the switch when there is no associated `<label>` element. */
+  @property({ reflect: true, attribute: 'accessible-label' }) accessibleLabel?: string;
```
zeroedin, line 35: wouldn't this change techncially be a major as we are not deprecating the `label` prop and creating a new `accessible-label`?

bennypowers: you're right

bennypowers: instead of removing `label` altogether, let's add `accessible-label`, fallback to `label`, and mark `label` as deprecated

zeroedin (approved):

PR #3095, opened 2026-04-15
feat(label): add link variant

bennypowers:
## Summary

Added `href` property to `pf-label`. When set, the label text renders
inside an `<a>` element with transparent background, no border, and
underlined text inheriting the label's color.

Closes #2499

## Test plan

- [ ] Verify `<pf-label href="https://example.com">Link</pf-label>` renders an anchor
- [ ] Verify labels without `href` are unchanged

PR #3104, opened 2026-04-16
fix(label): only thicken border on link hover, not close button

bennypowers:
## Summary

Use `#container:has(#link:hover)` instead of `:host(:hover)` so the
border only thickens when hovering the link text, not the close button.
Matches PFv4 behavior per @adamjohnson's review on #3095.

## Test plan

- [x] Hover link text: border thickens to 2px
- [x] Hover close button: border stays 1px
- [x] Compare with https://v4-archive.patternfly.org/v4/components/label/react/router-link/

### DIFF
--- a/elements/pf-label/pf-label.css
+++ b/elements/pf-label/pf-label.css
@@ -305,15 +305,8 @@ pf-button {
   cursor: pointer;
 }
 
-:host(:hover) #link ~ *,
-:host(:focus-within) #link ~ *,
-#link:hover,
-#link:focus {
-  cursor: pointer;
-}
-
-:host(:hover) #container:has(#link)::before,
-:host(:focus-within) #container:has(#link)::before {
+#container:has(#link:hover)::before,
+#container:has(#link:focus)::before {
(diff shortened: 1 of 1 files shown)


adamjohnson (commented):
Here's what the focus state looks like in PF Label:

pf-label with close button focused, focus state intact

And here's what it looks like in PFE Label:

pfe label with focus border being cut off by overflow hidden

The focus outline is getting cut off. This can be fixed by removing `overflow: hidden;` on the `#container`. That said, there may be other side effects to removing that.

More or less, this LGTM and, if you deem it fit, removing the `overflow: hidden;` is a nice add.

elements/pf-label/pf-label.css
```diff
@@ -20,7 +20,8 @@ pf-icon, ::slotted(pf-icon) {
 }
 
 #container {
-  overflow: hidden;
+  overflow: clip;
+  overflow-clip-margin: 3px;
```
adamjohnson:
```suggestion
  overflow-clip-margin: 6px;
```

6px makes the entire focus outline visible.

adamjohnson (approved):
Links Get Tactile Markers

PR #3105, opened 2026-04-17
docs: adding LGTM tradition to README and storing full list of LGTMs

markcaron:
## What I did

1. Added LGTM tradition section to README
2. Stored full list of LGTMs as a `.md` file

## Testing Instructions

1. Check out both. What do you think?!

## Notes to Reviewers

Used an AI skill to generate/update the list and update the README.


docs/lgtms.md
bennypowers, line 1: please move this to the docs/ dir

bennypowers (approved):

PR #3115, opened 2026-04-29
fix: strip v5 prefix from docs titles, replace jspm with mappa

bennypowers:
## Summary

- Strip "V5 " prefix from element names in docs sidebar, page titles, and demo headings (e.g. "V5 Accordion" becomes "Accordion")
- Replace `@jspm/generator` with `@pwrs/mappa` for import map generation, fixing the `Unable to resolve npm:lit@3.3.2` build error
- Add v5 entry to the docs version dropdown

## Changes

**Title stripping** (`DocsPage.ts`, `Manifest.ts`):
- `prettyTag()` now accepts an optional `prefix` parameter to strip the full tag prefix instead of only the first word segment
- `DocsPage` passes the configured `tagPrefix` so `pf-v5-accordion` becomes "Accordion" not "V5 Accordion"

**Import map generation** (`importMap.cjs`, `import-map-generator.ts`, `create-import-map.cjs`):
- Replaced `@jspm/generator` with `@pwrs/mappa` in all three import map files
- CDN import map uses `generate()` with explicit dependency list and `exclude` for workspace packages
- Dev server plugin uses `resolve()` for local node_modules resolution
- Removed jspm-specific provider/resolver configuration from dev server config

**Version dropdown** (`versions.json`):
- Added v5 as current version, moved v4 to non-current

## Test plan

- [x] `npm run clean && npm run serve` builds and starts successfully (281 pages)
- [x] Sidebar shows "Accordion", "Alert", etc. without "V5 " prefix
- [x] Page titles show "Accordion - PatternFly Elements"
- [x] Component pages load at `/components/accordion/`, `/components/button/`, etc.
- [x] Version dropdown shows v5 (current/disabled), v4, v3, v2, v1


zeroedin (approved):
Legacy Generator Transition: Mappa

PR #3116, opened 2026-04-30
fix(core): iframe support in floating dom controller

bennypowers:
## What I did

1. fix floating-dom-controller in iframes

## Testing Instructions

1. add `<iframe src="/components/tooltip/demo/" height="800"></iframe>` to any docs page
2. `npm run serve`
3. note that tooltip works


zeroedin (approved):
LGTM, tested downstream as well

PR #3132, opened 2026-05-08
chore(tools): update tooling config for v6 prefix

### LINKED ISSUE #3130: chore: update tooling and config for v6 prefix
## Phase 0.2: Rename tooling and config for v6 prefix

Part of the v6 migration (#3110). Tooling and config need updating so that `pf-v6-*` elements are recognized, and `pf-v5-*` exports are removed since the v6 release will not ship v5 elements.

### Tasks

- [ ] Update `eslint.config.js` ignore globs — currently hardcoded to `pf-v5-icon` demo path (line 16), needs to include `pf-v6-icon` (or generalize)
- [ ] Update `elements/package.json` exports — replace all `pf-v5-*` export entries with their `pf-v6-*` equivalents. The v6 release will not contain pf-v5 elements, so these should be updated, not duplicated
- [ ] Audit `tsconfig` paths for any v5-specific entries
- [ ] Update `web-test-runner.config.js` and `playwright.config.ts` globs if needed
- [ ] Verify `commitlint.config.js` handles both prefixes (currently strips `pf-v\d+-` dynamically — looks fine already)

### Notes

- `commitlint.config.js` already normalizes scopes dynamically via regex (`pf-(?:v\d+-)?`), so no changes needed there
- `tsconfig`, `web-test-runner`, and `playwright` configs don't appear to have hardcoded v5 globs currently, but should be verified as elements are ported
- This can be done incrementally as v6 elements land, or as a batch once enough elements exist

### Related

- Migration plan: `.claude/V6-MIGRATION-PLAN.md` (Phase 0.2)
- Parent issue: #3110

zeroedin:
### Phase 0.2

- Generalize eslint ignore glob for icon demo path to handle both `pf-v5-icon` and `pf-v6-icon`
- Update eslint `no-missing-package-exports` rule to match only `pf-v6-*` element directories
- Remove all `pf-v5-*` export entries from `elements/package.json` — v6 release will not ship v5 elements
- Bump elements package version to `6.0.0-next.0`
- Audited tsconfig, web-test-runner, playwright, and commitlint configs — no changes needed

Closes #3130


zeroedin:
@bennypowers, we have a `.config/cem.yaml` in the elements dir, and another in the root. It is my assumption that the one in the `elements` dir needs to be removed. Not sure why that wouldn't just sit at the root? It was causing problems with the agents running. Thoughts?

.config/cem.yaml
```diff
@@ -1,5 +1,6 @@
 sourceControlRootUrl: https://github.com/patternfly/patternfly-elements/tree/main/
 generate:
+  output: ./elements/custom-elements.json
```
bennypowers: not sure about this... we also want to generate a manifest for core

elements/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@patternfly/elements",
   "license": "MIT",
-  "version": "5.0.0",
+  "version": "6.0.0-next.0",
```
bennypowers: this is something that changesets should manage, not us manually

eslint.config.js
```diff
@@ -25,7 +25,7 @@ export default tseslint.config(
     plugins: { '@patternfly/elements': pfePlugin },
     rules: {
       '@patternfly/elements/no-missing-package-exports': [States.ERROR, {
-        matches: ['elements/*/pf-*.js'],
+        matches: ['elements/pf-v6-*/pf-*.js'],
```
bennypowers: the icon change is good, but revert this line

bennypowers (changes requested):

.config/cem.yaml
bennypowers: I'm going to work on multiple outputs from monorepo roots in cem

PR #3134, opened 2026-05-08
chore: staging/pfv6

zeroedin:
## What I did

1.

## Testing Instructions

1.

## Notes to Reviewers

1.


zeroedin (commented):
pf-tokens.css CSS rules are loaded in the `cem serve` ui chroming. Visual testing needs to be done with `?rendering=chromeless`, as this query param removes the `pf-tokens.css` present in the full Chrome UI. Because these are ultimately the same tokens, even though loaded for CEM UI and not for pf6 development, they overlap and leak in.  However, components should be more fault-tolerant and include the styles they need to look the way we want independently of that file's existence through failovers and setting font rules necessary per component.

PR #3135, opened 2026-05-08
chore: add claim-element skill

zeroedin:
## What I did

1. Added  `/claim-element` skill.

This skill will check the current status of the project and see if anyone else is currently workin gon a component. It uses the Phase migration in `V6-MIGRATION-PLAN` and the issue epic and sub-issues for each element.  

Run `/claim-element,` and it will generate the recommended element for you to pick up, avoiding ones already claimed by others. It should warn you if you try to pick a Phase 2 element when its dependencies have not yet been merged from Phase 1, and the same with Phase 3.

## Testing Instructions

1.

## Notes to Reviewers

1.

### DIFF
--- a/.claude/skills/claim-element/SKILL.md
+++ b/.claude/skills/claim-element/SKILL.md
@@ -1,0 +1,184 @@
+---
+name: claim-element
+description: >
+  Claim a v6 element to work on. Checks issue assignees, phase ordering,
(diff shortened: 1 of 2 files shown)


bennypowers:
github issue assignments are sufficient, let's not overcomplicate things

PR #3137, opened 2026-05-08
feat(spinner)!: port to pf-v6-spinner

### LINKED ISSUE #3037: [feat]: Update `<pf-v6-spinner>` for PatternFly v6
## Summary

Update `<pf-v6-spinner>` to implement the PatternFly v6 [Spinner](https://www.patternfly.org/components/spinner) component.

Loading indicator with size variants, used inline or as a page-level loading state.

## Prior Art

| Source | Element | Status |
|--------|---------|--------|
| PFE v4 | `<pf-spinner>` | Exists |
| cem | -- | -- |
| cockpit | -- | -- |
| chickadee | `<chickadee-spinner>` | Exists |

## Requirements

- MUST provide end-user feature parity with `@patternfly/react-core` Spinner

zeroedin:
## Summary
- Removes `<pf-v5-spinner>`
- Ports `<pf-v5-spinner>` to `<pf-v6-spinner>` web component for PatternFly v6
- SVG-based animated spinner with `role="progressbar"` via ElementInternals
- Supports `size` (xs/sm/md/lg/xl), `inline`, `diameter`, and `accessible-label` attributes
- CSS animation with `prefers-reduced-motion` support
- Demos: basic, size-variations, custom-size, inline-size

Closes #3037
Depends on #3130

## Intentional divergences from React
- `xs` size supported (exists in PF core SCSS but not React's enum)
- `size` defaults to `undefined` (CSS fallback achieves xl visual default without sprouting attribute)
- `accessible-label` abstracts `ariaLabel`/`ariaValueText` per ADVICE.md
- Uses modern CSS `rotate` property instead of `transform: rotate()`
- `prefers-reduced-motion` handled via media query (React uses `.pf-m-no-motion` class)

## Test plan
- [ ] `npm run test` passes
- [ ] Demos render correctly at `localhost:8000` with `?rendering=chromeless`
- [ ] All size variants visually match patternfly.org
- [ ] Screen reader announces `progressbar` role with correct label
- [ ] Animation stops with `prefers-reduced-motion: reduce`

Assisted-By: Claude <noreply@anthropic.com>


.config/cem.yaml
```diff
@@ -1,5 +1,6 @@
 sourceControlRootUrl: https://github.com/patternfly/patternfly-elements/tree/main/
 generate:
+  output: ./elements/custom-elements.json
```
bennypowers: revert unrelated

elements/package.json
```diff
@@ -49,6 +49,7 @@
     "./pf-v5-progress/pf-v5-progress.js": "./pf-v5-progress/pf-v5-progress.js",
     "./pf-v5-search-input/pf-v5-search-input.js": "./pf-v5-search-input/pf-v5-search-input.js",
     "./pf-v5-spinner/pf-v5-spinner.js": "./pf-v5-spinner/pf-v5-spinner.js",
+    "./pf-v6-spinner/pf-v6-spinner.js": "./pf-v6-spinner/pf-v6-spinner.js",
```
bennypowers: we can probably get rid of all of these, mappa handles trailing-slash exports

elements/pf-v6-spinner/pf-v6-spinner.ts
```diff
@@ -0,0 +1,73 @@
...
+    role: 'progressbar',
+    ariaValueText: 'Loading...',
+    ariaLabel: 'Loading...',
+  });
+
+  override willUpdate(changed: Map<PropertyKey, unknown>): void {
+    if (changed.has('accessibleLabel')) {
+      const label = this.accessibleLabel ?? 'Loading...';
+      this.#internals.ariaLabel = label;
+      this.#internals.ariaValueText = label;
```
bennypowers: this should be a separate public property

cc @adamjohnson

bennypowers (changes requested):
The issue in question is an "update" issue, not a "create" issue:
- commit message must be `feat(spinner)!: port to pf-v6-spinner`
- delete elements/pf-v5-spinner

elements/pf-v6-spinner/pf-v6-spinner.ts
```diff
@@ -0,0 +1,73 @@
...
+  @property({ reflect: true }) size?: SpinnerSize;
+
+  /**
+   * When true, the spinner inherits its font size from the surrounding text,
+   * allowing it to display inline with content.
+   */
+  @property({ type: Boolean, reflect: true }) inline = false;
+
+  /** Custom diameter of spinner set as CSS variable */
+  @property({ reflect: true }) diameter?: string;
```
bennypowers: this should be an exclusively css thing, it also overlaps with size

bennypowers (changes requested):
should remove all existing pf-v5-spinner references

elements/pf-v6-spinner/pf-v6-spinner.ts
zeroedin: Ok, will remove this. It was actually prompted whether the agent should do that. I left it as is to stay closer to the React API. That said, would you go so far as to say we are setting a repeatable precedent that, if the prop sets a value in React (80px), or whatever it may be, we would only handle this as a CSS var? Something we could encode into the skill?

elements/pf-v6-spinner/README.md
```diff
@@ -0,0 +1,51 @@
...
+<pf-v6-spinner size="xl">Loading...</pf-v6-spinner>
+```
+
+### Custom size
+
+```html
+<pf-v6-spinner style="--pf-v6-c-spinner--diameter: 80px">Loading...</pf-v6-spinner>
+```
+
+## Discrepancies from React
```
zeroedin: @bennypowers thoughts?

elements/pf-v6-spinner/pf-v6-spinner.ts
bennypowers: Cases like this, where the js api *just sets a css property on this.style* with no other effect, and no independent semantic meaning, yes it should be CSS only

In other words no style attributes which don't carry independent semantic weight.

A mental test for this: if setting the attribute would bear meaning even without the css application, consider keeping. If not, css only.

In all cases, document css properties and class props to satisfy cem health checks

elements/pf-v6-spinner/README.md
zeroedin: updated in https://github.com/patternfly/patternfly-elements/pull/3137/changes/896df94e991556b64800c4d18251148a7e66004b

zeroedin:
Tests are failing here because  pf-v5-button imports pf-v5-spinner, and pf-v5-alert uses pf-v5-button internally. The chain is pf-v5-alert → pf-v5-button → pf-v5-spinner (which this PR deleted).  

Should we update pf-v5-button to point at pf-v6-spinner or just move on?

elements/pf-v6-spinner/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Custom size
+
+```html
+<pf-v6-spinner style="--pf-v6-c-spinner--diameter: 80px">Loading...</pf-v6-spinner>
+```
+
+## Divergences from React `Spinner`
+
+### Not implemented
```
bennypowers: this is a "changed". "not implemented" means "can't do it with the custom element"

elements/pf-v6-spinner/README.md
```diff
@@ -0,0 +1,50 @@
...
+| React prop | Notes |
+|------------|-------|
+| `aria-labelledBy` | Use `accessible-label` attribute instead, or set `aria-labelledby` via ElementInternals from a framework wrapper. |
+
+### Changed API
+
+| React prop | Web component | Difference |
+|------------|---------------|------------|
+| `diameter` | `--pf-v6-c-spinner--diameter` CSS custom property | React abstracts the CSS custom property behind a prop. In HTML, set it directly via `style`. |
+| `aria-valuetext` | `accessible-label` attribute | Sets both `aria-label` and `aria-valuetext` on the element internals. Defaults to `"Loading..."`. |
```
bennypowers: cc @adamjohnson will this double-announce? we'd probably prefer to have both `accessible-label` and `value-text` attrs

elements/pf-v6-spinner/README.md
```diff
@@ -0,0 +1,50 @@
...
+| `aria-label` | `accessible-label` attribute | Merged with `aria-valuetext` into a single `accessible-label` attribute. |
+
+### Added
+
+| Web component API | Notes |
+|-------------------|-------|
+| `size="xs"` | Extra-small size preset not available in React. |
+| `inline` attribute | Boolean attribute equivalent to React's `isInline`. |
+| `--pf-v6-c-spinner--Color` | CSS custom property to change the spinner stroke color. |
```
bennypowers: all three of these already exist on the react component

bennypowers (changes requested):

bennypowers (approved):

PR #3141, opened 2026-05-11
feat(tooltip)!: port to `pf-v6-tooltip`

### LINKED ISSUE #3048: [feat]: Update `<pf-v6-tooltip>` for PatternFly v6
## Summary

Update `<pf-v6-tooltip>` to implement the PatternFly v6 [Tooltip](https://www.patternfly.org/components/tooltip) component.

Text popup on hover/focus for supplementary information, with configurable position and delay.

## Prior Art

| Source | Element | Status |
|--------|---------|--------|
| PFE v4 | `<pf-tooltip>` | Exists |
| cem | -- | -- |
| cockpit | -- | -- |
| chickadee | -- | -- |
| portable-genie | `<genie-tooltip>` | Exists |

## Requirements

- MUST provide end-user feature parity with `@patternfly/react-core` Tooltip

bennypowers:
## Summary

- New `pf-v6-tooltip` element ported from v5 with v6 design tokens
- FloatingDOMController with arrow element positioning
- Cancellable `TooltipShowEvent`/`TooltipHideEvent` with `reason` enum for selective trigger prevention
- `trigger` attribute (IDREF) and property (Element ref) for external trigger elements
- `entry-delay`/`exit-delay` (default 300ms), `alignment` (start/end/left/right), `position`, `no-flip`, `flip-behavior`
- `light-dark()` CSS fallbacks for correct inverse color scheme
- `ariaDescribedByElements` wiring between trigger and tooltip content
- 9 demos: basic, placement, trigger-ref, trigger-element, slotted-content, dynamic-content, left-aligned, color-scheme, options playground
- 17 unit tests

## Test plan

- [x] Verify tooltip shows on hover and focus in all demos
- [x] Verify tooltip hides on mouseleave, focusout, and Escape
- [x] Verify light-dark color scheme inversion in color-scheme demo
- [x] Verify external trigger via IDREF (trigger-ref) and Element ref (trigger-element)
- [x] Verify `alignment` attribute changes text-align
- [x] Verify `entry-delay`/`exit-delay` in options playground
- [ ] Verify `no-flip` disables repositioning
- [x] Run `/review-a11y`, `/review-api`, `/review-demos`
- [x] Visual parity with patternfly.org tooltip

Closes #3048


bennypowers:
### Cross-root `aria-describedby` limitation

The tooltip sets `ariaDescribedByElements` on the trigger to reference the shadow-internal `#content` div. This is the correct API for cross-root ARIA relationships, but it silently fails in all browsers today: the accessibility tree shows no `describedby` relationship.

Per the current spec, ARIA element reflection only works when shadow roots are in an ancestor/descendant relationship, and even then only in the shadow-to-light direction. Our use case (light DOM trigger referencing shadow DOM content) goes the wrong way. The [WPT test](https://github.com/web-platform-tests/wpt/blob/master/custom-elements/element-internals-aria-element-reflection.html) explicitly asserts that cross-root refs produce an empty computed label.

This means tooltip triggers currently have **no programmatic `describedby` relationship** in any browser. The tooltip content is still announced when visible (via `role="tooltip"` and proximity in the accessibility tree), but there is no formal ARIA association.

The [Reference Target proposal](https://github.com/WICG/webcomponents/issues/1086) (Chrome origin trial) would solve the slotted trigger case. The external trigger case (sibling elements) would need the relaxation discussed in [WICG/aom#192](https://github.com/WICG/aom/issues/192). Both [whatwg/html#5401](https://github.com/whatwg/html/issues/5401) and [whatwg/html#10707](https://github.com/whatwg/html/issues/10707) track the broader spec work.

Keeping the `ariaDescribedByElements` call as progressive enhancement -- it will start working once browsers ship a solution. No workaround attempted; the alternatives (hidden light DOM spans, `aria-label` mirroring) add complexity for a temporary gap.

@adamjohnson and @hellogreg -- could you verify the actual screen reader experience here? Want to confirm the tooltip content is still announced via proximity/`role="tooltip"` even without the formal `describedby` association.

zeroedin (commented):
Rename PR `feat(tooltip)!: port to pf-v6-tooltip`

.changeset/keen-newts-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+"@patternfly/elements": minor
+---
+
+`pf-v6-tooltip`: New tooltip element with hover/focus activation, configurable placement, entry/exit delays, flip behavior, external trigger support, rich content via slot, and cancelable show/hide events.
```
zeroedin: This should be a major and follow the major changeset format

```md
---
"@patternfly/elements": major
---

✨ Added `<pf-v6-tooltip>` replacing `<pf-v5-tooltip>`. Tooltip now follows
PatternFly v6 design specs.

{{ example usage here }}

** Breaking Changes from v5 **

- {{Breaking changes here}}
```

elements/pf-v6-tooltip/README.md
```diff
@@ -0,0 +1,68 @@
...
+| `maxWidth`                 | `--pf-v6-c-tooltip--MaxWidth`         | CSS custom property instead of prop                                                     |
+| `children` / `triggerRef`  | default slot or `trigger` attribute   | Trigger is slotted child or ID string/Element reference                                 |
+| `onTooltipHidden`          | `hide` event                          | Cancelable DOM event with `reason` field. Also fires `show` event (no React equivalent) |
+| `animationDuration`        | CSS transitions                       | Override via stylesheet instead of prop                                                 |
+| `distance`                 | Hardcoded 15px                        | Not configurable; React default also 15                                                 |
+| `isVisible`                | `visible`                             | Boolean attribute; also controllable via `.show()` / `.hide()` methods                  |
+| `minWidth`                 | Not supported                         | No CSS custom property exposed                                                          |
```
zeroedin: `minWidth` should be moved to Not Implemented.

elements/pf-v6-tooltip/README.md
```diff
@@ -0,0 +1,68 @@
...
+| `zIndex`                   | Not supported                         | Hardcoded to 10000 in shadow DOM                                                        |
```
zeroedin: I'm hesitant about the hardcoded value here. I feel like we should offer a way to change it. I understand that there is no current CSS var, react is using an inline style prop drill. I can see this as an issue in some implementations if it cannot be fixed. Edge cases, sure, but hard-coding a high z-index is a bad practice.

would it be better to offer a net new css var here to allow an escape hatch and document it under "Added"?

elements/pf-v6-tooltip/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Added
+
+| Web component API             | Notes                                                               |
+| ----------------------------- | ------------------------------------------------------------------- |
+| `show` event                  | Cancelable event before tooltip shows, with trigger `reason`        |
+| `hide` event                  | Cancelable event before tooltip hides, with trigger `reason`        |
+| `.show()` / `.hide()` methods | Programmatic visibility control                                     |
+| `content` slot                | Rich HTML content, not available in React (which takes `ReactNode`) |
+| CSS custom properties         | 13 custom properties for fine-grained style control without JS      |
```
zeroedin: The CSS properties not not "Added" this line should be removed

bennypowers:
NB: i found _no_ instances of react `Tooltip` component's `minWidth` prop in the wild, scanning across 510 repos (160 importing Tooltip). I found _one_ instance of `Popover` react component's `minWidth` prop, so I think we're justified in removing it, since there's a (already popularly used) workaround - using `minWidth` css on the contents.

See generated report below. I don't endorse every recommendation here, this is just for reference.

<details><summary>Tooltip Audit Summary</summary>

**2296 instances** across **105/160 repos** | **48 spread usages**

## Props: PF React vs Actual Usage

| Prop | PF React | Used | Count | Recommendation |
|------|----------|------|-------|----------------|
| `content` | yes | yes | 2277 (99%) | **Keep** - core prop |
| `position` | yes | yes | 629 (27%) | **Keep** |
| `children` | yes | yes | (implicit) | **Keep** - trigger slot |
| `triggerRef` | yes | yes | 125 (5%) | **Keep** - ref-based trigger |
| `trigger` | yes | yes | 110 (5%) | **Keep (simplified)** - mostly `mouseenter`/`click`/`manual` |
| `id` | yes | yes | 63 (3%) | **Native HTML** |
| `isContentLeftAligned` | yes | yes | 59 (3%) | **Keep (simplified)** - CSS class toggle |
| `enableFlip` | yes | yes | 56 (2%) | **Keep** - Popper/Floating UI concern |
| `entryDelay` | yes | yes | 52 (2%) | **Keep** |
| `isVisible` | yes | yes | 40 (2%) | **Keep** - programmatic control |
| `aria` | yes | yes | 24 (1%) | **Keep (simplified)** - `describedby`/`labelledby`/`none` |
| `aria-live` | yes | yes | 20 (1%) | **Keep** |
| `maxWidth` | yes | yes | 21 (1%) | **Keep** - CSS custom property |
| `appendTo` | yes | yes | 21 (1%) | **Drop candidate** - web components handle shadow DOM differently |
| `distance` | yes | yes | 20 (1%) | **Keep** - CSS custom property |
| `exitDelay` | yes | yes | 19 (1%) | **Keep** |
| `flipBehavior` | yes | yes | 15 (1%) | **Keep (simplified)** - Floating UI handles natively |
| `className` | yes | yes | 28 (1%) | **Native HTML** - `class` |
| `animationDuration` | yes | yes | 8 (<1%) | **Keep** - CSS custom property |
| `onTooltipHidden` | yes | yes | 6 (<1%) | **Keep** - event |
| `hidden` | HTML | yes | 38 (2%) | **Native HTML** |
| `removeFindDomNode` | yes | yes | 7 (<1%) | **Drop** - React-specific workaround |
| `zIndex` | yes | yes | 1 (<1%) | **Drop candidate** - CSS custom property |
| `minWidth` | yes | yes | 1 (<1%) | **Drop candidate** - 1 usage across 2296 instances |
| `role` | HTML | yes | 2 (<1%) | **Native HTML** |
| `data-testid` | HTML | yes | 28 (1%) | **Native HTML** |
| `data-test` | HTML | yes | 5 (<1%) | **Native HTML** |
| `aria-label` | HTML | yes | 30 (1%) | **Native HTML** |

## Web Component API Recommendations

**Core attributes:** `content`, `position`, `trigger`, `distance`, `entry-delay`, `exit-delay`, `max-width`, `animation-duration`, `enable-flip`

**Boolean attributes:** `content-left-aligned`, `visible`, `no-flip` (invert of `enableFlip`)

**Events:** `pf-hidden` (replaces `onTooltipHidden`)

**Drop:** `minWidth`, `zIndex`, `removeFindDomNode`, `appendTo`, `flipBehavior` (Floating UI handles). All handleable via CSS custom properties or Floating UI defaults.

**48 spread usages** may pass unlisted props -- worth spot-checking those repos.

</details>

adamjohnson:
> @adamjohnson and @hellogreg -- could you verify the actual screen reader experience here? Want to confirm the tooltip content is still announced via proximity/`role="tooltip"` even without the formal `describedby` association.

I can confirm from manual testing that the tooltip content is still available without describedby, but SR announcement today depends on the live-region workaround, not role="tooltip" alone.

I pushed some docs changes to further note/document the limitations of `ariaDescribedByElements` lightdom -> shadowdom and how we work around that with the announcer (`role="status"`). Added some tests around this as well.

hellogreg:
Checked the demos with the usual screen reader suspects. Two notes:

The **click trigger** demo tooltip displays on keyboard focus, not just on click. The Windows screen readers read the tooltip content when this happens. Mac VoiceOver does not read the tooltip content in this demo ever--neither on click nor on focus.

The **manual trigger** demo reads the tooltip when the trigger checkbox is checked, but does _not_ also read it when the button associated with the tooltip receives focus. Not sure if this is intended or not, but thought I'd point it out.

adamjohnson:
Thanks for your comments, Greg! 

I went ahead and removed the click trigger demo because it's basically an antipattern for pf tooltip. Users really want a toggle tip here and should use the native [Popover API ](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API/Using) for this functionality. 

As far as the manual trigger, I believe this functionality is intentional. But we do want to make sure we communicate that appropriately. So, I improved the description for that demo and added a quick note in a `<p>` tag on the demo page. This helps implementors know the functional and accessibility implications of this behavior.

adamjohnson (approved):
Lightweight Generation Thoroughly Modernized

NB: the markdown files reference a slotted `<button>` element for the trigger; however, the actual demos reference `<pf-v5-button>` elements. Both work, but it's something to keep an eye on for later when we have `<pf-v6-button>`.

PR #3143, opened 2026-05-12
feat(banner)!: port to pf-v6-banner

### LINKED ISSUE #2983: [feat]: Update `<pf-v6-banner>` for PatternFly v6
## Summary

Update `<pf-v6-banner>` to implement the PatternFly v6 [Banner](https://www.patternfly.org/components/banner) component.

Full-width inline notification banner for conveying status or promotional information.

## Prior Art

| Source | Element | Status |
|--------|---------|--------|
| PFE v4 | `<pf-banner>` | Exists |
| cem | -- | -- |
| cockpit | -- | -- |
| chickadee | `<chickadee-banner>` | Exists |

## Requirements

- MUST provide end-user feature parity with `@patternfly/react-core` Banner

zeroedin:
## Summary

Closes #2983

- Removes `pf-v5-banner`
- Ports `pf-v5-banner` to `pf-v6-banner`  and aligned with the React v6 component
- Replaces v5 `variant` attribute with separate `color` (8 decorative colors) and `status` (5 semantic statuses) attributes, matching React's split API
- Adds `screen-reader-text` attribute for visually-hidden status announcements (WCAG 1.3.1)
- Retains `sticky` attribute (maps to React `isSticky`)
- Removes v5 `icon` attribute/slot — icons are now composed in the default slot, matching React v6
- All CSS uses v6 design tokens with hex fallbacks via `light-dark()` for standalone usage

## Deviations from React

- **Status demo uses `pf-v5-icon`**: The status demo composes icons in the default slot using `pf-v5-icon` because `pf-v6-icon` is not yet merged to `staging/pfv6`. Once pf-v6-icon lands, the status demo should be updated to use `pf-v6-icon` with the `rh-ui-*-fill` icon set to match the React demos visually.
- **Sticky demo**: React has `isSticky` as a prop but no dedicated demo on patternfly.org. We include a standalone sticky demo since the behavior requires scrollable content to verify — this was carried forward from the v5 element.

## Test plan

- [x] 36 unit tests passing (instantiation, upgrade, all 8 colors, all 5 statuses, sticky positioning, slotted link a11y, screen-reader-text presence/absence)
- [x] Visual comparison against patternfly.org in both light and dark mode
- [x] `/review-a11y` — no critical issues
- [x] `/review-api` — no critical issues
- [x] `/review-demos` — all demos match patternfly.org parity
- [x] `cem health` — 87/100 (above 80% threshold)
- [ ] Verify demos at `http://localhost:8000/elements/banner/demo/{slug}?rendering=chromeless`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


elements/pf-v6-banner/pf-v6-banner.ts
```diff
@@ -0,0 +1,87 @@
...
+  @property({ reflect: true }) color?: BannerColor;
+
+  /** Status style for the banner. Conveys semantic meaning and overrides `color`. */
+  @property({ reflect: true }) status?: BannerStatus;
+
+  /** Whether the banner sticks to the top of its container. */
+  @property({ type: Boolean, reflect: true }) sticky = false;
+
+  /** Text announced by screen readers to indicate the type of banner. */
+  @property({ attribute: 'screen-reader-text' }) screenReaderText?: string;
```
bennypowers: this attribute name is vague: it's not immediately clear which text this is. only after reading the jsdoc does the user discover that this text describes the "type" of banner. "type", mind you, is not an attribute on this element, so that term is also vague. moreover, we use `accessible-label` as the name for screen reader texts in other elements.

@adamjohnson wdyt?

elements/pf-v6-banner/pf-v6-banner.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  override render(): TemplateResult {
+    return html`
+      <div id="container" class=${classMap({
+        [this.color ?? '']: !!this.color,
+        [this.status ?? '']: !!this.status,
+      })}>
+        ${!this.screenReaderText ? nothing
```
bennypowers:
```suggestion
    const { color = '', status = '' } = this;
    return html`
      <div id="container" class="${classMap({ [color]: !!color, [status]: !!status })}">
        ${!this.screenReaderText ? nothing
```

minor formatting nit

elements/pf-v6-banner/pf-v6-banner.ts
```diff
@@ -0,0 +1,87 @@
...
+          : html`<span class="sr-only">${this.screenReaderText}</span>`}
+        <slot></slot>
+      </div>
```
bennypowers:
```suggestion
        <span class="sr-only" ?hidden="${!this.screenReaderText}">${this.screenReaderText}</span>
        <slot></slot>
      </div>
```

bennypowers (changes requested):

bennypowers:
this should delete pf-v5-banner and update all references

elements/pf-v6-banner/pf-v6-banner.ts
zeroedin: It's directly from React, `screenReaderText` is the prop name in Banner.tsx. The agent mapped it to `screen-reader-text` as the HTML attribute (dash-case convention).

I definitely think this is where we can make a judgment call, but we have to ensure we are consistent across the board here.

zeroedin:
> this should delete pf-v5-banner and update all references

Ok, good, that answers a question I had about the exports previously. We could update the `/update-element` skill to do this upon completion, unless you find value in keeping the folder around for any reason post-agent update?

elements/pf-v6-banner/pf-v6-banner.ts
bennypowers: in this case, let's lean to "authentic to the end user react experience" more, and "authentic to the pfe developer experience" more

zeroedin:
Need to update README.md to include divergences still.

elements/pf-v6-banner/pf-v6-banner.ts
adamjohnson: I just changed `screen-reader-text` to `accessible-label` in [pf-v6-helper-text](https://github.com/patternfly/patternfly-elements/pull/3155#discussion_r3309236333).

bennypowers:
The React Banner uses a visually-hidden `<span>` to prepend screen reader
text before the banner content:

```tsx
{screenReaderText && <span className="pf-v6-screen-reader">{screenReaderText}</span>}
{children}
```

For the web component, I'd like to use `ElementInternals` instead of a sr-only span. Here's what I've found exploring the options.

## The problem with `ariaLabel`

`ariaLabel` via ElementInternals **replaces** the computed accessible name. Given:

```html
<pf-v6-banner status="success" accessible-label="Success:">
  Job 25 completed
</pf-v6-banner>
```

With `internals.ariaLabel = "Success:"`, a screen reader announces

> "Success:, status"

The slotted content "Job 25 completed" is hidden from the name computation entirely.

There's also a spec issue: `aria-label` has no effect on elements with `role="generic"` (the default for custom elements), so without adding a role it's silently ignored.

## Proposed approach: `role="status"` + `ariaDescription`

```ts
#internals = InternalsController.of(this);

@observes('status')
protected statusChanged(): void {
  this.#internals.role = this.status ? 'status' : null;
}

@observes('accessibleLabel')
protected accessibleLabelChanged(): void {
  this.#internals.ariaDescription = this.accessibleLabel || null;
}
```

`ariaDescription` is announced **after** the content, not instead of it. Screen reader would announce:

**"Job 25 completed"** ... pause ... **"Success"** ... **"status"**

Content first, then context. No DOM hacks, no sr-only spans.

### Why `role="status"`

- Not a landmark (no landmark pollution from multiple banners)
- Implicit `aria-live="polite"` -- dynamically inserted status banners get announced automatically, which is the correct behavior
- Already-present banners don't re-announce on page load (live regions only announce subsequent changes)
- Only applied when `status` attribute is set -- color-only decorative banners stay `role="generic"`, no ARIA needed
- Supports `ariaDescription`

### Conditional: only on status banners

Color-only banners (`<pf-v6-banner color="blue">`) are decorative -- they get no role and no description. The docs already say `accessible-label` SHOULD only be used with status banners, so tying role + description to the `status` attribute enforces that guidance at the API level.

## Comparison

| Approach                            | SR announces                              | Pros                                               | Cons                                                                      |
| ----------------------------------- | ----------------------------------------- | -------------------------------------------------- | ------------------------------------------------------------------------- |
| sr-only span (React)                | "Success: Job 25 completed"               | Matches React exactly, simple                      | Extra DOM node, no live region, no semantic role                          |
| `ariaLabel`                         | "Success:, status"                        | No extra DOM                                       | Content hidden from AT, wrong                                             |
| `ariaDescription` + `role="status"` | "Job 25 completed ... Success ... status" | Content first, live region, no extra DOM, semantic | Different ordering than React, `ariaDescription` is newer (baseline 2024) |

Looking for a11y review on whether the `ariaDescription` ordering (content first, then status context) is acceptable, or whether the React-style prepended text is preferred.

cc @adamjohnson @hellogreg

bennypowers:
in badge, we removed screenReaderText and advised users to slot visually hidden text into the anonymous slot. Should do that here for API simplicity, or is the difference intentional -- banner is more opinionated (status banners SHOULD have sr text, so the element provides a shortcut) while badge is simpler (just a pill, user decides)?

bennypowers (approved):

PR #3145, opened 2026-05-13
chore: remove v5 element after update

zeroedin:
## What I did

1. Ensure we remove v5 element after update to v6

## Testing Instructions

1.

## Notes to Reviewers

1.

### DIFF
--- a/.claude/skills/update-element/SKILL.md
+++ b/.claude/skills/update-element/SKILL.md
@@ -257,7 +257,7 @@ See `.claude/skills/review-api/SKILL.md` Phase 4 for full test checklist.
 
 ### Phase 8: Delete v5 element
 
-Delete the directory and it's contents from ./elements/`pf-v5-{name}`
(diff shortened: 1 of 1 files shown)


.claude/skills/update-element/SKILL.md
```diff
@@ -255,7 +255,11 @@ Update existing tests -- don't rewrite from scratch:
 
 See `.claude/skills/review-api/SKILL.md` Phase 4 for full test checklist.
 
-### Phase 8: Audit
+### Phase 8: Delete v5 element
+
+Delete the directory and it's contents from ./elements/`pf-v5-{name}`
```
bennypowers:
```suggestion
Delete the directory and it's contents from ./elements/`pf-v5-{name}`. replace all references to `pf-v5-{name}`, except historical changelogs, and update their usage to conform to new APIs. This includes demos for v6 elements which use v5 dependencies.
```

bennypowers (approved):

PR #3146, opened 2026-05-13
chore: improve phase 5 cem alignment documentation comments

zeroedin:
## What I did

1.  Improved skill to catch incorrect usage of JSDoc or css vars, css part and slot which should be documented inline.

## Testing Instructions

1.

## Notes to Reviewers

1.


bennypowers (approved):
locally generated textual manuals

PR #3147, opened 2026-05-13
chore: improve review-api skill, lit template quoted attributes

zeroedin:
## What I did

1.  Add rule to check for quoted attributes in lit templates

## Testing Instructions

1.

## Notes to Reviewers

1.


bennypowers (approved):
lucidly greeted textual middlewords

PR #3152, opened 2026-05-19
feat(progress)!: port to pf-v6-progress element

zeroedin:
## Summary

- Ports `<pf-v5-progress>` to `<pf-v6-progress>` web component for PatternFly v6
- `role="progressbar"` via ElementInternals with `ariaValueNow`, `ariaValueMin`, `ariaValueMax`
- `description` for visible title text, `accessible-label` for screen reader name, `accessible-labelledby` and `accessible-describedby` for cross-root ARIA via ElementInternals
- Accessible name fallback chain: `accessible-labelledby` → `accessible-label` → `description` → `"Progress status"`
- Supports `size` (sm/lg), `measure-location` (outside/inside/none), `variant`
  (success/warning/danger)
- `truncated` boolean for CSS ellipsis overflow on description text
- `hide-status-icon` to suppress variant icon while keeping variant bar color
- `value-text` for custom aria-valuetext (finite step, step instruction)
- `helper-text` slot for supplementary content below the bar
- Singleline layout derived automatically when no `description` is set
- Danger variant includes subtle jiggle animation with `prefers-reduced-motion` support
- All CSS uses `--pf-v6-c-progress--*` design tokens with hex fallbacks
- 20 demos, 33 unit tests

## Intentional divergences from React

- `title` → `description` attribute to avoid shadowing native HTML `title`
- `aria-label` abstracted behind `accessible-label` attribute per ADVICE.md
- `aria-labelledby` abstracted behind `accessible-labelledby` attribute, resolves IDs to `ariaLabelledByElements` via ElementInternals
- `aria-describedby` abstracted behind `accessible-describedby` attribute, resolves IDs to `ariaDescribedByElements` via ElementInternals
- `isTitleTruncated` → `truncated` boolean attribute
- `label` (ReactNode) → `value-text` (string only)
- `measureLocation="top"` is the default (no attribute needed)
- `helperText` prop → `helper-text` slot for rich content
- `tooltipPosition` not implemented (blocked on `pf-v6-tooltip`)
- Status icons use FontAwesome SVGs; will switch to `pf-v6-icon` when available

## Test plan

- [ ] `npm run test` passes
- [ ] Demos render correctly at `localhost:8000` with `?rendering=chromeless`
- [ ] Visual parity with patternfly.org progress demos
- [ ] Screen reader announces `progressbar` role with correct label
- [ ] `accessible-label` sets aria-label without rendering visible text
- [ ] `accessible-labelledby` resolves IDs and sets `ariaLabelledByElements`
- [ ] `accessible-describedby` resolves IDs and sets `ariaDescribedByElements`
- [ ] `truncated` truncates description with ellipsis
- [ ] `hide-status-icon` suppresses variant icon while keeping bar color
- [ ] Variant icons display for success, warning, and danger
- [ ] `prefers-reduced-motion` disables danger jiggle animation
- [ ] `cem health` score above 80%

Assisted-By: Claude <noreply@anthropic.com>

## Testing Instructions

1.

## Notes to Reviewers

1.


elements/pf-v6-progress/pf-v6-progress.ts
```diff
@@ -0,0 +1,219 @@
...
+  @property({ type: Boolean, attribute: 'hide-status-icon' }) hideStatusIcon = false;
+
+  /** Custom text for aria-valuetext, used for finite step and step instruction displays */
+  @property({ attribute: 'value-text' }) valueText?: string;
+
+  /** Space-separated ID(s) of elements that label this progress bar. Resolves cross-root aria-labelledby via ElementInternals. */
+  @property({ attribute: 'accessible-labelledby' }) accessibleLabelledby?: string;
+
+  /** Space-separated ID(s) of elements that describe this progress bar. Resolves cross-root aria-describedby via ElementInternals. */
+  @property({ attribute: 'accessible-describedby' }) accessibleDescribedby?: string;
```
bennypowers: these won't work, x-root aria.

cc @adamjohnson

bennypowers (changes requested):
please implement https://www.patternfly.org/components/progress#interactive-status-icon-and-measure-location demo (and all other missing demos from react site)

and rework the prose in the README tables - it's not clear what, if any, differences exist for e.g. the `measureLocation` prop.

elements/pf-v6-progress/pf-v6-progress.ts
zeroedin: Maybe I misunderstood, but this uses the reflection api resolving IDs outward.  `ariaLabelledByElements`, `ariaDescribedByElements`. That said if the ID doesn't exist in the lightdom, but a shadowroot of another component, then that is an issue for sure. If that is the suggestion in this comment, then we can remove it.

bennypowers: thanks for pushing back. tl;dr: we can just drop these apis altogether. exposing them is a footgun for x-root aria, and there are workarounds for every call site I found. we can restore them when reference target lands

<details><summary>Cross-repo analysis of React component `aria-describedby` usage</summary>

# Progress A11y API Analysis

Cross-root ARIA ID references (`accessible-labelledby`, `accessible-describedby`) proposed in pf-v6-progress are a footgun. This report audits real-world usage to determine if they can be replaced with simpler alternatives.

## Scan Results

Scanned **160** repos, **34** use `<Progress>`, **101** total instances.

| Prop                   | Instances | %      |
| ---------------------- | --------- | ------ |
| `title`                | 56        | 55%    |
| `label`                | 32        | 32%    |
| `aria-label`           | 29        | 29%    |
| `valueText`            | 6         | 6%     |
| **`aria-labelledby`**  | **2**     | **2%** |
| **`aria-describedby`** | **2**     | **2%** |

`aria-labelledby` and `aria-describedby` together account for 4 instances across 3 unique callsites (one is a mirror repo).

## The Problem

In shadow DOM, `aria-labelledby="some-id"` on an element inside a shadow root cannot reference an element by ID outside that shadow root. The IDs live in different DOM scopes.

`ElementInternals.ariaLabelledByElements` resolves outward from a custom element's internals, but only to elements in the host's light DOM. If the target is in a *different* shadow root, it silently fails -- no error, just a missing accessible name.

The proposed `accessible-labelledby` / `accessible-describedby` attributes paper over this with ElementInternals, but the limitation remains: they only work when the referenced ID is in the custom element's parent scope.

## Callsite Analysis

### 1. stackrox -- `aria-labelledby` (ID ref to sibling element)

**File:** `ui/apps/platform/src/Containers/Compliance/Dashboard/ComplianceScanProgress.tsx`

```tsx
function ComplianceScanProgress({ runs, ...props }: ComplianceScanProgressProps) {
    const unfinishedRunCount = runs.filter((run) => run.state !== 'FINISHED').length;
    const finishedRunCount = runs.length - unfinishedRunCount;
    const title = unfinishedRunCount === 0
        ? 'Compliance scanning complete'
        : 'Compliance scanning in progress';

    return (
        <Card {...props}>
            <CardTitle id="compliance-scan-progress-title">{title}</CardTitle>
            <CardBody>
                <Progress
                    aria-labelledby="compliance-scan-progress-title"
                    size="sm"
                    variant={unfinishedRunCount === 0 ? 'success' : undefined}
                    value={finishedRunCount}
                    min={0}
                    max={runs.length}
                    measureLocation={'outside'}
                    label={`${finishedRunCount} of ${runs.length} runs`}
                    valueText={`${finishedRunCount} of ${runs.length} runs`}
                />
            </CardBody>
        </Card>
    );
}
```

**What it does:** Labels the progress bar with the CardTitle text ("Compliance scanning in progress" / "Compliance scanning complete").

**Web component workaround:** Use `aria-labelledby` directly, since the `progressbar` role is on the pf-v6-progress host.

```html
<pf-v6-card>
  <h2 id="title" slot="title">It's my title and i'll cry if i want to</h2>
  <pf-v6-progress aria-labelledby="title"
                  value="3" max="5"
                  value-text="3 of 5 runs">
  </pf-v6-progress>
</pf-v6-card>
```

The `aria-label` duplicates the visible heading, which is fine -- the heading provides visual context, `aria-label` provides the programmatic name. In React the `title` prop already does exactly this (renders visible text *and* sets `aria-labelledby` to it internally).

### 2. content-services -- `aria-labelledby` (passthrough prop)

**File:** `src/Pages/.../UploadContent/components/UploadStatusItem.tsx`

```tsx
<Progress
  className={progressVariant === 'danger' ? classes.errorText : ''}
  title={<span className={...}>
    <ConditionalTooltip ...>
      <span>{reduceStringToCharsWithEllipsis(fileName, maxFileNameLength)}</span>
    </ConditionalTooltip>
    <span>{getHumanReadableFileSize(fileSize || 0)}</span>
  </span>}
  value={progressValue}
  label={progressLabel}
  variant={progressVariant}
  aria-label={progressAriaLabel}
  aria-labelledby={progressAriaLabelledBy}
  id={progressId}
  helperText={progressHelperText}
/>
```

**What it does:** Passes `aria-labelledby` through from a parent component. But this component *also* sets `aria-label` and `title` on the same element. In React's Progress implementation, `title` takes precedence for `aria-labelledby` (see line 110 of Progress.tsx: it overrides the passed `aria-labelledby` when `title` is present). So the `aria-labelledby` prop here is dead code -- `title` wins.

**Web component workaround:** Not needed. `aria-label` already present, title already present. Remove the dead `aria-labelledby`.

### 3. open-cluster-management -- `aria-describedby` (broken ID ref)

**File:** `frontend/src/wizards/RoleAssignment/CommonProjectCreateProgressBar.tsx`

```tsx
export const CommonProjectCreateProgressBar = ({
  successCount = 0, errorCount = 0, totalCount, hideTitle = false,
}: CommonProjectCreateProgressBarProps) => {
  const { t } = useTranslation()
  const progressValue = ((successCount + errorCount) / totalCount) * 100
  return (
    <Progress
      aria-describedby="common-project-create-progress-bar"
      value={progressValue}
      title={hideTitle ? undefined : t('Creating common projects')}
      aria-label={t('Creating common projects')}
      helperText={
        <HelperText id="common-project-create-progress-bar-helper-text">
          <HelperTextItem variant="error">
            {t('Failed to create common projects. Error: ...')}
          </HelperTextItem>
        </HelperText>
      }
      variant={isErrorState ? 'danger' : 'success'}
    />
  )
}
```

**What it does:** Tries to associate the progress bar with a description element. But this is already a bug: `aria-describedby` references `"common-project-create-progress-bar"` while the HelperText has id `"common-project-create-progress-bar-helper-text"`. The IDs don't match, so `aria-describedby` resolves to nothing.

**Web component workaround:** Use the helper-text slot. pf-v6-progress can internally wire `aria-describedby` to slotted helper text via ElementInternals, keeping everything within the same shadow root:

```html
<pf-v6-progress accessible-label="Creating common projects" value="75">
  <pf-v6-helper-text slot="helper-text">
    <pf-v6-helper-text-item variant="error">
      Failed to create common projects. Error: 2. Success: 3.
    </pf-v6-helper-text-item>
  </pf-v6-helper-text>
</pf-v6-progress>
```

</details>

bennypowers: summary of conclusions on call

- remove these props
- users can use aria-labelledby directly on the host, since it has the implicit role
- we'll add a slotchange listener on helper-text slot, which adds all helper-text elements to the host's internals ariaDescribedBy list - no x-root aria problem, everything is light dom.

bennypowers (approved):

PR #3153, opened 2026-05-19
fix(tools): support tagPrefix as an array of tags

zeroedin:
## Summary

  Adds support for `tagPrefix` as `string | string[]` in `.pfe.config.json`, enabling both `pf-v5-*` and `pf-v6-*` elements to coexist in the same repo during the v6 migration.

  Previously, `tagPrefix` was a single string (`"pf-v5"`). The test runner middleware, docs pages, and manifest tooling all assumed a single prefix, so v6 elements couldn't get `.js` → `.ts` redirects in tests,
  correct prefix stripping in docs, or proper demo metadata.

  ## Changes

  - `.pfe.config.json` — changed `tagPrefix` from `"pf-v5"` to `["pf-v5", "pf-v6"]`
  - `tools/pfe-tools/config.ts` — updated `PfeConfig.tagPrefix` type to `string | string[]`; updated `deslugify()` to check all prefixes using `[value].flat()`
  - `tools/pfe-tools/test/config.ts` — normalized prefix to array; middleware now matches any prefix via `.some()`
  - `tools/pfe-tools/11ty/DocsPage.ts` — finds matching prefix for tagName from array
  - `tools/pfe-tools/custom-elements-manifest/lib/Manifest.ts` — same prefix-matching pattern for demo metadata
  - `tools/pfe-tools/11ty/plugins/types.ts` — updated `DemoRecord.tagPrefix` type to `string | string[]`

  ## Known Issues

  - **Permalink collisions**: 11ty generates the same permalink for elements sharing a base name (e.g. `pf-v5-spinner` and `pf-v6-spinner` both map to `/components/spinner/`). Only one version should exist at a time
   — remove the v5 element when the v6 replacement is ready.
  - **Cross-element dependencies**: Some v5 elements import other v5 elements directly (e.g. `pf-v5-button` imports `pf-v5-spinner`). When a v5 element is replaced by its v6 version, any v5 elements that depend on
  it must also be updated to import the v6 replacement, or the old v5 element must remain until all its consumers are migrated.
  - **`docs/main.mjs` manual updates**: This file eagerly imports all elements for the 11ty docs site. When swapping a v5 element for its v6 replacement, the import must be manually updated.

  ## Test plan

  - [ ] `npm run build` — no type errors
  - [ ] `npx wtr elements/pf-v5-accordion/test/pf-accordion.spec.ts` — v5 element tests still pass
  - [ ] Copy a `pf-v6-*` element into the repo and run its tests without pre-building — `.js` → `.ts` redirect works
  - [ ] Dev server docs pages render correct element names for both prefixes

## Testing Instructions

1.

## Notes to Reviewers

1.

### DIFF
--- a/.changeset/fancy-keys-hug.md
+++ b/.changeset/fancy-keys-hug.md
@@ -1,0 +1,5 @@
+---
+"@patternfly/pfe-tools": minor
+---
+
+Accept `tagPrefix` as an array of strings, supporting repos with multiple element prefixes
--- a/.pfe.config.json
+++ b/.pfe.config.json
@@ -1,4 +1,4 @@
 {
   "renderTitleInOverview": false,
-  "tagPrefix": "pf-v5"
+  "tagPrefix": ["pf-v5", "pf-v6"]
 }
--- a/tools/pfe-tools/11ty/DocsPage.ts
+++ b/tools/pfe-tools/11ty/DocsPage.ts
@@ -1,5 +1,5 @@
 import type { PfeConfig } from '../config.js';
-import { getPfeConfig } from '../config.js';
+import { getPfeConfig, matchPrefix } from '../config.js';
 import { Manifest } from '../custom-elements-manifest/lib/Manifest.js';
 
 import slugify from 'slugify';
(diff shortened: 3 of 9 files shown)


bennypowers:
> - Permalink collisions
> - Cross-element dependencies

Arguably, these are both legitimately error states. This is so specified in the skill files

> - `docs/main.mjs` manual updates

Ideally we'd import in situ instead of a big master barrel import. Could be done in a later pr

bennypowers (approved):

